PR #10, opened 2018-09-24
global: add dependencies to setup.py

zzacharo:


ntarocco (approved):
From a quick look, should the `ui/public` folder be committed? Or it is the result of the build?

zzacharo:
@ntarocco the public folder is a result of the `create-react-app` and is copied on the build folder. Currently we store the home background image there and potentially in the future all the static images we are gonna use. If nothing else should be changed feel free to merge this.

PR #33, opened 2018-10-12
tests: fix tests and run-tests.sh

ntarocco:
So...

No solution for now for `check-manifest` but also `python setup.py test`: these commands included in `run-tests.sh` keep checking `node_modules` folders, which are **big**, despite my temptatives of ignoring them with .gitignore and MANIFEST.in.
We will see in the future how to do it when we have the final UI build setup.


ntarocco:
@zzacharo we will need to review the:
```
@main_blueprint.route("/<path:path>", methods=["GET"])
@backoffice_blueprint.route("/<path:path>", methods=["GET"])
```
because there are some issues...

zzacharo:
> @zzacharo we will need to review the:
> 
> ```
> @main_blueprint.route("/<path:path>", methods=["GET"])
> @backoffice_blueprint.route("/<path:path>", methods=["GET"])
> ```
> because there are some issues

I think I know :) You mean something like that?
```
@main_blueprint.route("/", methods=["GET"])
@main_blueprint.route("/<path:path>", methods=["GET"])
```

I am making some changes in my upcoming PR which I found while testing the views. Let's discuss it on Monday!

tests/ui/test_views.py
```diff
@@ -9,23 +9,48 @@
...
-    resp = client.get(url_for('invenio_app_ils_backoffice_ui.index'))
+@pytest.mark.skip("Main view is intercepting and serving backoffice view")
+def test_anonymous_cannot_access_backoffice(client):
+    """Test anonymous cannot access the backoffice view."""
+    backoffice = url_for('invenio_app_ils_backoffice_ui.backoffice', path="")
+    resp = client.get(backoffice)
     assert resp.status_code == 401
+
+
+@pytest.mark.skip("Main view is intercepting and serving backoffice view")
```
zzacharo: Oh this was what I have started working on! Ok so let's discuss on Monday some changes that I did on the views that you have already spotted.

invenio_app_ils/views.py
```diff
@@ -56,6 +54,12 @@ def index(path):
     return render_template("invenio_app_ils/main.html")
 
 
+@main_blueprint.route('/ping', methods=['HEAD', 'GET'])
```
zzacharo, line 58: Because you have done the testing part already, I would change the line 53 with the below
```
@main_blueprint.route("/", methods=["GET"])
@main_blueprint.route("/<path:path>", methods=["GET"])
```
as otherwise the route to slash is giving an error.

invenio_app_ils/views.py
```diff
@@ -56,6 +54,12 @@ def index(path):
...
+def ping():
+    """Ping blueprint used by loadbalancer."""
+    return 'OK'
+
+
```
zzacharo, line 63: same for line 72 
```
@backoffice_blueprint.route("/", methods=["GET"])
@backoffice_blueprint.route("/<path:path>", methods=["GET"])
```

zzacharo (changes requested):

invenio_app_ils/views.py
zzacharo: Also to be consistent I had changed the line 72 from `backoffice` to index :)

topless (approved):
Apart from the skip isort section, where I guess there must be a problem or issue looks good :)

zzacharo (approved):

PR #36, opened 2018-10-17
loanDetails: add tests

zzacharo:


invenio_app_ils/ui/backoffice/package.json
```diff
@@ -9,6 +9,7 @@
     "pre-build": "react-scripts build",
     "build": "npm-run-all pre-build move-js move-css",
     "test": "react-scripts test --env=jsdom",
+    "test:ci": "CI=true react-scripts test --env=jsdom",
```
ntarocco, line 12: Was it for what `CI=true` again?

.travis.yml
```diff
@@ -61,14 +66,30 @@ script:
...
+          - node_modules
+      before_install:
+        - cd "./invenio_app_ils/ui/backoffice"
+      install:
+        - npm install
+      script:
+        - npm run test:ci
+    - stage: deploy
+      script: skip
+      deploy:
```
ntarocco, line 86: I think we will never deploy this, we will deploy from a Python build, no?

invenio_app_ils/ui/backoffice/src/App.test.js
```diff
@@ -1,9 +1,9 @@
-import React from "react";
-import ReactDOM from "react-dom";
-import App from "./App";
+import React from 'react';
```
ntarocco, line 1: Are we using Prettier? because, for me, it is putting this `"` in the import...
We need to all align on this.

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/components/LoanActions/tests/LoanAction.test.js
```diff
@@ -0,0 +1,41 @@
...
+import LoanAction from '../LoanActions';
+
+describe('LoanAction', () => {
+  it('renders initial state', () => {
+    const props = {
+      data: {},
+      onAction: () => 'onAction',
+      actions: {},
+    };
+    const component = shallow(<LoanAction {...props} />);
```
ntarocco, line 12: I guess you are aware, but this is also for @topless and @kprzerwa, `shallow` will not call `componentDidMount` and `componentDidUpdate` of your React component, often used. 
If you use that hooks in your component, and you need to test the full behaviour of the component, you have to use `mount` instead of `shallow`... just warning because I already had headaches with this :)
https://stackoverflow.com/questions/38710309/when-should-you-use-render-and-shallow-in-enzyme-react-tests

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/components/LoanActions/tests/__snapshots__/LoanAction.test.js.snap
```diff
@@ -0,0 +1,284 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
ntarocco, line 1: I think it is going to be hard to review all the time the snapshots, especially with enzyme, they are not very human readable and quite long.
Here what our friend Ben from Eventbrite at the React workshop was saying:
https://github.com/benmvp/react-workshop/tree/master/src/testing/02-render-components

About snapshots:
`However, the ease of writing the test comes at a cost for maintaining the test. For one, the snapshot for the one test is over 100 lines. It's not very easy to read. But it's very easy for the test to fail when changing unrelated things.`
`Each one of these cause the test the fail, when all we really are trying to verify is that a corresponding <EmailListItem /> is created for every email in the emails prop array. Jest allows us to update the snapshot when it fails, but with multiple developers the snapshots will get auto-adopted without really verifying if the failure was real or just a false-positive.
Instead, manually verify the specific things we care about`

So basically, he was saying that with multiple people updating snapshots, it can be tricky, better checking the needed props manually:
`expect(emailListItem).toHaveProp('email', EMAILS[index]);`

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/state/actions.js
```diff
@@ -1,53 +1,53 @@
...
-        type: SET_LOAN_ACTION_ERROR,
-        payload: reason,
-      });
-    });
-    if (details) {
-      dispatch({
-        type: SET_LOAN_ACTION,
-        payload: serializeLoanDetails(details.data),
+    return postLoan(url, data)
+      .then(details =>
```
ntarocco, line 40: 👏

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/state/tests/actions.test.js
```diff
@@ -0,0 +1,159 @@
...
+import axios from 'axios';
+// Actions to be tested
+import * as actions from '../actions';
+import { initialState } from '../reducer';
+import { serializeLoanDetails } from '../selectors';
+import * as types from '../types';
+
+const middlewares = [thunk];
+const mockStore = configureMockStore(middlewares);
+const loansBaseUrl = 'https://localhost:5000/api/circulation/loans';
```
ntarocco, line 13: `baseUrl = https://localhost:5000/api/` and put this somewhere in common to all tests?

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/state/tests/actions.test.js
```diff
@@ -0,0 +1,159 @@
...
+      mock.onGet(`${loansBaseUrl}/1`).reply(() => {
+        return new Promise((resolve, reject) =>
+          setTimeout(() => {
+            resolve([200, JSON.stringify(response)]);
+          }, 1000)
+        );
+      });
+      mock.onGet(`${loansBaseUrl}/2`).reply(() => {
+        return new Promise((resolve, reject) =>
+          setTimeout(() => {
```
ntarocco, line 36: why the `setTimeout`?

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/state/tests/actions.test.js
```diff
@@ -0,0 +1,159 @@
...
+    });
+
+    it('fires a fetch loan details action', done => {
+      const expectedActions = [
+        {
+          type: types.SET_LOAN_FETCH_LOADING,
+        },
+      ];
+
+      store.dispatch(actions.fetchLoanDetails(1)).then(() => {
```
ntarocco, line 50: Here it is a matter of readibility, as you were saying...
Do we consider more readable the current version or the `return` without `done()`? 
```
return store.dispatch(actions.fetchLoanDetails(1)).then(() => {
        let actions = store.getActions();
        expect(actions[0]).toEqual(expectedActions[0]);
```

ntarocco (commented):
Wow, nice setup and amount of tests, good job! 👏 :)

invenio_app_ils/ui/backoffice/package.json
zzacharo: To prevent jest to run in watch mode ;)

invenio_app_ils/ui/backoffice/src/App.test.js
zzacharo: Yes we use it! I had the same issue but I don't remember how I fixed it.

.travis.yml
```diff
@@ -61,14 +66,30 @@ script:
...
-    condition: $DEPLOY = true
+jobs:
+  include:
+    - stage: ui-test
+      language: node_js
+      node_js:
+        - 8
+      cache:
+        directories:
+          - node_modules
```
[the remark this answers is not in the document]
zzacharo: No clue actually, but I can try to make it global and see the time.

invenio_app_ils/ui/backoffice/package.json
zzacharo: To run jest in non watch mode!

.travis.yml
zzacharo: I moved the python deploy to a new stage so that it will deploy only if both backend and ui tests fail.

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/components/LoanActions/tests/LoanAction.test.js
zzacharo: Also there is the render function that just renders all the children components of your component which is the case if you want to tests your final rendered template.

invenio_app_ils/ui/backoffice/package.json
ntarocco: Right, then, what about changing it to something like:
`"test:nowatch": "nowatch=true react-scripts test --env=jsdom",`

Will be any useful to run the tests locally too? I just wouldn't "specialize it" for CI.

invenio_app_ils/ui/backoffice/src/App.test.js
ntarocco: But did you put any config file anywhere?

ntarocco (approved):

invenio_app_ils/ui/backoffice/package.json
zzacharo: Is this valid? I just saw that the CI=true is the only way to run without watch.

zzacharo: Still I can change the script name!

PR #37, opened 2018-10-18
records: permissions implementation for basic roles

### LINKED ISSUE #18: Support for restricted documents and items
Handle *_access* on documents/items

kpsherva:
* NEW Implements librarian, superuser and patron access policy
  to a record by using _access field of a record

* CHANGES config for RECORDS_REST_ENDPOINTS

My idea was to take 3 basic roles `Librarian`, `Patron`, `Admin` and implement the policy for actions:

- `read`: allowed by default for everyone, might be restricted by adding role or user to 
```
    {'_access': 'read': {}}
``` 
- `create`: allowed for `Librarian` and `Admin`, might be also allowed for specific users or roles by adding them in a field `create` in `_access`
- `update`: same policy as `create`
- `delete`: by default allowed only for `Admin`,  might be also allowed for specific users or roles by adding them in a field `delete` in `_access`

this seems for me to be the most flexible, and it might be overwritten depending on the needs. Also it's now implemented generally for record.
Feedback for this logic much appreciated.
(closes #18)

### DIFF
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -18,9 +18,9 @@ from __future__ import absolute_import, print_function
 from datetime import timedelta
 
 from invenio_app.config import APP_DEFAULT_SECURE_HEADERS
-from invenio_records_rest.utils import allow_all
 
 from .records.api import Document, InternalLocation, Item, Location
+from .records.permissions import RecordPermission
 from .search.api import DocumentSearch, InternalLocationSearch, ItemSearch, \
     LocationSearch
 
@@ -231,8 +231,10 @@ RECORDS_REST_ENDPOINTS = dict(
         default_media_type="application/json",
         max_result_window=10000,
         error_handlers=dict(),
-        create_permission_factory_imp=allow_all,
-        update_permission_factory_imp=allow_all,
+        read_permission_factory_imp=RecordPermission,
+        create_permission_factory_imp=RecordPermission,
+        update_permission_factory_imp=RecordPermission,
+        delete_permission_factory_imp=RecordPermission,
     ),
     itemid=dict(
         pid_type=ITEM_PID_TYPE,
@@ -255,8 +257,10 @@ RECORDS_REST_ENDPOINTS = dict(
         default_media_type="application/json",
         max_result_window=10000,
         error_handlers=dict(),
-        create_permission_factory_imp=allow_all,
-        delete_permission_factory_imp=allow_all,
+        read_permission_factory_imp=RecordPermission,
+        create_permission_factory_imp=RecordPermission,
+        update_permission_factory_imp=RecordPermission,
+        delete_permission_factory_imp=RecordPermission,
     ),
     locid=dict(
         pid_type=LOCATION_PID_TYPE,
@@ -279,7 +283,7 @@ RECORDS_REST_ENDPOINTS = dict(
         default_media_type="application/json",
         max_result_window=10000,
         error_handlers=dict(),
-        create_permission_factory_imp=allow_all,
+        create_permission_factory_imp=RecordPermission,
     ),
     ilocid=dict(
         pid_type=INTERNAL_LOCATION_PID_TYPE,
@@ -302,12 +306,12 @@ RECORDS_REST_ENDPOINTS = dict(
         default_media_type="application/json",
         max_result_window=10000,
         error_handlers=dict(),
-        create_permission_factory_imp=allow_all,
+        read_permission_factory_imp=RecordPermission,
+        create_permission_factory_imp=RecordPermission,
+        update_permission_factory_imp=RecordPermission,
+        delete_permission_factory_imp=RecordPermission,
     ),
 )
-
-# Allow all GET requests
-RECORDS_REST_DEFAULT_READ_PERMISSION_FACTORY = allow_all
 
 # RECORDS UI
 # ==========
--- a/invenio_app_ils/records/permissions.py
+++ b/invenio_app_ils/records/permissions.py
@@ -1,0 +1,86 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Ils records' permissions."""
+from flask_principal import ActionNeed, RoleNeed, UserNeed
+from invenio_access import Permission, superuser_access
+
+librarian_role = RoleNeed('librarian')
+create_records_action = ActionNeed('create-records')
+
+
+class RecordPermission(Permission):
+    """Record permission.
+
+    - Create action given to librarian, admin and specified users.
+    - Read access given to everyone with possibility to hide.
+    - Delete access to admin and specified users.
+    """
+
+    create_actions = ['create']
+    read_actions = ['read']
+    update_actions = ['update']
+    delete_actions = ['delete']
+
+    def __init__(self, record, action):
+        """Constructor."""
+        self.record = record
+        self.action = action
+        self._permissions = None
+        record_needs = self._dispatch()
+        super(RecordPermission, self).__init__(*record_needs)
+
+        # removes need of superuser for reading access
+        if self.allow_by_default:
+            self.explicit_needs.remove(superuser_access)
+
+    def _dispatch(self):
+        """Dispatch permission policy per action."""
+        if self.action in self.read_actions:
+            return self.read_permissions()
+        elif self.action in self.create_actions:
+            return [create_records_action] + self.librarian_permissions()
+        elif self.action in self.update_actions:
+            return self.record_needs() + self.librarian_permissions()
+        else:
+            return self.record_needs()
+
+    def read_permissions(self):
+        """Define read permission policy per record."""
+        if self.is_public():
+            # allow by default per everyone when read is empty
+            self.allow_by_default = True
+            return []
+        else:
+            return self.record_needs() + self.librarian_permissions()
+
+    def librarian_permissions(self):
+        """Define librarian role permission."""
+        return [librarian_role]
+
+    def record_allows(self):
+        """Read what record allows per action."""
+        return self.record.get('_access', {}).get(self.action, [])
+
+    def record_needs(self):
+        """Create needs of the record."""
+        needs = []
+        for access_entity in self.record_allows():
+            try:
+                needs.append(RoleNeed(access_entity.lower()))
+            except AttributeError:
+                needs.append(UserNeed(access_entity))
+        return needs
+
+    def is_public(self):
+        """Check if the record is fully public.
+
+        In practice this means that the record doesn't have the ``access``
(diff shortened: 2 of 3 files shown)


invenio_app_ils/records/permissions.py
```diff
@@ -0,0 +1,164 @@
...
+"""Ils records' permissions."""
+from flask_login import current_user
+from flask_principal import RoleNeed
+from invenio_access import DynamicPermission, action_factory
+from invenio_admin.permissions import action_admin_access
+
+from invenio_app_ils.records.utils import get_user_provides
+
+librarian_role = RoleNeed('librarian')
+create_action = action_factory('create')
```
ntarocco: Maybe more specific action name? `create-records` action?

invenio_app_ils/records/permissions.py
```diff
@@ -0,0 +1,164 @@
...
+    """Check if record provides access for user or role to perform action."""
+    user_id = int(user.get_id()) if user.is_authenticated else None
+
+    user_provides = get_user_provides()
+    allowed_users = record.get('_access', {}).get(action, [])
+    if allowed_users and not set(user_provides).isdisjoint(set(allowed_users)):
+        return True
+
+
+def has_admin_permission(user=None, record=None):
```
ntarocco: I am not sure you need this, because if you use `Permission` class, it is already adding admin access.
See: https://github.com/inveniosoftware/invenio-access/blob/master/invenio_access/permissions.py#L112

So, any permission you define, admin can already do it. (to be verified with tests! :))

invenio_app_ils/records/permissions.py
```diff
@@ -0,0 +1,164 @@
...
+def is_public(data, action):
+    """Check if the record is fully public.
+
+    In practice this means that the record doesn't have the ``access`` key or
+    the action is not inside access or is empty.
+    """
+    return '_access' not in data or not data.get('_access', {}).get(action)
+
+
+def has_record_permission(user, record, action):
```
ntarocco: it might be possible to take full advantage of `invenio-access` and rewrite this part as (pseudo-code here, warning!):
```
class RecordPermission(Permission):
    """Return Permission to evaluate if the current user can perform the action in the given record."""
    
    def __init__(self, record, action):
        """Constructor."""
        allowed_users = record.get('_access', {}).get(action, [])
        user_needs = [UserNeed(user_id) for user_id in allowed_users]
        super(RecordPermission, self).__init__(**user_needs)
```

And you don't need the utils.
The point is to write to minimize the code we have to create/maintain, and rely on the invenio packages.
**warning**: maybe this doesn't work :(

tests/api/test_record_permissions.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # delete access for user and librarian
+    ({'_access': {'delete': [1, 'librarian']}}, 'delete', True),
+
+])
+def test_record_generic_access(db, users, access, action, is_allowed):
+    """Test access control for records."""
+
+    @identity_loaded.connect
+    def mock_identity_provides(sender, identity):
```
ntarocco, line 44: nice!

tests/api/test_record_permissions.py
```diff
@@ -0,0 +1,97 @@
...
+        """Add additional role to the user."""
+        roles = [RoleNeed('test@invenio')]
+        identity.provides |= set(roles)
+
+    def login_and_test(user_id):
+        login_user(User.query.get(user_id))
+        # Create record
+        id = uuid.uuid4()
+        record = Record.create(access, id_=id)
```
ntarocco, line 55: maybe rename the param `record` to be more comprehensible?

tests/api/test_record_permissions.py
```diff
@@ -0,0 +1,97 @@
...
+
+])
+def test_record_generic_access(db, users, access, action, is_allowed):
+    """Test access control for records."""
+
+    @identity_loaded.connect
+    def mock_identity_provides(sender, identity):
+        """Add additional role to the user."""
+        roles = [RoleNeed('test@invenio')]
+        identity.provides |= set(roles)
```
ntarocco, line 48: 😯 I would say it is less common to use the bitwise OR operator, compared to `+=` or `-=`. I would add a comment explaining what we are trying to do here.

ntarocco (commented):
Nice job! 👏 Clean and concise!
Pardon me if I put many comments... check if they are relevant or ignore! :)

invenio_app_ils/records/permissions.py
```diff
@@ -0,0 +1,164 @@
...
+
+"""Ils records' permissions."""
+from flask_login import current_user
+from flask_principal import RoleNeed
+from invenio_access import DynamicPermission, action_factory
+from invenio_admin.permissions import action_admin_access
+
+from invenio_app_ils.records.utils import get_user_provides
+
+librarian_role = RoleNeed('librarian')
```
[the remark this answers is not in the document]
kpsherva: I was thinking about using the `backoffice_permission`, but the name suggested that it is going to be used for accessing `backoffice` specifically so UI, and this one was meant to be record access related, maybe in some cases not through UI. In this case I think either the permission should be renamed for something more generic than `backoffice` or record package should implement it's own set of permissions like I did. What do you think?

invenio_app_ils/config.py
```diff
@@ -267,12 +274,15 @@ def _(x):
...
+        read_permission_factory_imp=record_read_permission_factory,
+        create_permission_factory_imp=record_create_permission_factory,
+        update_permission_factory_imp=record_update_permission_factory,
+        delete_permission_factory_imp=record_delete_permission_factory,
     ),
 )
 
 # Allow all GET requests
-RECORDS_REST_DEFAULT_READ_PERMISSION_FACTORY = allow_all
+RECORDS_REST_DEFAULT_READ_PERMISSION_FACTORY = record_read_permission_factory
```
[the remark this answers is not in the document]
kpsherva: I thought it is meant to be there in case if some of the future `Record` types have no configured factory, I can remove it if it is not the case

invenio_app_ils/records/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
kpsherva: True, I can move some bits

invenio_app_ils/records/permissions.py
kpsherva: 👍

tests/api/test_record_permissions.py
kpsherva: += operator does not work for sets

invenio_app_ils/records/permissions.py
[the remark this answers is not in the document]
ntarocco: 3 time that I write this comment, let's if now it works.
I would try to merge both into one to keep it simple. I don't think there is the need of both.
Maybe check with @ludmilamarian?

invenio_app_ils/records/permissions.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self, record, action):
+        """Constructor."""
+        self.record = record
+        self.action = action
+        self._permissions = None
+        record_needs = self._dispatch()
+        super(RecordPermission, self).__init__(*record_needs)
+
+        # removes need of superuser for reading access
+        if self.allow_by_default:
```
ntarocco: I don't get this! 😨

invenio_app_ils/records/permissions.py
```diff
@@ -0,0 +1,86 @@
...
+    def record_allows(self):
+        """Read what record allows per action."""
+        return self.record.get('_access', {}).get(self.action, [])
+
+    def record_needs(self):
+        """Create needs of the record."""
+        needs = []
+        for access_entity in self.record_allows():
+            try:
+                needs.append(RoleNeed(access_entity.lower()))
```
ntarocco, line 69: Why this attribute error? Is it in case it is an `int`, like a user id?

kpsherva: yes, in case of `int`

invenio_app_ils/records/permissions.py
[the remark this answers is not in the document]
ntarocco: Maybe just rename this to something like `need_librarian_role`, so it means that the user needs to have the `librarian role`

invenio_app_ils/records/permissions.py
```diff
@@ -0,0 +1,86 @@
...
+    def read_permissions(self):
+        """Define read permission policy per record."""
+        if self.is_public():
+            # allow by default per everyone when read is empty
+            self.allow_by_default = True
+            return []
+        else:
+            return self.record_needs() + self.librarian_permissions()
+
+    def librarian_permissions(self):
```
ntarocco: what about `def get_needs_for_librarian_role(self)` to keep consistency? or just get rid of this function and use the list directly?

I just fear that we will be confused between roles, actions, needs, permissions, etc... naming

invenio_app_ils/records/permissions.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def record_allows(self):
+        """Read what record allows per action."""
+        return self.record.get('_access', {}).get(self.action, [])
+
+    def record_needs(self):
+        """Create needs of the record."""
+        needs = []
+        for access_entity in self.record_allows():
+            try:
```
ntarocco, line 63: If I understand correctly, `int` -> user id, `string` -> role... and the distinction is done using the `AttributerError` on the `.lower()` function success/fail.
I find maybe a bit "weak" or not clear. We should re-think this maybe?

invenio_app_ils/records/permissions.py
```diff
@@ -0,0 +1,86 @@
...
+    read_actions = ['read']
+    update_actions = ['update']
+    delete_actions = ['delete']
+
+    def __init__(self, record, action):
+        """Constructor."""
+        self.record = record
+        self.action = action
+        self._permissions = None
+        record_needs = self._dispatch()
```
ntarocco: `dispatch()` -> `collect_needs()`? it is more about getting rather than dispatching?

invenio_app_ils/records/permissions.py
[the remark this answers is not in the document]
kpsherva: as we have explained it is a RoleNeed not a role duplication

invenio_app_ils/records/permissions.py
kpsherva: just for leaving a note because we have discuss about it: the line removes the superuser-access from the set of Needs

invenio_app_ils/records/permissions.py
kpsherva: I've changed the logic, I'm checking now if it might be converted to int, WDYT?

invenio_app_ils/records/permissions.py
kpsherva: duplicated comment, please read below

ntarocco (approved):

zzacharo (approved):

PR #39, opened 2018-10-19
Refactored api and details pages

### LINKED ISSUE #35: Shared module
Currently we have some modules that are almost identical. Remove duplications and move shared logic and components under shared module.

topless:
- Removed `basemame` and added it to every route 
- Api moved under common functionality (close #35)
- Default exports of details pages are named exports now from index wrapped withRouter


invenio_app_ils/ui/backoffice/src/common/api.js
```diff
@@ -0,0 +1,27 @@
+import axios from 'axios';
+
+// In development create a personal token to be able to perform requests and
+// and put it in a .env file under the name `REACT_APP_JWT_TOKEN`
+const token =
+  process.env.NODE_ENV === 'development'
+    ? process.env.REACT_APP_JWT_TOKEN
+    : document.getElementsByName('authorized_token')[0].value;
+
+const axiosIls = axios.create({
```
zzacharo: I would rename that in $axios or $http. Also what about exporting in case we need it an other file?

invenio_app_ils/ui/backoffice/src/common/api.js
```diff
@@ -0,0 +1,27 @@
+import axios from 'axios';
+
+// In development create a personal token to be able to perform requests and
+// and put it in a .env file under the name `REACT_APP_JWT_TOKEN`
+const token =
```
zzacharo, line 14: this is gonna fail on initialization time on production  I think. For production I would export a function and set it manually from the request. So something like 

```

// In development create a personal token to be able to perform requests and
// and put it in a .env file under the name `REACT_APP_JWT_TOKEN`

const environment = process.env.NODE_ENV === 'production' ? 'production' : 'development';
const devToken = environment === 'development' ? process.env.REACT_APP_JWT_TOKEN : '';

const axiosIls = axios.create({
  baseURL:
    process.env.NODE_ENV === 'development'
      ? 'https://localhost:5000/api'
      : '/api',
  headers: {
    Authorization: `bearer ${token}`,
  },
  withCredentials: true,
});

setBearer() {
  let bearer = environment == 'production' ? (document.getElementsByName('authorized_token') || '') : devToken;
  if(bearer.length) {
    bearer = bearer[0].value;
    return {
      headers: {
        Authorization: `bearer ${bearer}`
      }
    }
  }
}
export const fetchRecord = (url, recid) => {
  return axiosIls.get(`${url}/${recid}`, setBearer());
};

export const postRecord = (url, data) => {
  return axiosIls.post(url, data, setBearer());
}
```

zzacharo, line 14: I am afraid this will fail on production because of the query on dom. I would do something similar to this.

```
// In development create a personal token to be able to perform requests and
// and put it in a .env file under the name `REACT_APP_JWT_TOKEN`

const environment = process.env.NODE_ENV === 'production' ? 'production' : 'development';
const devToken = environment === 'development' ? process.env.REACT_APP_JWT_TOKEN : '';

const axiosIls = axios.create({
  baseURL:
    process.env.NODE_ENV === 'development'
      ? 'https://localhost:5000/api'
      : '/api',
  withCredentials: true,
});

setBearer() {
  let bearer = environment == 'production' ? (document.getElementsByName('authorized_token') || '') : devToken;
  if(bearer.length) {
    bearer = bearer[0].value;
    return {
      headers: {
        Authorization: `bearer ${bearer}`
      }
    }
  }
}
export const fetchRecord = (url, recid) => {
  return axiosIls.get(`${url}/${recid}`, setBearer());
};

export const postRecord = (url, data) => {
  return axiosIls.post(url, data, setBearer());
};
```

zzacharo: Also probably we need to have in the common module a constants.js with all the global constants of our app such as the 
```
const environment = process.env.NODE_ENV === 'production' ? 'production' : 'development';
const devToken = environment === 'development' ? process.env.REACT_APP_JWT_TOKEN : '';
```

zzacharo (changes requested):

invenio_app_ils/ui/backoffice/src/common/api.js
topless: I was looking for a name, prefixing it with `$` didn't even cross my mind, cheers! Will export it too!

invenio_app_ils/ui/backoffice/src/common/api.js
```diff
@@ -0,0 +1,27 @@
...
+  process.env.NODE_ENV === 'development'
```
drjova: This might be useful ☺️  https://www.npmjs.com/package/dotenv

invenio_app_ils/ui/backoffice/src/App.js
```diff
@@ -2,21 +2,21 @@ import React, { Component } from 'react';
...
+import { ItemDetails } from './pages/ItemDetails';
+import NotFound from './common/components/NotFound';
 
 import './App.scss';
 
 export default class App extends Component {
   render() {
     return (
-      <Router basename="/backoffice">
+      <Router>
```
ntarocco, line 14: Didn't like the basename?

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/state/actions.js
```diff
@@ -5,7 +5,7 @@ import {
   LOAN_ACTION_SUCCESS,
   SET_LOAN_ACTION_ERROR,
 } from './types';
-import { fetchLoan, postLoan } from './api';
+import { fetchRecord, postRecord } from '../../../common/api';
```
ntarocco, line 8: To avoid the `../../../`, we could just add a `package.json` at the root, for example:
```
{
    "name": "backoffice"
}
```
And then:
`import { fetchRecord, postRecord } from 'backoffice/common/api';`

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/index.js
```diff
@@ -16,4 +17,4 @@ export default connect(
     error: state.loanDetails.error,
   }),
   mapDispatchToProps
-)(withRouter(LoanDetails));
+)(withRouter(detailsClass));
```
ntarocco: MINOR: `DetailsComponent` maybe?

invenio_app_ils/ui/backoffice/src/common/api.js
```diff
@@ -0,0 +1,36 @@
...
+const token =
+  process.env.NODE_ENV === 'production'
+    ? getProductionToken()
+    : process.env.REACT_APP_JWT_TOKEN;
+
+export const $axios = axios.create({
+  baseURL:
+    process.env.NODE_ENV === 'production'
+      ? '/api'
+      : 'https://127.0.0.1:5000/api',
```
ntarocco, line 23: I think your tests are now calling the a real url because you have changed everything from `localhost` to `127.0.0.1` but in the tests the mocks are still:
`const loansBaseUrl = 'https://localhost:5000/api/circulation/loans';`

Anyway, keep in mind that this baseUrls will disappear when we integrate the flask-js lib: https://github.com/inveniosoftware/invenio-app-ils/issues/41

ntarocco:
@topless the issue is happening because of the `axios.create`, see: https://stackoverflow.com/questions/51261781/write-test-axios-mock-adapter-with-axios-create

zzacharo:
@topless I think that you should remove the mock-adapter and just mock the fetch and post actions in api.

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/state/actions.js
topless: @ntarocco apart from the `package.json` with `name: backoffice` that exists in the root of backoffice? I tried a couple of things, but wasn't able to use it as an injection point.

invenio_app_ils/ui/backoffice/src/pages/LoanDetails/index.js
topless: Yes, I was looking for a name, thank you, a bit more explicit though, LoanDetailsComponent etc ..

invenio_app_ils/ui/backoffice/src/App.js
topless: We do, but there were some issues, where the backoffice was resolving to /, in the upcoming commits I am working with routes, so there is a possibility to be restored.

PR #40, opened 2018-10-20
records: models and schema changes

ntarocco:
* fixes schemas and mappings
* creates a common module to define PIDs
* creates Internal Location data model
* adds circulation status resolver
* adds item re-indexing on loan change via signals
* change demo data creation to be random

Blocked and Travis not happy because of: https://github.com/inveniosoftware/invenio-circulation/pull/75


invenio_app_ils/circulation/receivers.py
```diff
@@ -0,0 +1,29 @@
...
+    loan_state_changed.connect(index_item_after_loan_change,
+                               sender=app, weak=False)
+
+
+def index_item_after_loan_change(_, loan):
+    """Reindex item when attached loan changes."""
+    item_pid = loan.get(Item.pid_field)
+    if item_pid:
+        item = Item.get_record_by_pid(item_pid)
+        RecordIndexer().index(item)
```
ntarocco: Not a fan of harcoding `RecordIndexer` here because it could be overridden in config, we can discuss together if you have better ideas, what I have done in `circulation` to avoid this cannot be used here :(

invenio_app_ils/records/api.py
```diff
@@ -25,56 +33,66 @@ def get_record_by_pid(cls, pid, with_deleted=False):
...
-    def __init__(self, data, model=None):
-        """."""
-        super(Item, self).__init__(data, model)
-
     @classmethod
     def create(cls, data, id_=None, **kwargs):
         """Create Item record."""
-        data["$schema"] = current_jsonschemas.path_to_url(cls._schema)
+        data["circulation_status"] = {
+            "$ref": "{scheme}://{host}/api/circulation/items/{pid_value}/loan"
```
ntarocco, line 65: Another thing, this url is hardcoded, and it is the same used by the `jsonresolver` decorator in the `resolvers.py` module. Any idea how we put them in common? Maybe we can use a constant?

invenio_app_ils/cli.py
```diff
@@ -7,226 +7,261 @@
...
+        } for i in range(1, randint(10, 20))
+    ]
+    len_docs = len(documents) - 1
+    len_internal_locations = len(internal_locations) - 1
+    LEN_ITEM_CIRCULATION_RESTRICTIONS = len(ITEM_CIRCULATION_RESTRICTIONS) - 1
+    LEN_ITEM_MEDIUMS = len(ITEM_MEDIUMS) - 1
+    LEN_ITEM_STATUSES = len(ITEM_STATUSES) - 1
+
+    items = []
+    for i in range(1, randint(40, 50)):
```
ntarocco: I didn't want to parametrize these number as argument of the cli to avoid yet another config, but if you think it will be useful we could have something:
`ils demo data --docs 100 --items 2000 --loans 5000`

invenio_app_ils/circulation/receivers.py
ntarocco: Fixed

invenio_app_ils/cli.py
ntarocco: Added

zzacharo:
@ntarocco you should update the snapshots for the ui because you changed ItemDetails component. Then we should merge it because you fix circualation import error that causes all the other prs to fail.

invenio_app_ils/records/api.py
kpsherva: could we get the URL by url_for or similar?

kpsherva (approved):

PR #50, opened 2018-10-25
Refactored API module

### LINKED ISSUE #49: Back-office ui: api
Restructure api to separate concerns.

topless:
Api has been split in different modules. We support now `import { item } from 'common/api'` and then `item.getList()`, same for loans. They all share the same axios instance.

Closes #49


ntarocco (approved):
LGTM! But please fix the tests because they are failing.

kpsherva (approved):

PR #55, opened 2018-10-31
records: fix Record permission factory integration

### LINKED ISSUE #18: Support for restricted documents and items

kpsherva:
* fixes permission factory integration with invenio-access
  Permission class

* maps the request method of record allowed actions

* test cases for request methods for Item endpoint

(closes #18 )

To be merged after release of invenio-accounts (https://github.com/inveniosoftware/invenio-accounts/pull/271)


invenio_app_ils/records/permissions.py
```diff
@@ -21,26 +26,28 @@ class RecordPermission(Permission):
     - Delete access to admin and specified users.
     """
 
-    create_actions = ['create']
-    read_actions = ['read']
-    update_actions = ['update']
-    delete_actions = ['delete']
+    _actions = {'POST': 'create',
```
topless: Is it needed to remap the request method to custom actions?
Would it make sense to write `collect_needs` like:
```
if request.method is 'POST':
  do_stuff()
...
```

topless: I would start writing the object properties on new line like you do for `NEW_ITEM` in `test_item_permissions_endpoint.py`
```
_actions = {
  'POST': 'create,
...
}
```

kpsherva: yep, I was thinking about it too, but I decided to go with mapping per actions because:

- PUT and PATCH would require repeating the same array with permissions (unless we want to make it more flexible)
- request method names are not very human readable if someone does not know what do they mean, and I remember someone mentioning the 2nd line support being able to read the record's properties (I might be wrong about this one), but you are right and it is too CERN specific, it should be more generic

ntarocco (changes requested):
My opinion is to change this, but I also understand if someone does not agree, so open to discussion.

ntarocco (approved):

PR #60, opened 2018-11-08
global: fix setup when developing UI

ntarocco:
**WARNING** breaking change

This is to fix the setup when developing the UI.

1.  If you have one, you need to move the TOKEN to from `.env` to `.env.development`
2. You need to change the `config.py` and change the booleans on the vars documented in INSTALL.rst

You don't need anymore the extension to disable CORS.


invenio_app_ils/config.py
```diff
@@ -172,6 +172,13 @@ def _(x):
 #: Hostname used in URLs for local JSONSchemas.
 JSONSCHEMAS_HOST = "ils.mydomain.org"
 
+# CORS
+# ====
+REST_ENABLE_CORS = True
+# change this only while developing
+CORS_SEND_WILDCARD = True
+CORS_SUPPORTS_CREDENTIALS = False
```
topless, line 180: Should we make these part of the .env files?

topless (approved):

topless:
@ntarocco I just wanted to add a note on this particular one, cause I looked a bit into the CORS issue and I would personally like to thank you for resolving that hassle 🥇

zzacharo (approved):

PR #63, opened 2018-11-13
Ref resolvers

topless:
* Removed location_id from internal_location and internal_location_id from item
* Replaced with `$ref` the aforementioned ids.
* Resolvers for location inside internal_location, and for internal_location inside item.
* Moved responsibilities to https://github.com/inveniosoftware/invenio-app-ils/issues/72


invenio_app_ils/errors.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
```
ntarocco: Not needed anymore?

invenio_app_ils/records/api.py
```diff
@@ -62,12 +62,21 @@ class Item(IlsRecord):
...
+                scheme=current_app.config["JSONSCHEMAS_URL_SCHEME"],
+                host=current_app.config["JSONSCHEMAS_HOST"],
+                pid_value=data[cls.pid_field],
+            )
+        }
+        if InternalLocation.pid_field not in data:
+            raise IlsException("Internal Location pid is required")
+
+        data["internal_location"] = {
+            "$ref": "{scheme}://{host}/api/internal-locations/{pid_value}".format(
```
ntarocco: This works just because the host is different, otherwise it is exactly the same view registered by `invenio-records-rest`.
What about if we change this to something without the different host and:
`/api/refs/internal-locations/{pid-value}`
or
`/api/resolve/internal-locations/{pid-value}`

invenio_app_ils/records/jsonresolver/internal_location.py
```diff
@@ -0,0 +1,27 @@
...
+
+import jsonresolver
+from flask import current_app
+from invenio_pidstore.errors import PersistentIdentifierError
+
+from ..api import InternalLocation
+
+
+@jsonresolver.route(
+    "/api/internal-locations/<pid_value>", host="ils.mydomain.org"
```
ntarocco: I did the part the harcoding the host, but it is dangerous: if we override the config and change the host/scheme, this will brake.

Same for all the others

invenio_app_ils/records/loaders/schemas/json/internal_location.py
```diff
@@ -0,0 +1,21 @@
...
+"""Internal Location schema for marshmallow loader."""
+
+
+from marshmallow import Schema, fields
+
+
+class InternalLocationSchemaV1(Schema):
+    """Internal Location schema."""
+
+    location_pid = fields.Str(required=True)
```
ntarocco, line 18: About this, we have to fix `invenio-records-rest`. See: https://github.com/inveniosoftware/invenio-records-rest/issues/232

invenio_app_ils/records/jsonresolver/internal_location.py
zzacharo: I also think we should change this in order not to conflict with invenio-records-rest.

invenio_app_ils/records/jsonresolver/item.py
```diff
@@ -11,8 +11,9 @@
 from invenio_circulation.api import get_loan_for_item
 
 
-@jsonresolver.route('/api/circulation/items/<pid_value>/loan',
-                    host='ils.mydomain.org')
+@jsonresolver.route(
+    "/api/circulation/items/<pid_value>/loan", host="ils.mydomain.org"
```
zzacharo: Probably the host should be taken from config?

invenio_app_ils/records/jsonresolver/location.py
```diff
@@ -0,0 +1,25 @@
...
+"""Ils Location record resolver."""
+
+import jsonresolver
+from flask import current_app
+from invenio_pidstore.errors import PersistentIdentifierError
+
+from ..api import Location
+
+
+@jsonresolver.route("/api/locations/<pid_value>", host="ils.mydomain.org")
```
zzacharo: Same?

invenio_app_ils/records/jsonresolver/item.py
topless: @zzacharo If we try to load it from `current_app` it fails because it is out of request context and if we load it like `from config import ...` we loose the ability to override. I will create a ticker for this so we can handle it properly.

ntarocco: Please link the ticket in this PR, then I think we can approve and merge :)

ntarocco (approved):

zzacharo (approved):

PR #76, opened 2018-11-26
Backoffice: Display minimal user details

kpsherva:
How it looks now:
image
TODO: 
1. add active loans
2. add the extended metadata from invenio-userprofiles
3. add active loans list of a queried user

### DIFF
--- a/invenio_app_ils/ui/backoffice/src/common/api/user.js
+++ b/invenio_app_ils/ui/backoffice/src/common/api/user.js
@@ -1,0 +1,12 @@
+import { http } from './base';
+
+const userURL = '/users/';
+
+const get = userPid => {
+  return http.get(`${userURL}${userPid}`);
+};
+
+export const user = {
+  url: userURL,
+  get: get,
+};
--- a/invenio_app_ils/ui/backoffice/src/common/urls.js
+++ b/invenio_app_ils/ui/backoffice/src/common/urls.js
@@ -5,4 +5,5 @@ export const URLS = {
   itemDetails: id => `/items/${id}`,
   loansSearch: '/loans',
   loanDetails: id => `/loans/${id}`,
+  patronDetails: `/users/:userPid`,
 };
--- a/invenio_app_ils/ui/backoffice/src/pages/UserDetails/UserDetailsContainer.js
+++ b/invenio_app_ils/ui/backoffice/src/pages/UserDetails/UserDetailsContainer.js
@@ -1,0 +1,36 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Container } from 'semantic-ui-react';
+import { UserDetails } from './components';
+
+export default class UserDetailsContainer extends Component {
+  constructor(props) {
+    super(props);
+    this.fetchUserDetails = this.props.fetchUserDetails;
+  }
+
+  componentDidMount() {
+    this.unlisten = this.props.history.listen(location => {
+      if (location.state && location.state.userPid) {
+        this.fetchUserDetails(location.state.userPid);
+      }
+    });
+    this.fetchUserDetails(this.props.match.params.userPid);
+  }
+
+  componentWillUnmount() {
+    this.unlisten();
+  }
+
+  render() {
+    return (
+      <Container>
+        <UserDetails />
+      </Container>
+    );
+  }
+}
+
+UserDetailsContainer.propTypes = {
+  fetchUserDetails: PropTypes.func.isRequired,
+};
--- a/invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/UserMetadata/UserMetadata.js
+++ b/invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/UserMetadata/UserMetadata.js
@@ -1,0 +1,55 @@
+import React, { Component } from 'react';
+import { Grid, Segment, List, Header, Label } from 'semantic-ui-react';
+import PropTypes from 'prop-types';
+
+export default class UserMetadata extends Component {
+  render() {
+    const data = this.props.userDetails.data;
+
+    let label;
+    if (data && data.active) {
+      label = (
+        <Label as="a" color="green" ribbon>
+          Active
+        </Label>
+      );
+    } else {
+      label = (
+        <Label as="a" color="red" ribbon>
+          Deactivated
+        </Label>
+      );
+    }
+
+    return (
+      <Segment className="item-metadata">
+        {label}
+        <Grid padded columns={2}>
+          <Grid.Column width={10}>
+            <Header as="h1">Patron - {data.id}</Header>
+          </Grid.Column>
+          <Grid.Column>
+            <List relaxed size="large">
+              <List.Item>
+                <List.Header>Patron PID</List.Header>
+                {data.id}
+              </List.Item>
+              <List.Item>
+                <List.Header>Email</List.Header>
+                {data.email}
+              </List.Item>
+              <List.Item>
+                <List.Header>Active</List.Header>
+                {data.active.toString()}
+              </List.Item>
+            </List>
+          </Grid.Column>
+        </Grid>
+      </Segment>
+    );
+  }
+}
+
+UserMetadata.propTypes = {
+  userDetails: PropTypes.object.isRequired,
+};
--- a/invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/__tests__/actions.js
+++ b/invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/__tests__/actions.js
@@ -3,13 +3,13 @@ import thunk from 'redux-thunk';
 import * as actions from '../actions';
 import { initialState } from '../reducer';
 import * as types from '../types';
-import { item as itemApi } from 'common/api';
+import { user as userApi } from 'common/api';
 
 const middlewares = [thunk];
 const mockStore = configureMockStore(middlewares);
 
 const mockGet = jest.fn();
-itemApi.get = mockGet;
+userApi.get = mockGet;
 
 const response = { data: {} };
 const expectedPayload = {};
@@ -22,9 +22,9 @@ beforeEach(() => {
   store.clearActions();
 });
 
-describe('Item details actions', () => {
-  describe('Fetch item details tests', () => {
-    it('should dispatch an action when fetching an item', done => {
+describe('User details actions', () => {
+  describe('Fetch user details tests', () => {
+    it('should dispatch an action when fetching an user', done => {
       mockGet.mockResolvedValue(response);
 
       const expectedActions = [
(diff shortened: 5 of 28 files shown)


invenio_app_ils/ui/backoffice/src/common/api/user.js
```diff
@@ -0,0 +1,12 @@
...
+
+const userURL = '/users/';
+
+const getRecord = userPid => {
+  return http.get(`${userURL}${userPid}`);
+};
+
+export const user = {
+  url: userURL,
+  getRecord: getRecord,
```
ntarocco: it is not a record, it might be misleading.
I would also change the other to have `get` only or `getLoan`, the react app doesn't care if it is a record... but I let @topless comment on this 😅

invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/UserMetadata/UserMetadata.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default class UserMetadata extends Component {
+  render() {
+    const data = this.props.userDetails.data;
+
+    let label;
+    if (data.active) {
+      label = (
+        <Label as="a" color="green" ribbon>
+          {' '}
```
ntarocco: css instead? I personally find `{' '}` ugly, but maybe it is just me :)

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/__tests__/actions.js
```diff
@@ -0,0 +1,87 @@
...
+    store.clearActions();
+  });
+
+  describe('Fetch users details tests', () => {
+    let mock;
+    let response = {};
+    beforeEach(() => {
+      mock = new MockAdapter(http);
+      mock.onGet(`${usersBaseUrl}/1`).reply(() => {
+        return new Promise((resolve, reject) =>
```
ntarocco: I would really avoid to do this, I thought I have already removed it everywhere.
I would suggest to mock the `getRecord` method and not the http call, it is much easier. On top of that, Redux action does not know that there is axios behind, it knows that calling the method `getRecord` it will receive a Promise.

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/actions.js
```diff
@@ -0,0 +1,25 @@
+import { IS_LOADING, SUCCESS, HAS_ERROR } from './types';
+import { user } from 'common/api';
+
+export const fetchUserDetails = userPid => {
```
ntarocco, line 4: `userPid` -> `userId`? Sorry if I am picky, but I struggled a lot when learning Invenio to distinguish what is a PID and what is an ID, and I would personally prefer we call them with the right name to avoid confusion.

invenio_app_ils/ui/backoffice/src/common/urls.js
```diff
@@ -5,4 +5,6 @@ export const URLS = {
   itemDetails: id => `/items/${id}`,
   loansSearch: '/loans',
   loanDetails: id => `/loans/${id}`,
+  patronsSearch: '/users',
```
ntarocco: does this exist?

invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/UserMetadata/UserMetadata.js
```diff
@@ -0,0 +1,57 @@
...
+          Active{' '}
+        </Label>
+      );
+    } else {
+      label = (
+        <Label as="a" color="red" ribbon>
+          {' '}
+          Desactivated{' '}
```
zzacharo: typo :)

invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/UserMetadata/UserMetadata.js
zzacharo: Why is that `{' '}`?

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/__tests__/reducer.js
```diff
@@ -0,0 +1,45 @@
...
+
+  it('should change error state on error action', () => {
+    const action = {
+      type: types.HAS_ERROR,
+      payload: 'Error',
+    };
+    expect(reducer(initialState, action)).toEqual({
+      ...initialState,
+      isLoading: false,
+      data: 'Error',
```
zzacharo: I wouldn't mix the data type to be both an object or string, just for consistency and to know what we expect.

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/reducer.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default (state = initialState, action) => {
+  switch (action.type) {
+    case IS_LOADING:
+      return { ...state, isLoading: true };
+    case SUCCESS:
+      return {
+        ...state,
+        isLoading: false,
+        data: action.payload,
```
zzacharo, line 17: I think `data` should be null as `hasError` is the responsible for the error state. Maybe I would add an `error` state if we want to show the errorred response. But I don't think that the response is quite explanatory in our case.

zzacharo (changes requested):

invenio_app_ils/ui/backoffice/src/common/api/user.js
kpsherva: @zzacharo true, it feels more intuitive. Should I change the other api calls for Loans and Items too ? @ntarocco

invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/UserMetadata/UserMetadata.js
kpsherva: thanks, I've missed this one somehow

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/actions.js
kpsherva: 👍 I can rename it, but it was represented as Pid before in the code, maybe I missed something

invenio_app_ils/ui/backoffice/src/common/urls.js
kpsherva: no it does not, but I think it would make sense also to have users' search, WDYT?

invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/UserMetadata/UserMetadata.js
kpsherva: thanks!

invenio_app_ils/ui/backoffice/src/pages/UserDetails/UserDetailsContainer.js
```diff
@@ -0,0 +1,36 @@
...
+import { Container } from 'semantic-ui-react';
+import { UserDetails } from './components';
+
+export default class UserDetailsContainer extends Component {
+  constructor(props) {
+    super(props);
+    this.fetchUserDetails = this.props.fetchUserDetails;
+  }
+
+  componentDidMount() {
```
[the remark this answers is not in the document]
kpsherva: I kept it consistent with the rest of the pages, so I guess changing it will be connected with also changing the rest.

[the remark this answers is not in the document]
zzacharo: No, it's fine for now but if you want just open a ticket to address it in the upcoming sprints :)

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/__tests__/actions.js
ntarocco: I still dislike this very much 😄, IMHO overcomplicated

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/reducer.js
ntarocco: Wait, what is happening here?
Until now, what has been done by @topless was to have:
```
isLoading: boolean
data: contains result payload on success or error payload on error
hasError: boolean
```
If we change it, we change it everywhere.
Correct me if I am wrong please.

invenio_app_ils/ui/backoffice/src/common/api/user.js
topless: @ntarocco the shorter the better, @zzacharo was faster on this one, that is the for I would use to get a user and every other model.

invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/UserMetadata/UserMetadata.js
topless: Nope its not just you, `{'  '}`  is just ugly :)

invenio_app_ils/ui/backoffice/src/common/urls.js
topless: It depends on the case, if librarians need to navigate in users to perform actions on their behalf maybe would make sense. If it just adds complexity, maybe we can just pop up a searchable list of users when it is needed.

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/reducer.js
topless: @ntarocco yes that was the flow till now, the component will enter error state or not according to `hasError` flag. An approach closer to what @zzacharo suggests would be to have 

```
{
  ...
  data: {title: '...', ...} or null if it has an error,
  error: {msg: '..', status:' ...'} or null if it doesn't have an error,
  ...
}
```
Like this we don't depend on Booleans to figure it out but the object themselves. I am absolutely fine with both or similar approaches.

kpsherva: for now I kept it consistent with the rest of the cases, if we want to change it we should change it in all the places, shall we discuss ?

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/reducer.js
```diff
@@ -0,0 +1,31 @@
+import { IS_LOADING, SUCCESS, HAS_ERROR } from './types';
+
+export const initialState = {
+  isLoading: true,
+  hasError: false,
+  data: {},
+  error: {},
```
ntarocco: Oups! :relaxed:

ntarocco (approved):

invenio_app_ils/ui/backoffice/src/pages/UserDetails/state/reducer.js
kpsherva: ups :flushed:

zzacharo (approved):

PR #84, opened 2018-12-03
WIP: Users loan

### LINKED ISSUE #23: Back-office UI: Librarian can see patron information
* account info
* past and current loans

depends on #71

kpsherva:
(closes #23 )


invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/PatronLoansTable/PatronLoansTable.js
```diff
@@ -0,0 +1,245 @@
+import React, { Component } from 'react';
```
ntarocco, line 1: I am sorry, I am a bit confused. It looks that this component would like to be generic (accepting an optional `patron` or `loanState` as props for example), but in reality it is very specific, it is for patron loans and cannot be reused because it is connected to a specific part of the state.

I think that what we wanted to do was to have a generic loans component for presentation only, but the wrapper is specific. For example:
```
<PatronLoans>
    <LoansTable
        headers={[]}
        loans={[]}
        onSortByChangeHandler={func}
        onSortOrderChangeHandler={func}
    />
<PatronLoans>
```

If we don't what to do it now, then there is no point to have this optional props.

invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/PatronLoansTable/state/selectors.js
```diff
@@ -0,0 +1,19 @@
+import _isEmpty from 'lodash/isEmpty';
```
ntarocco, line 1: Might be better to have this reusable to avoid duplication? As response serialization for any loan maybe?

invenio_app_ils/ui/backoffice/src/store.js
```diff
@@ -14,6 +15,7 @@ const rootReducer = combineReducers({
   userSession: userSessionReducer,
   itemDetails: itemDetailsReducer,
   itemPendingLoans: itemPendingLoansReducer,
+  patronLoansTable: patronLoansTableReducer,
```
ntarocco, line 18: `Table` is a presentation concept, what about just `patronLoans`?

invenio_app_ils/ui/backoffice/src/routes/index.js
```diff
@@ -1,2 +1,3 @@
 export { ItemRoutes } from './ItemRoutes';
 export { LoanRoutes } from './LoanRoutes';
+export { UserRoutes } from './UserRoutes';
```
zzacharo, line 3: Why we should have all these Route components? The `Switch` component here is redundant I think.

invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/PatronLoansTable/PatronLoansTable.js
```diff
@@ -0,0 +1,245 @@
...
+        <Table.Cell>{startDate}</Table.Cell>
+        <Table.Cell>{endDate}</Table.Cell>
+      </Table.Row>
+    );
+  };
+
+  _handleShowAllClick = () => {
+    const { document_pid, item_pid } = this.props.item
+      ? this.props.item.metadata
+      : this.initProps;
```
zzacharo, line 84: InitialProps maybe? InitProps sounds like a function to me...

invenio_app_ils/ui/backoffice/src/pages/UserDetails/components/PatronLoansTable/PatronLoansTable.js
```diff
@@ -0,0 +1,245 @@
...
+      this.patronLoansChangeSortBy(
+        document_pid,
+        item_pid,
+        loan_state,
+        patron_pid
+      );
+    }
+  };
+
+  _renderLoans = loans => {
```
zzacharo, line 139: I think that the reusable stuff should be only the way that the results will be presented...So this component just expects some props in order to render the results...I will agree with @ntarocco on this one that redux should come from the component that will use this one along with the props needed.

invenio_app_ils/ui/backoffice/src/routes/index.js
ntarocco: Fixed by me in the next refactoring to merge everything in one app.

ntarocco:
Created related ticket: https://github.com/inveniosoftware/invenio-app-ils/issues/91

PR #89, opened 2018-12-06
Openshift infrastructure

zzacharo:


invenio_app_ils/config.py
```diff
@@ -90,6 +91,38 @@ def _(x):
...
+def _parse_env_bool(var_name, default=None):
+    if str(os.environ.get(var_name)).lower() == 'true':
+        return True
+    elif str(os.environ.get(var_name)).lower() == 'false':
+        return False
+    return default
+
+# Search
+# ======
+ELASTICSEARCH_HOST = os.environ.get('ELASTICSEARCH_HOST', 'localhost')
```
lnielsen, line 103: This is already taken care of by Invenio-Config? It's just that you have to name your env variables ``INVENIO_ELASTICSEARCH_HOST``. Then you could reduce this one to:

``ELASTICSEARCH_HOST = 'localhost'``

zzacharo: @lnielsen Yes I know, but is it the case for the other ones too, like `ELASTICSEARCH_USE_SSL` that I want to parse it inside the config? I thought just to be consistent to have everything defined like that but you are right that it can be simplified.

slint: The problem is that any env-originated configuration has to be passed directly to [`(INVENIO_)SEARCH_ELASTIC_HOSTS`](https://github.com/inveniosoftware/invenio-search/blob/6f49bf98ddb19afd0f5f7f5121060c8d46925b47/invenio_search/config.py#L15). If you have a more complex configuration, like e.g. a password for basic auth, or multiple hosts (if you have a multi-node cluster), the only option via `INVENIO_` variables, is to have some sort of really long Python AST-parsable string, which would be error-prone to build and have as part of your OpenShift secrets (since you want to also have a password in there somwhere). So the options are:

```shell
# Generated env variable to store in OpenShift/Kubernetes ConfigMap.
# This has to be valid Python-parsable code 
export INVENIO_SEARCH_ELASTIC_HOSTS="
[{'host': 'es-node-c1', 'port': 443, 'url_prefix': 'es', 'use_ssl': True, 'verify_certs': False, 'http_auth': ('myuser', '*************')}, 
{'host': 'es-node-c2', 'port': 443, 'url_prefix': 'es', 'use_ssl': True, 'verify_certs': False, 'http_auth': ('myuser', '*************')}, 
{'host': 'es-node-c3', 'port': 443, 'url_prefix': 'es', 'use_ssl': True, 'verify_certs': False, 'http_auth': ('myuser', '*************')}]"
```

vs.

```python
# export ES_HOSTS="es-node-c1 es-node-c2 es-node-c3"
ES_HOSTS = os.environ.get('ES_HOSTS', 'localhost').split()
ES_PORT = int(os.environ.get('ES_PORT', '9200'))
ES_USER = os.environ.get('ES_USER')
ES_PASSWORD = os.environ.get('ES_PASSWORD')
ES_URL_PREFIX = os.environ.get('ES_URL_PREFIX', '')
ES_USE_SSL = _parse_env_bool('ES_USE_SSL')
ES_VERIFY_CERTS = _parse_env_bool('ES_VERIFY_CERTS')

es_host_params = {'port': ES_PORT}
if ES_USER and ES_PASSWORD:
    es_host_params['http_auth'] = (ES_USER, ES_PASSWORD)
if ES_URL_PREFIX:
    es_host_params['url_prefix'] = ES_URL_PREFIX
if ES_USE_SSL is not None:
    es_host_params['use_ssl'] = ES_USE_SSL
if ES_VERIFY_CERTS is not None:
    es_host_params['verify_certs'] = ES_VERIFY_CERTS

SEARCH_ELASTIC_HOSTS = [{'host': host, **es_host_params} for host in ES_HOSTS]
```

We should document these practices and possibly provide some helpers in `invenio-search` to make all of this a bit easier.

zzacharo:
@lnielsen @slint Thank you a lot for your comments. We decided to close this pr and whenever we integrate ils in cds-books to put the configuration there. I will update an issue regarding the elasticsearch configuration in openshift that we have in order to address your comments.

PR #96, opened 2018-12-21
Librarian perform loan extend or return

### LINKED ISSUE #21: Back-office UI:Librarian performs loan return or extend

zzacharo:
closes #21

### DIFF
--- a/invenio_app_ils/cli.py
+++ b/invenio_app_ils/cli.py
@@ -139,7 +139,7 @@ def get_loans_for_items(
         if status == "ITEM_ON_LOAN":
             items_on_loans.append(item[Item.pid_field])
 
-        padron_id = randint(1, len_patron_ids)
+        patron_id = randint(1, len_patron_ids)
         transaction_date = datetime(
             current_year, randint(1, 12), randint(1, 28)
         )
@@ -147,8 +147,7 @@ def get_loans_for_items(
 
         loan = {
             Loan.pid_field: "{}".format(i),
-            Item.pid_field: "{}".format(item[Item.pid_field]),
-            "patron_pid": "{}".format(padron_id),
+            "patron_pid": "{}".format(patron_id),
             "state": "{}".format(status),
             "transaction_date": transaction_date.strftime("%Y-%m-%d"),
             "transaction_location_pid": "{}".format(loc_pid),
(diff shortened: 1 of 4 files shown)


invenio_app_ils/cli.py
```diff
@@ -157,6 +156,11 @@ def _get_valid_status(item, items_on_loans):
             "request_expire_date": expire_date.strftime("%Y-%m-%d"),
         }
 
+        if status == "PENDING":
+            loan[Document.pid_field] = "{}".format(item[Document.pid_field])
```
topless, line 160: We need the `Document.pid_field` in every case not only when its `PENDING`

invenio_app_ils/config.py
```diff
@@ -50,11 +50,8 @@
 from invenio_circulation.transitions.transitions import (  # isort:skip
     ToItemOnLoan,
     CreatedToPending,
-    ItemInTransitHouseToItemReturned,
-    ItemOnLoanToItemInTransitHouse,
     ItemOnLoanToItemReturned,
-    PendingToItemAtDesk,
-    PendingToItemInTransitPickup,
+    ItemOnLoanToItemOnLoan,
```
topless, line 54: I understand the state-to-state naming, but I think `ExtendItemOnLoan` would suffice, what happens internally with the states shouldn't bother the names.

topless (commented):
Minor comments, can be omitted of course.

invenio_app_ils/config.py
zzacharo: but this one is on circulation side...create a ticket about that so we can handle it :)

invenio_app_ils/cli.py
zzacharo: Yes but right now is not supported in invenio-circulation so to be consistent I din't include it here. I created a ticket about that to handle it https://github.com/inveniosoftware/invenio-circulation/issues/85

topless: 👍 for this one, I am about to address also the issue of the demo loan data and how they reflect reality. We need this one.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoansSearch/LoansSearch.js
```diff
@@ -34,21 +33,6 @@ import {
 import './LoansSearch.scss';
 import { default as config } from './config';
 
-const resultsPerPageValues = [
```
ntarocco, line 37: Thanks, I missed this one!

ntarocco (approved):

kpsherva (approved):

PR #113, opened 2019-01-16
items: reference document from item record

zzacharo:


invenio_app_ils/cli.py
```diff
@@ -92,6 +92,18 @@ def get_documents_items(internal_locations, n_docs, n_items):
             "medium": "{}".format(ITEM_MEDIUMS[randint(0, LEN_ITEM_MEDIUMS)]),
             "status": "{}".format(ITEM_STATUSES[randint(0, LEN_ITEM_STATUSES)]),
         }
+        # import ipdb; ipdb.set_trace()
```
ntarocco: oups!

invenio_app_ils/cli.py
```diff
@@ -92,6 +92,18 @@ def get_documents_items(internal_locations, n_docs, n_items):
...
+
+        item["document"] = {
+            "$ref": "{scheme}://{host}/api/resolver/items/document/{pid_value}"
+            .format(
+                scheme="https",
+                host="127.0.0.1:5000",
```
ntarocco: I guess at some point we will decide to replace this with some other fake domain name and we need to align it with the config `JSONSCHEMA_HOST`.

invenio_app_ils/records/api.py
```diff
@@ -76,6 +76,17 @@ def create(cls, data, id_=None, **kwargs):
         return super(Item, cls).create(data, id_=id_, **kwargs)
 
 
+    def attach_document(self, document_pid, **kwargs):
+        """."""
```
ntarocco: Add the comment so we don't have to another pass later?

ntarocco (commented):
Travis is not happy, some tests are failing :(

invenio_app_ils/cli.py
zzacharo: I will point to the config then!

invenio_app_ils/records/api.py
```diff
@@ -76,6 +76,17 @@ def create(cls, data, id_=None, **kwargs):
         return super(Item, cls).create(data, id_=id_, **kwargs)
 
 
+    def attach_document(self, document_pid, **kwargs):
+        """Attach a document ref to the record."""
+        self["document"] = {
+            "$ref": "{scheme}://{host}/api/resolver/items/document/<pid_value>".format(
+                scheme=current_app.config["JSONSCHEMAS_URL_SCHEME"],
+                host=current_app.config["JSONSCHEMAS_HOST"],
+                pid_value=document_pid,
```
kpsherva, line 85: this has no corresponding format value (pid_value)

invenio_app_ils/records/api.py
```diff
@@ -76,6 +76,17 @@ def create(cls, data, id_=None, **kwargs):
         return super(Item, cls).create(data, id_=id_, **kwargs)
 
 
+    def attach_document(self, document_pid, **kwargs):
+        """Attach a document ref to the record."""
+        self["document"] = {
+            "$ref": "{scheme}://{host}/api/resolver/items/document/<pid_value>".format(
```
kpsherva: Maybe we should have this url in some variable, because it's repeated

invenio_app_ils/records/api.py
zzacharo: oups you are right!

invenio_app_ils/records/api.py
zzacharo: It's not fully repeated only the `api/resolver/` part, but creating only for this now I am not sure...Too many config variables after WDYT?

invenio_app_ils/records/api.py
zzacharo: @kprzerwa fixed!

ntarocco (approved):

PR #120, opened 2019-01-22
resolver: add circulation info in document

zzacharo:


invenio_app_ils/records/jsonresolver/circulation.py
```diff
@@ -0,0 +1,64 @@
...
+        loanable_items = 0
+        number_of_items = 0
+        items = circulation_items_retriever(document_pid)
+
+        for item in items:
+            number_of_items += 1
+            if not is_item_available(item):
+                active_loans += 1
+
+            if item["state"] is "LOANABLE":
```
ntarocco: you can use `circulation_is_item_available` method (to be renamed in the future)

ntarocco (dismissed):

invenio_app_ils/circulation/receivers.py
```diff
@@ -12,18 +12,23 @@
...
 
-def index_item_after_loan_change(_, loan):
+def index_record_after_loan_change(_, loan):
     """Reindex item when attached loan changes."""
     item_pid = loan.get(Item.pid_field)
     if item_pid:
         item = Item.get_record_by_pid(item_pid)
         current_app_ils_extension.item_indexer.index(item)
+
+    document_pid = loan.get(Document.pid_field)
```
topless, line 31: I am not sure this is going to work. I noticed that there is no document_pid displayed on the loan list. Looked into the loans api response and the document_pid appears only inside the `document` property which gets resolved. It might is a duplication having the `document_pid` in `loan.metadata` and inside the `loan.metadata.document` but I think we need it there. I also noticed that in cli the `document_pid` was removed on the creation of loans.

invenio_app_ils/circulation/receivers.py
```diff
@@ -12,18 +12,23 @@
...
 
 def register_circulation_signals(app):
     """Register Circulation signal."""
-    loan_state_changed.connect(index_item_after_loan_change,
+    loan_state_changed.connect(index_record_after_loan_change,
                                sender=app, weak=False)
 
 
-def index_item_after_loan_change(_, loan):
+def index_record_after_loan_change(_, loan):
```
topless, line 24: This is more food for thought than a request. We will possibly have to index more things than item and document and we will have more cases that we need to trigger manual indexing. My proposal would be each model to have an individual function 

`def index_item(data):...`
`def index_doc(data):...`

and then compose the indexers according to the case:
```
def index_record_after_loan_change(_, loan):
    index_item(loan)
    index_doc(loan)
    ...
```

invenio_app_ils/search/api.py
```diff
@@ -10,6 +10,12 @@
 from invenio_search.api import RecordsSearch
 
 
+class IlsSearch(RecordsSearch):
+    """."""
```
ntarocco: add comment please

invenio_app_ils/circulation/receivers.py
zzacharo: Yes but in this particular case here is just the indexing after the loan signal fired. For the other cases I will create a specific indexer per record type (e.g item, document) which will be responsible for the indexing and all the side effects.

invenio_app_ils/records/api.py
```diff
@@ -55,31 +70,40 @@ class Item(IlsRecord):
...
+    _document_resolving_path = (
+        "{scheme}://{host}/api/resolver/items/document/{pid_value}"
+    )
 
     @classmethod
     def create(cls, data, id_=None, **kwargs):
         """Create Item record."""
         data["circulation_status"] = {
-            "$ref": "{scheme}://{host}/api/resolver/circulation/items/{pid_value}/loan".format(
+            "$ref": cls._loan_resolving_path.format(
```
topless: This is very nice, extracting it to class properties; 
Irrelevant with programming, in English language -ing indicates duration in action, I would prefer `*_resolver_path` or `*_resolve_path`.

invenio_app_ils/records/jsonresolver/circulation.py
```diff
@@ -0,0 +1,60 @@
...
+def jsonresolver_loader(url_map):
+    """Resolve the circulation for document record."""
+    from flask import current_app
+
+    def _circulation_for_document_resolver(document_pid):
+        """Return circulation info for the given document."""
+        circulation = {}
+        active_loans = 0
+        loanable_items = 0
+        number_of_items = 0
```
topless: Personal convention I follow, whenever I have to count something I postfix it with _count in order avoid numbers_of, num_of etc... in the particular example `items_count`.

tests/data/documents.json
```diff
@@ -3,35 +3,30 @@
...
-    "items": []
+    "abstracts": ["This is an abstract"]
   },
   {
     "document_pid": "docid-5",
     "title": "Less: A Novel",
     "authors": "Andrew Sean Greer",
-    "abstracts": ["This is an abstract"],
-    "items": ["itemid-7", "itemid-8", "itemid-9", "itemid-10"]
+    "abstracts": ["This is an abstract"]
```
topless, line 30: We got it, this is an abstract.. :)

topless (dismissed):

invenio_app_ils/circulation/utils.py
```diff
@@ -12,28 +12,26 @@
...
-from ..records.api import Document, Item
+from ..records.api import Item
+from ..search.api import ItemSearch
 
 
 def circulation_build_item_ref(item_pid):
     """Build $ref for item."""
     return {
-        "$ref": "{scheme}://{host}{item_resolving_path}{item_pid}".format(
+        "$ref": "{scheme}://{host}/api/resolver/circulation/items/{item_pid}".format(
```
kpsherva, line 22: why do we change it back to hardcoded url? 😢

zzacharo: because we don't want to expose it as it can lead to inconsistencies :)

ntarocco (approved):

Issue #112, opened 2019-01-16
backoffice: endpoint for loan update request


topless:
We we want to change an item on loan with another one, its `$ref` manipulation should happen from the backend and not manually on the request.

PR #131, opened 2019-01-24
backend: Loan update endpoint

topless:
Requires [pr#89](https://github.com/inveniosoftware/invenio-circulation/pull/89) from circulation to be merged first.

* update a loan from circulation backend
* receiver for signal to index changed item
* switched patch to post request
* fixed missing document_pid on loan list
* addresses #112

### DIFF
--- a/invenio_app_ils/circulation/receivers.py
+++ b/invenio_app_ils/circulation/receivers.py
@@ -9,7 +9,7 @@
 
 from __future__ import absolute_import, print_function
 
-from invenio_circulation.signals import loan_state_changed
+from invenio_circulation.signals import loan_replace_item, loan_state_changed
 
 from ..proxies import current_app_ils_extension
 from ..records.api import Document, Item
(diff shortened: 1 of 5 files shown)


invenio_app_ils/circulation/receivers.py
```diff
@@ -9,16 +9,30 @@
...
+    loan_item_changed.connect(
+        index_old_item_after_loan_update, sender=app, weak=False
+    )
+
+
+def index_old_item_after_loan_update(_, item_pid):
+    """Register Circulation signal to index item."""
+    import ipdb
+
+    ipdb.set_trace()
```
ntarocco: oups! :)

invenio_app_ils/ui/src/invenio_app_ils/common/api/loan.js
```diff
@@ -3,10 +3,8 @@ import { toISO } from './date';
 import { DateTime } from 'luxon';
 
 const loanURL = '/circulation/loans/';
-
-const get = loanPid => {
-  return http.get(`${loanURL}${loanPid}`);
-};
+const get = loanPid => http.get(`${loanURL}${loanPid}`);
+const getPatchUrl = loanPid => `${loanURL}/${loanPid}/update`;
```
ntarocco: What about if we get rid of the name `Patch`? not it is a POST.
`replaceItem`?

ntarocco (changes requested):

invenio_app_ils/circulation/receivers.py
topless: Right oups..

invenio_app_ils/ui/src/invenio_app_ils/common/api/loan.js
topless: I have refactored it in the circulation thing, and changed it to `LoanUpdate`, because thats what it does, and it could be used further, rather than just replacing the item. Your thoughts (and prayers) please.

zzacharo (dismissed):

kpsherva (commented):
I think the same as in PR https://github.com/inveniosoftware/invenio-app-ils/pull/89 that the naming should be more specific as the custom view is implemented for item replace action

ntarocco (approved):

PR #138, opened 2019-01-28
resolvers: refactor route implementation

### LINKED ISSUE #134: resolver: change all hardcoded resolver urls
- [x] Apply the principle that we use here https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/records/jsonresolver/circulation.py#L20 to all other resolvers.

- [x] Replace all hardcoded 5 seconds time for indexing tasks with a config variable that can we can override runtime.

zzacharo:
closes #134 
- Refactors resolvers to use the `jsonresolver.hookimpl` decorator and get rid of the hardcoded localhost in the route
- Make the indexers tasks delay configurable


ntarocco (approved):
Fantastic thanks!

kpsherva (approved):

Issue #122, opened 2019-01-22
refactor api list and get functions 


kpsherva:
For another refactoring, not related to your changes, what about have this nicer and use  a Builder pattern?
Like this you would have something like:
```
const loansFetcher = 
    LoansFetcher(documentPid)
        .withState('PENDING')
        .build();

await loansFetcher.fetch().then()....
```
In general it is a nice way to handle optional params.
It is also valid for Items and Documents (ping @zzacharo @topless if they have also thoughts about this)

_Originally posted by @ntarocco in https://github.com/inveniosoftware/invenio-app-ils/pull/121_

kpsherva:
finally decided to have predefined functions to build necessary queries, without query builder

PR #140, opened 2019-01-28
fix: state queries and document_pid

topless:
* fixed queries for state and multiple states for loanApi fetchLoans method
* fixed queries document_pid which was missing from the metadata of the record but exists in the resolved property of document
* fixed missing document_pid from items search


ntarocco (dismissed):
I agree that this is all very ugly and needs to be fixed.

zzacharo (dismissed):

kpsherva:
> I agree that this is all very ugly and needs to be fixed.

It is already planned in the issue: https://github.com/inveniosoftware/invenio-app-ils/issues/122

kpsherva (dismissed):

ntarocco (approved):

PR #142, opened 2019-01-29
circulation: fix create new loan request

zzacharo:
The ui wasn't sending the `document_pid` when it was requesting a new loan for an item and thus the indexer was assuming that was there. I made it optional when the indexer tries to index the document after the loan is changed and also tried to extract the document from the item when the `create_loan` api was being called.

### DIFF
--- a/invenio_app_ils/circulation/api.py
+++ b/invenio_app_ils/circulation/api.py
@@ -14,6 +14,8 @@ from invenio_circulation.errors import CirculationException
 from invenio_circulation.pidstore.minters import loan_pid_minter
 from invenio_circulation.proxies import current_circulation
 
+from invenio_app_ils.circulation.utils import circulation_document_retriever
+
 
 def request_loan(params):
     """Create a loan and trigger the first transition to create a request."""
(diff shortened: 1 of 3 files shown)


invenio_app_ils/circulation/utils.py
```diff
@@ -38,7 +38,9 @@ def circulation_document_retriever(item_pid):
     """Retrieve items given a document."""
```
topless: The description does not match with what it actually returns

topless (changes requested):

kpsherva (dismissed):

invenio_app_ils/circulation/utils.py
```diff
@@ -35,10 +35,12 @@ def circulation_items_retriever(document_pid):
 
 
 def circulation_document_retriever(item_pid):
-    """Retrieve items given a document."""
+    """Retrieves document given an item."""
     item = Item.get_record_by_pid(item_pid)
     item = item.replace_refs()
-    return item["document"]["document_pid"]
+    if item and item.get("document") and item["document"].get("document_pid"):
```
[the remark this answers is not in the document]
zzacharo: true I will change it ;)

topless (approved):

invenio_app_ils/circulation/api.py
```diff
@@ -51,6 +53,10 @@ def create_loan(params):
         raise CirculationException(
             "Patron has already a request or active loan on this item.")
 
+    if "document_pid" not in params:
+        document_pid = circulation_document_retriever(params["item_pid"])
+        if document_pid:
+            params["document_pid"] = document_pid
```
ntarocco, line 59: should we actually fail if `None`?

ntarocco (approved):
Thank you for fixing this!

PR #151, opened 2019-02-04
Refactored serializers

### LINKED ISSUE #123: ui: refactor serialisers in the components
the data obtained by GET in the components have inconsistent serialisation - sometimes it is serialised sometimes not, it needs refactoring (see comments in https://github.com/inveniosoftware/invenio-app-ils/pull/121) also remember about checking if data exists (when the field is not required in the schema)

kpsherva:
( closes #123 )


invenio_app_ils/ui/src/invenio_app_ils/common/api/loans/serializer.js
```diff
@@ -0,0 +1,42 @@
+import _isEmpty from 'lodash/isEmpty';
+import { fromISO, toShortDateTime } from '../date';
+
+function serializeResponse(hit) {
+  const result = {};
```
zzacharo: I would propose in general to do something like `let result = {...hit.metadata}` and then in then serilizer just handle only the fields that need some special check or rename. WDYT?

kpsherva: done!

ntarocco (commented):
@kprzerwa  @zzacharo  @topless what was the outcome of your discussion about flattening that `metadata` object?
It was not bothering me much, but I agree with what you have decided together!

My comment is just that I would not have exposed serializers. The domain layer (redux, our state), shouldn't probably know about the API layer.
Document, for example, I would have done:
```
const get = documentPid => {
  return http
    .get(`${documentURL}${documentPid}`)
    .then(response -> fromJSON(response));
};
```
Like this, we also avoid to call serializers everywhere in redux.

Similarly, for tests, `__tests__/document.js`:
```
mockResponse = {
      id: '123',
      updated: stringDate,
      created: stringDate,
      metadata: {},
    }
// mock get() method to return mockResponse, and let the serializer to map the mocked response
model = documentApi.get()
expect(model).toEqual({
    document_pid: '123',
    created: fromISO(stringDate),
    updated: fromISO(stringDate),
})
```

kpsherva:
done, serialisers are not exposed to redux anymore

ntarocco (dismissed):

zzacharo (dismissed):

PR #158, opened 2019-02-07
ui: backoffice fix assign item api call URL

kpsherva:
* moves function to correct folder
* fixes api call URL

### DIFF
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/api/loans/loan.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/api/loans/loan.js
@@ -3,14 +3,16 @@ import { toISO } from '../date';
 import { DateTime } from 'luxon';
 import { serializer } from './serializer';
 
-const loanURL = '/circulation/loans/';
+const loanListURL = '/circulation/loans/';
+const loanURL = loanPid => `${loanListURL}${loanPid}`;
+
 const get = loanPid =>
-  http.get(`${loanURL}${loanPid}`).then(response => {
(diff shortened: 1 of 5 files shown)


invenio_app_ils/ui/src/invenio_app_ils/common/api/loans/loan.js
```diff
@@ -10,7 +10,7 @@ const get = loanPid =>
     return response;
   });
 
-const getLoanReplaceItemUrl = loanPid => `${loanURL}/${loanPid}/replace-item`;
+const getLoanReplaceItemUrl = loanPid => `${loanURL}${loanPid}/replace-item`;
```
ntarocco: Just a comment: I often find myself wondering: "do I have to add the `/` or not?" :) And then... bug :)

What about if we expose all those `url` as functions:
```
getUrl = (loanId='') => {
    return `/circulation/loans/<loanId>`;
}
```

ntarocco (dismissed):

invenio_app_ils/ui/src/invenio_app_ils/common/api/loans/loan.js
topless: `<loanId>` this is not going to work but I get your point :) My personal convention to this super common problem is to always close them on the base urls, and then act accordingly. The double slash doesn't hurt the url but missing one .. will cause issues.

topless (dismissed):

ntarocco (dismissed):

PR #178, opened 2019-02-22
Fix for display number of total hits in tables

### LINKED ISSUE #154: Document active loan counter counts missing items as loaned
https://github.com/inveniosoftware/invenio-app-ils/blob/c4e3d7a0b55b8e9fb698369d6d4d18e4d0f6d955/invenio_app_ils/records/jsonresolver/circulation.py#L35

This line does not take into account that items might be missing and counts them as active loans

### LINKED ISSUE #173: change query for overbooked loans - dashboard
in state: ITEM_ON_LOAN, currently gets all the loans

### LINKED ISSUE #176: Showing total number does not work
The total number of records is always 10 in ResultTable component (due to elastic search config of pagination), it should take data from total field of the response

vlad-bm:
(Closes #173)
(Closes #176)
(Closes #154)

### DIFF
--- a/invenio_app_ils/records/jsonresolver/circulation.py
+++ b/invenio_app_ils/records/jsonresolver/circulation.py
@@ -8,7 +8,6 @@
 """Ils Records Documents record circulation resolver."""
 
 import jsonresolver
-from invenio_circulation.api import is_item_available
 from invenio_circulation.search.api import search_by_pid
 from werkzeug.routing import Rule
 
@@ -24,18 +23,22 @@ def jsonresolver_loader(url_map):
     def _circulation_for_document_resolver(document_pid):
         """Return circulation info for the given document."""
         circulation = {}
-        active_loans_count = 0
         loanable_items_count = 0
         items_count = 0
         item_pids = circulation_items_retriever(document_pid)
 
         for pid in item_pids:
             items_count += 1
-            if not is_item_available(pid):
-                active_loans_count += 1
 
             if circulation_is_item_available(pid):
                 loanable_items_count += 1
+
+        active_loans_count = search_by_pid(
+            document_pid=document_pid,
+            filter_states=current_app.config.get(
+                "CIRCULATION_STATES_LOAN_ACTIVE", []
+            ),
+        ).execute().hits.total
 
         search = search_by_pid(
             document_pid=document_pid,
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js
@@ -45,7 +45,7 @@ const queryBuilder = () => {
 
 const list = query => {
   return http.get(`${documentURL}?q=${query}`).then(response => {
-    let totalHits = response.data.hits.total ? response.data.hits.total : 0;
+    const totalHits = response.data.hits.total;
     response.data = response.data.hits.hits.map(hit =>
       serializer.fromJSON(hit)
     );
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js
@@ -45,9 +45,11 @@ const queryBuilder = () => {
 
 const list = query => {
   return http.get(`${documentURL}?q=${query}`).then(response => {
+    let totalHits = response.data.hits.total ? response.data.hits.total : 0;
     response.data = response.data.hits.hits.map(hit =>
       serializer.fromJSON(hit)
     );
+    response.data.totalHits = totalHits;
     return response;
   });
 };
--- a/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/Home/components/lists/OverdueLoansList/state/actions.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/Home/components/lists/OverdueLoansList/state/actions.js
@@ -12,6 +12,7 @@ export const fetchOverdueLoans = () => {
         loanApi
           .query()
           .overdue()
+          .withState('ITEM_ON_LOAN')
           .qs()
       )
       .then(response => {
--- a/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemsSearch/components/ResultsList/ResultsList.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemsSearch/components/ResultsList/ResultsList.js
@@ -24,12 +24,17 @@ export class ResultsList extends Component {
 
   render() {
     const rows = this.prepareData();
+    const numRowsToShow =
+      rows.length > ResultsTable.defaultProps.showMaxRows
+        ? rows.length
+        : ResultsTable.defaultProps.showMaxRows;
 
     return rows.length ? (
       <ResultsTable
         rows={rows}
         name={'All items'}
         actionClickHandler={this.viewDetailsClickHandler}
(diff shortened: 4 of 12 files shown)


invenio_app_ils/circulation/utils.py
```diff
@@ -70,6 +70,14 @@ def circulation_is_item_available(item_pid):
     return False
 
 
+def circulation_is_item_in_loanable_state(item_pid):
+    """Check if item is in a loanable state."""
+    item = Item.get_record_by_pid(item_pid)
+    if item:
+        return item["status"] not in ["MISSING", "IN_BINDING", "SCANNING"]
```
ntarocco: There is already a method doing the same thing just above, `circulation_is_item_available`.
We should use that one probably, but I realised that it *could* also throw an exception.

Could you please use the already existing method and improve it with something like:
```
from invenio_pidstore.errors import PersistentIdentifierError

...
try:
    return item["status"] == "LOANABLE"
except PersistentIdentifierError as ex:
    return False
```

invenio_app_ils/records/jsonresolver/circulation.py
```diff
@@ -31,7 +31,7 @@ def _circulation_for_document_resolver(document_pid):
 
         for pid in item_pids:
             items_count += 1
-            if not is_item_available(pid):
+            if not is_item_available(pid) and circulation_is_item_in_loanable_state(pid):
```
ntarocco: Just to note that I have change this method in my PR, so better if you merge it before me and I will resolve the conflict.

invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js
```diff
@@ -45,9 +45,11 @@ const queryBuilder = () => {
 
 const list = query => {
   return http.get(`${documentURL}?q=${query}`).then(response => {
+    let totalHits = response.data.hits.total ? response.data.hits.total : 0;
```
ntarocco: `let` -> `const`

I think `ES` always returns the `hits.total` even if 0 results or did you see a case that it does not?
Same for the other occurrences of this line in this PR.

invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js
```diff
@@ -45,9 +45,11 @@ const queryBuilder = () => {
...
     response.data = response.data.hits.hits.map(hit =>
       serializer.fromJSON(hit)
     );
+    response.data.totalHits = totalHits;
```
ntarocco, line 52: Even if I understand it is the easiest solution, I am personally not super fan of enhancing the Array and adding a custom property, it can lead to nasty issues (like [iteration issues](https://stackoverflow.com/questions/9952126/add-property-to-javascript-array)).

I would rather have:
```
response: {
    data: {
        hits: [],
        total: 245
    }
}
```
but then you will have to change more things, because the object structure is changing.

Let's see what the other reviewers think about it.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/Home/components/lists/OverdueLoansList/state/actions.js
```diff
@@ -12,6 +12,7 @@ export const fetchOverdueLoans = () => {
         loanApi
           .query()
           .overdue()
+          .withState('ITEM_ON_LOAN')
```
ntarocco, line 15: We managed to avoid hardcoding `ITEM_ON_LOAN` until now... but I don't have a better solution...

tests/api/test_circulation_utils.py
```diff
@@ -46,3 +47,9 @@ def test_circulation_item_exists(testdata):
 def test_circulation_item_not_exist(testdata):
     """Test return False if item does not exist."""
     assert not circulation_item_exists("not-existing-item-pid")
+
+
+def tes_circulation_is_item_in_loanable_state(testdata):
```
ntarocco: typo `tes` -> `test` otherwise your test won't run. Anyway, you can remove it if you remove the method.

ntarocco (commented):
Nice job! Congrats! 👏 
Since it is your first PR, I added many comments (and I will in the first PRs if you don't mind!)

invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js
kpsherva: I think it was because of tests, because example data didn't have the hits.total

invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js
kpsherva: I thought about it as the easiest solution, in the other case we would have to change structure passed in all the reducers.
There shouldn't be iteration issues if we won't use the `for ... in`.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/Home/components/lists/OverdueLoansList/state/actions.js
kpsherva: we could read all of the states from config, like in the other cases

kpsherva (commented):
🎊 first PR! Well done!
I've added some comments under Nicola's, we can check how to solve them.

invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js
zzacharo: Es is returning the total number for sure, if example data are not well formed I think we should fix them. Assuming we have the total key there the `totalHits` is redundant I think.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/Home/components/lists/OverdueLoansList/state/actions.js
zzacharo: I think is better to address that after when we will have a more clear picture of what is needed to be in the config. Except if is clear that goes there :)

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemsSearch/components/ResultsList/ResultsList.js
```diff
@@ -24,12 +24,17 @@ export class ResultsList extends Component {
...
+      rows.length > ResultsTable.defaultProps.showMaxRows
+        ? rows.length
+        : ResultsTable.defaultProps.showMaxRows;
 
     return rows.length ? (
       <ResultsTable
         rows={rows}
         name={'All items'}
         actionClickHandler={this.viewDetailsClickHandler}
+        showMaxRows={numRowsToShow}
```
zzacharo: just to be consistent is it better to call it `maxRowsToShow`?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoansSearch/components/ResultsList/ResultsList.js
```diff
@@ -25,12 +25,17 @@ export class ResultsList extends Component {
...
+      rows.length > ResultsTable.defaultProps.showMaxRows
+        ? rows.length
+        : ResultsTable.defaultProps.showMaxRows;
 
     return (
       <ResultsTable
         rows={rows}
         name={''}
         actionClickHandler={this.viewDetailsClickHandler}
+        showMaxRows={numRowsToShow}
```
zzacharo: Same here.

PR #221, opened 2019-03-27
indexer: Location indexer and iloc re-index

### LINKED ISSUE #186: Re-index ilocs when editing loc
When editing any field of a location, we should re-index the internal locations that refer to it.
In general, we have to check all the `$ref` and re-index when the data is updated using the record editor or anything else.

topless:
* when updating a location property, we re-index all internal locations referring to it
* changed bulk_index to explicit index each record
* resolves #186

### DIFF
--- a/invenio_app_ils/indexer.py
+++ b/invenio_app_ils/indexer.py
@@ -19,20 +19,21 @@ from invenio_indexer.api import RecordIndexer
 
 from invenio_app_ils.circulation.utils import circulation_document_retriever, \
     circulation_items_retriever
-from invenio_app_ils.records.api import Document, Item
+from invenio_app_ils.records.api import Document, InternalLocation, Item, \
+    Location
+from invenio_app_ils.search.api import InternalLocationSearch
+
+indexer = RecordIndexer()
 
 
 @shared_task(ignore_result=True)
 def index_loans_after_item_indexed(item_pid):
     """Index loan to refresh item reference."""
     loan_search = search_by_pid(item_pid=item_pid)
-    loan_ids = []
     for loan in loan_search.scan():
         record = Loan.get_record_by_pid(loan[Loan.pid_field])
         if record:
-            loan_ids.append(record.id)
-
-    RecordIndexer().bulk_index(loan_ids)
+            indexer.index(record)
 
 
 @shared_task(ignore_result=True)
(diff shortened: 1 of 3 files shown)


invenio_app_ils/indexer.py
```diff
@@ -19,20 +19,19 @@
...
     """Index loan to refresh item reference."""
     loan_search = search_by_pid(item_pid=item_pid)
-    loan_ids = []
     for loan in loan_search.scan():
         record = Loan.get_record_by_pid(loan[Loan.pid_field])
         if record:
-            loan_ids.append(record.id)
-
-    RecordIndexer().bulk_index(loan_ids)
+            RecordIndexer().index(record)
```
zzacharo: Why not indexing asynchronous here?

invenio_app_ils/indexer.py
```diff
@@ -50,28 +49,25 @@ class ItemIndexer(RecordIndexer):
...
     """Index item to refresh document reference."""
     item_pids = circulation_items_retriever(document_pid)
-    item_ids = []
     for pid in item_pids:
         record = Item.get_record_by_pid(pid)
-        item_ids.append(record.id)
-
-    RecordIndexer().bulk_index(item_ids)
+        if record:
+            RecordIndexer().index(record)
```
zzacharo: same here.

invenio_app_ils/indexer.py
topless: According to @ntarocco, we don't want to wait for the 5 minutes for the task to be picked up. But this is just our idea, feel free to elaborate.

invenio_app_ils/indexer.py
topless: Same here :p

invenio_app_ils/indexer.py
zzacharo: where is defined the 5 minutes? I assume you mean the indexing task?

ntarocco: `bulk_index` is adding to the queue, then we need to `process` the queue or wait for 5 minutes.
See: https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/cli.py#L323

@zzacharo WDYT?
In this way, it is already async with `@shared_task`

zzacharo: I mean is already asynchronous the task so this assumes that we don't really waiting for the result either way. But I see the point of not waiting, not sure if is a problem. I think that the ES is more performant in the bulk index form but I am not sure though. We can change it and see if it cause troubles or not in the future.

zzacharo (dismissed):

invenio_app_ils/indexer.py
ntarocco: Agree, it just that the process_queue triggered an error when we tried (we did not investigate).
@topless do you want to try that? It is a little performance gain.

PR #243, opened 2019-04-07
api: add electronic items data model

### LINKED ISSUE #126: backend: Add ebooks field to document schema
Document should have an ebook field point to available ebooks (ref to ebook item or just a computed field.)
( relates to https://github.com/CERNDocumentServer/cds-dojson/issues/193)

ntarocco:
I thought quite a while what's best around ElectronicItem, Digital*, EItem, in the end I went for EItem everywhere to keep it simple.

It contains:
* EItem data model
* reference to the Document and resolver
* refactoring of cli demo data generation
* some refactoring around indexers
( closes #126)


invenio_app_ils/circulation/utils.py
```diff
@@ -32,29 +30,16 @@ def circulation_build_item_ref(loan_pid):
...
 
 
-def circulation_document_retriever(item_pid):
-    """Retrieves document given an item."""
-    item = Item.get_record_by_pid(item_pid)
-    item = item.replace_refs()
-    if item.get("document") and item["document"].get("document_pid"):
-        return item["document"]["document_pid"]
-    return ""
+def circulation_document_pid_retriever(item_pid):
```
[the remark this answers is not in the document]
ntarocco: Agree, when I used it the first time I expected a `document`, and not a `pid`. I am in favor for more explicit and verbose in general, so I will change the other too, but let me know if you prefer the opposite.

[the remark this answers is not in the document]
zzacharo: I prefer explicit naming also the only thing is that the config name in circulation has been named differently but we can live with it I guess..

invenio_app_ils/records/resolver/loan.py
```diff
@@ -14,7 +14,7 @@
 
 
 def item_resolver(loan_pid):
-    """Resolve an Item given a Loan PID."""
+    """Resolve a Item given a Loan PID."""
```
equadon: Sorry for being a grammar police but **an** item is correct.

invenio_app_ils/records/resolver/jsonresolver/document_eitem.py
```diff
@@ -0,0 +1,45 @@
...
+
+@jsonresolver.hookimpl
+def jsonresolver_loader(url_map):
+    """Resolve the referred EItems for a Document record."""
+    from flask import current_app
+
+    def eitems_resolver(document_pid):
+        """Search and return the EItems that reference this Document."""
+        eitems = []
+        for hit in EItemSearch().search_by_document_pid(document_pid).scan():
```
equadon, line 30: I think from a data modelling point of view that this is the best approach where each EItem belongs to a document and has a ref to it. But since we mostly work with documents wouldn't storing the EItem pids in the document instead (similar to keywords) perform better since you don't need to scan the index? We'll resolve EItems for a document more often than the other way around, right?

Maybe it's negligible, just curious if this could impact the performance in a negative way.

equadon (commented):
Nice refactoring in the CLI data generation! What do you think about using a library like [faker](https://github.com/joke2k/faker) to produce more varied and "proper" data? Do you think there's any added value in doing that?

topless (dismissed):

invenio_app_ils/records/resolver/jsonresolver/document_eitem.py
zzacharo: But I think that the documents index will grow much more than the eitem one...

invenio_app_ils/mappings/v6/documents/document-v1.0.0.json
```diff
@@ -39,6 +39,26 @@
...
+            "description": {
+              "type": "text"
+            },
+            "_internal_notes": {
+              "type": "text"
+            },
+            "open_access": {
+              "type": "boolean"
+            },
+            "urls": {
```
zzacharo: Is this gonna be a string of multiple urls?

invenio_app_ils/schemas/eitems/eitem-v1.0.0.json
```diff
@@ -0,0 +1,47 @@
...
+    "description": {
+      "type": "string",
+      "title": "Any extra description for this EItem"
+    },
+    "_internal_notes": {
+      "type": "string",
+      "title": "Any extra description for this EItem reserved for internal usage"
+    },
+    "urls": {
+      "type": "object",
```
zzacharo: Here urls have another type, shouldn't we be consistent? Also is an array no?

invenio_app_ils/schemas/documents/document-v1.0.0.json
```diff
@@ -291,6 +291,14 @@
       "title": "List of languages of the document",
       "type": "array",
       "uniqueItems": true
+    },
+    "_items": {
```
zzacharo: `eitems`? Or you intend to put different item formats for the document e.g cd, cassette etc.?

invenio_app_ils/records/resolver/jsonresolver/document_eitem.py
equadon: Yes, I agree but does looking up specific pids take the same amount of time as scanning?

invenio_app_ils/records/resolver/loan.py
ntarocco: 👍

invenio_app_ils/records/resolver/jsonresolver/document_eitem.py
ntarocco: We discussed about this F2F :) Thanks for the comment!

zzacharo (dismissed):

zzacharo (approved):

kpsherva (approved):

PR #257, opened 2019-04-10
circulation: fix issue with anonymous search

### LINKED ISSUE #215: Validate backoffice_permission().allows(g.identity) works as intended
This comes from the error review from @kprzerwa 

Quoting the comment:
"Check if it works when user not logged in
Because there were problems with that, as g.identity does not exist yet and it might raise an exception"

https://github.com/inveniosoftware/invenio-app-ils/blob/9c15a73714b2998469d6b6f2aba8391bf50c55b6/invenio_app_ils/circulation/search.py#L87

equadon:
Handle the case of a non-authenticated user trying to search.

* Fixes #215

### DIFF
--- a/invenio_app_ils/circulation/search.py
+++ b/invenio_app_ils/circulation/search.py
@@ -11,6 +11,7 @@ import re
 
 from elasticsearch_dsl.query import Q
 from flask import current_app, g, request
+from flask_login import current_user
 from invenio_circulation.search.api import LoansSearch, search_by_pid
 from invenio_search.api import DefaultFilter
 
@@ -86,7 +87,12 @@ def circulation_search_factory(self, search, query_parser=None):
     from invenio_records_rest.sorter import default_sorter_factory
 
     query_string = request.values.get('q', '')
-    query = _default_parser(qstr=query_string)
+
+    if not current_user.is_authenticated:
+        raise UnauthorizedSearchError(query_string)
+
+    parser = query_parser or _default_parser
+    query = parser(qstr=query_string)
 
     # if the logged in user in not librarian or admin, validate the query
     if not backoffice_permission().allows(g.identity):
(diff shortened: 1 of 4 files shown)


invenio_app_ils/circulation/search.py
```diff
@@ -88,6 +89,9 @@ def _default_parser(qstr=None):
     query_string = request.values.get('q', '')
     query = _default_parser(qstr=query_string)
 
+    if not current_user.is_authenticated:
+        raise UnauthorizedSearchError(query_string, patron_pid=None)
```
topless: We can possible skip the `patron_pid=None`

equadon: Fixed now! I also updated `UnauthorizedSearchError` and the error tests.

kpsherva (dismissed):

invenio_app_ils/circulation/search.py
```diff
@@ -88,6 +89,9 @@ def _default_parser(qstr=None):
     query_string = request.values.get('q', '')
     query = _default_parser(qstr=query_string)
 
+    if not current_user.is_authenticated:
```
zzacharo, line 91: Just a minor comment isn't it nicer to start with this check and raise as soon as possible?

zzacharo: @topless can you also change in line 90 to something like
```
parser = query_parser or _default_parser
query = parser(qstr=query_string)
```
as we ignore I think the the `query_parser` argument.

topless: @zzacharo I can but @equadon can too :)

topless (dismissed):

invenio_app_ils/circulation/search.py
equadon: @zzacharo It would be nicer but then we can't raise a `UnauthorizedSearchError` since it expects the query string. Should we raise another exception instead?

Update: Changed so the exception is raised as early as possible and that we use `query_parser` if it's provided.

kpsherva (approved):

PR #278, opened 2019-04-17
delete location

### LINKED ISSUE #268: backend: delete location
- In order to be able to delete a `location` there should be no `internal_locations` referring to it.
- required by #129

### LINKED ISSUE #270: ui: create confirmation dialog for delete
required by #129 
connect #268

topless:
* resolves #268 
* resolves #270

### DIFF
--- a/invenio_app_ils/records/api.py
+++ b/invenio_app_ils/records/api.py
@@ -17,7 +17,9 @@ from invenio_pidstore.resolver import Resolver
 from invenio_records.api import Record
 from invenio_userprofiles.api import UserProfile
 
-from invenio_app_ils.errors import DocumentKeywordNotFoundError
+from invenio_app_ils.errors import DocumentKeywordNotFoundError, \
+    RecordHasReferencesError
+from invenio_app_ils.search.api import InternalLocationSearch
 
 from ..pidstore.pids import (  # isort:skip
     DOCUMENT_PID_TYPE,
@@ -154,6 +156,22 @@ class Location(IlsRecord):
     def create(cls, data, id_=None, **kwargs):
         """Create Location record."""
         return super(Location, cls).create(data, id_=id_, **kwargs)
+
+    def delete(self, **kwargs):
+        """Delete Location record."""
+        iloc_search = InternalLocationSearch()
+        iloc_search_res = iloc_search.search_by_location_pid(
+            location_pid=self[Location.pid_field]
+        )
+        if iloc_search_res.count():
+            raise RecordHasReferencesError(
+                record_type='Location',
+                record_id=self[Location.pid_field],
+                ref_type='Internal Location',
+                ref_ids=sorted([res[InternalLocation.pid_field]
+                                for res in iloc_search_res.scan()])
+            )
+        return super(Location, self).delete(**kwargs)
 
 
 class InternalLocation(IlsRecord):
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/api/locations/internalLocation.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/api/locations/internalLocation.js
@@ -3,34 +3,23 @@ import { internalLocationSerializer as serializer } from './serializer';
 
 const internalLocationURL = '/internal-locations/';
 
-const get = internalLocationPid => {
-  return http
-    .get(`${internalLocationPid}${internalLocationPid}`)
-    .then(response => {
-      response.data = serializer.fromJSON(response.data);
-      return response;
-    });
+const get = async internalLocationPid => {
+  const response = await http.get(
+    `${internalLocationURL}${internalLocationPid}`
+  );
+  response.data = serializer.fromJSON(response.data);
+  return response;
 };
 
-const list = query => {
-  if (query) {
-    return http.get(`${internalLocationURL}?q=${query}`).then(response => {
-      response.data.total = response.data.hits.total;
-      response.data.hits = response.data.hits.hits.map(hit =>
-        serializer.fromJSON(hit)
-      );
-      return response;
-    });
-  } else {
-    return http.get(`${internalLocationURL}`).then(response => {
-      response.data.total = response.data.hits.total;
-      response.data.hits = response.data.hits.hits.map(hit =>
-        serializer.fromJSON(hit)
-      );
-      return response;
-    });
-  }
+const list = async (query = '') => {
+  const response = await http.get(`${internalLocationURL}?q=${query}`);
+  response.data.total = response.data.hits.total;
+  response.data.hits = response.data.hits.hits.map(hit =>
+    serializer.fromJSON(hit)
+  );
+  return response;
 };
+
 export const internalLocation = {
   list: list,
   get: get,
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/config.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/config.js
@@ -5,3 +5,7 @@ const getInvenioConfig = () => {
 };
 
 export const invenioConfig = getInvenioConfig();
+
+// NOTE: the delay that is used when we fire a request with setTimeOut
+// in order to give some time to elastic search to index.
+export const TIMEOUT_DELAY = 3000;
--- a/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/LocationList.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/LocationList.js
@@ -1,37 +1,65 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { Container } from 'semantic-ui-react';
-import { location as locationApi } from '../../../common/api';
+import {
+  location as locationApi,
+  internalLocation as internalLocationApi,
+} from '../../../common/api';
 import { openRecordEditor } from '../../../common/urls';
 import { InternalLocationList } from './components';
 import { Error, Loader, ResultsTable } from '../../../common/components';
+import { DeleteRecordModal } from '../../backoffice/components';
 import { Button } from 'semantic-ui-react';
 import { NewButton } from '../components/buttons';
 import { formatter } from '../../../common/components/ResultsTable/formatters';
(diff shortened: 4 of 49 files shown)


invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/LocationList.js
```diff
@@ -24,16 +30,38 @@ export default class LocationList extends Component {
...
+              internalLocationApi.list(`location_pid:${row.location_pid}`)
+            }
+          />
+        </>
       );
       return omit(serialized, ['Created', 'Updated', 'Link']);
     });
   }
 
+  async _requestDelete(locationId) {
```
kpsherva: is there any reason why this is not in actions/reducer?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/LocationList.js
```diff
@@ -24,16 +30,38 @@ export default class LocationList extends Component {
...
+    await this.deleteLocation(locationId);
+    let { hasError } = this.props;
+    if (!hasError) {
+      remove(this.props.data.hits, entry => entry.location_pid === locationId);
+      // NOTE: setState in order to trigger a render
+      this.setState({ state: this.state });
```
equadon: Another option is to use https://reactjs.org/docs/react-component.html#forceupdate.

Also could there be issues by updating the props directly rather than changing the data in redux so the changes propagate properly to the component?

topless: Actually I left the comment there in order to catch the attention and learn something. 

My thinking behind it was mostly the last sentence of the documentation you mentioned.

`Normally you should try to avoid all uses of forceUpdate() and only read from this.props and this.state in render().`

By setting the assigning the state to the state itself it triggers a render without actually changing anything.

For the redux comment I agree, I should pass it there and propagate the removal from the table from the state.

equadon: According to https://stackoverflow.com/questions/43841930/in-react-what-is-the-difference-between-setstate-and-forceupdate `forceUpdate()` and setting the state to itself yields the same result (except one doesn't call `shouldComponentUpdate`). Then I suppose the React documentation refers to both ways of forcing an update(?).

Either way, it's probably better to handle removal in Redux to prevent any possible inconsistencies.

zzacharo: I think that the filtering of the data should be done inside the `deleteLocation` action. https://github.com/inveniosoftware/invenio-app-ils/blob/e077d69fe49899b8b56f5e2c21b88ebaa5e77bff/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/state/actions.js#L40 here you can get the current state and pass the new data in the action payload.

topless: I actually moved it in the reducer, re-review please and will adjust it accordingly.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/state/reducer.js
```diff
@@ -9,7 +16,7 @@ export const initialState = {
 
 export default (state = initialState, action) => {
   switch (action.type) {
-    case IS_LOADING:
+    case IS_LOADING || DELETE_IS_LOADING:
```
equadon: I'm not sure you can you use `||` in a `case` like this. Is it really gonna use `DELETE_IS_LOADING`? Usually fall-through is used when you need to match multiple cases in a `switch`.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/state/reducer.js
```diff
@@ -26,6 +33,25 @@ export default (state = initialState, action) => {
         error: action.payload,
         hasError: true,
       };
+    case DELETE_SUCCESS:
+      return {
+        ...state,
+        isLoading: false,
+        data: {
+          hits: state.data.hits.filter(
+            hit => hit.location_pid !== action.locationId
```
equadon: :+1:

invenio_app_ils/records/api.py
```diff
@@ -155,6 +157,18 @@ def create(cls, data, id_=None, **kwargs):
...
         return super(Location, cls).create(data, id_=id_, **kwargs)
 
+    def delete(self, **kwargs):
+        """Delete Location record."""
+        iloc_search = InternalLocationSearch()
+        count = iloc_search.search_by_location_pid(
+            location_pid=self['location_pid']
+        ).count()
+        if count:
+            raise RecordHasReferencesError(
```
ntarocco, line 167: Does it make sense to also say for which other records it has references?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/state/reducer.js
ntarocco: Is this because ES is still not up-to-date? Can we add a comment or something? It might look strange at first look.

invenio_app_ils/records/api.py
topless: @ntarocco do you mean to display the references also apart from the record that they refer to?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/state/reducer.js
```diff
@@ -26,6 +33,28 @@ export default (state = initialState, action) => {
...
         hasError: true,
       };
+    case DELETE_IS_LOADING:
+      return { ...state, isLoading: true };
+    case DELETE_SUCCESS:
+      return {
+        ...state,
+        isLoading: false,
+        data: {
+          hits: state.data.hits.filter(
```
topless: @ntarocco as it concerns this one, what happens here is, I manually remove the removed entry from the existing list. After delete I am not submitting a request to refetch them. 

My initial expectation was that when I delete, I would get back in the response the data of the deleted record, but my delete responses come back with empty data.

`{data: "", status: 204, statusText: "NO CONTENT", headers: {…}, config: {…}, …}`

zzacharo: @topless I would leave the reducer logic simple with no logic and do this filtering in your action before and just pass the data in the payload. So this would end up just updating the data with what comes from the action. WDYT?

invenio_app_ils/ui/src/invenio_app_ils/common/api/locations/internalLocation.js
```diff
@@ -3,34 +3,23 @@ import { internalLocationSerializer as serializer } from './serializer';
...
-  } else {
-    return http.get(`${internalLocationURL}`).then(response => {
-      response.data.total = response.data.hits.total;
-      response.data.hits = response.data.hits.hits.map(hit =>
-        serializer.fromJSON(hit)
-      );
-      return response;
-    });
-  }
+const list = async (query = '') => {
```
kpsherva, line 14: we should discuss if we prefer to use `await` - `async` or we stay with using `Promise` as we should keep the code consistent. 
Maybe this will help with the discussion: https://medium.com/limitedio/should-i-use-promises-or-async-await-126ab5c98789

invenio_app_ils/records/api.py
ntarocco: Yes maybe to be added later on...

invenio_app_ils/ui/src/invenio_app_ils/common/api/locations/internalLocation.js
ntarocco: await/async! :)

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/state/reducer.js
ntarocco: @topless this still need to be moved to the action as discussed, no?

kpsherva (dismissed):
Could you also add screenshots of how the deletion works on the UI?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/state/reducer.js
topless: @ntarocco this was removed completely, we are re-fetching with a small delay (the Karolina way) to ensure their deletion.

invenio_app_ils/ui/src/invenio_app_ils/common/config.js
```diff
@@ -5,3 +5,7 @@ const getInvenioConfig = () => {
 };
 
 export const invenioConfig = getInvenioConfig();
+
+// NOTE: the delay that is used when we fire a request with setTimeOut
+// in order to give some time to elastic search to index.
+export const TIMEOUT_DELAY = 3000;
```
ntarocco, line 11: Thanks for this. Can you please replace all occurrences of `3000` with this new config?

ntarocco (dismissed):

invenio_app_ils/ui/src/invenio_app_ils/common/config.js
topless: @ntarocco I was planning to do it in a clean separate pr, in order to add also config options like `showMaxRows`, `showMaxItems` etc.

topless:
delete_location_low_640

equadon (dismissed):

invenio_app_ils/records/api.py
topless: Its added now, the error displays also the ids of the reference that cause the problem.

kpsherva (approved):

ntarocco (approved):

PR #280, opened 2019-04-23
backoffice ui: fix errors when pasting barcode

### LINKED ISSUE #253: backoffice ui: Wrong barcode pasted raises an error
If the pasted barcode results in no matches this will raise an error. Is that something we should guard against?

_Originally posted by @equadon in https://github.com/inveniosoftware/invenio-app-ils/pull/240_

equadon:
* Fix error when pasting barcode with no results
* Show notification on barcode search error
* Bug fix: Notifications are now displayed properly if a JSON response is returned from Invenio (e.g. `InvalidQueryRESTError` in Invenio Records REST)
* Fixes #253

### DIFF
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/api/urls.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/api/urls.js
@@ -2,7 +2,7 @@ export const ApiURLS = {
   loans: {
     loan: '/circulation/loans/:loanPid',
     list: '/circulation/loans/',
-    checkout: '/circulation/loans/checkout',
+    create: '/circulation/loans/create',
     request: '/circulation/loans/request',
   },
   items: {},
--- a/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/ItemsCheckout/state/__tests__/actions.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/ItemsCheckout/state/__tests__/actions.js
@@ -20,7 +20,6 @@ const item = { item_pid: '2' };
 const user_pid = '2';
 const shouldForceCheckout = false;
 const doForceCheckout = true;
-const postURL = ApiURLS.loans.request;
 
 sessionManager.user = { id: '2', locationPid: '2' };
 
@@ -47,7 +46,7 @@ describe('ItemsCheckout actions tests', () => {
 
       store.dispatch(actions.checkoutItem(item, user_pid)).then(() => {
         expect(mockPOST).toHaveBeenCalledWith(
-          postURL,
+          ApiURLS.loans.create,
           item.item_pid,
           loan,
           sessionManager.user.id,
@@ -74,7 +73,7 @@ describe('ItemsCheckout actions tests', () => {
         .dispatch(actions.checkoutItem(item, user_pid, doForceCheckout))
         .then(() => {
           expect(mockPOST).toHaveBeenCalledWith(
-            postURL,
+            ApiURLS.loans.create,
             item.item_pid,
             loan,
             sessionManager.user.id,
@@ -100,7 +99,7 @@ describe('ItemsCheckout actions tests', () => {
 
       store.dispatch(actions.checkoutItem(item, user_pid)).then(() => {
         expect(mockPOST).toHaveBeenCalledWith(
-          postURL,
+          ApiURLS.loans.create,
           item.item_pid,
           loan,
           sessionManager.user.id,
@@ -125,7 +124,7 @@ describe('ItemsCheckout actions tests', () => {
 
       store.dispatch(actions.checkoutItem(item, user_pid)).then(() => {
         expect(mockPOST).toHaveBeenCalledWith(
-          postURL,
+          ApiURLS.loans.create,
           item.item_pid,
           loan,
           sessionManager.user.id,
--- a/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/ItemsSearch/ItemsSearch.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/ItemsSearch/ItemsSearch.js
@@ -24,6 +24,7 @@ export default class ItemsSearch extends Component {
     this.clearResults = this.props.clearResults;
     this.checkoutItem = this.props.checkoutItem;
     this.fetchUpdatedCurrentLoans = this.props.fetchUpdatedCurrentLoans;
+    this.state = { prevSearchQuery: '' };
   }
 
   componentDidMount() {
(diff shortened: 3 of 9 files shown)


zzacharo (dismissed):

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/UserDetails/components/ItemsSearch/ItemsSearch.js
```diff
@@ -32,28 +34,24 @@ export default class ItemsSearch extends Component {
...
+    this.setState({ currentQueryString: '' });
     this.executeSearch(queryString).then(data => {
-      if (this.props.items.hits[0].status === 'LOANABLE') {
-        this.checkoutItem(this.props.items.hits[0], this.props.patron).then(
-          () => {
-            this.clearResults();
-            this.fetchUpdatedCurrentLoans(this.props.patron);
-          }
-        );
+      if (hits && hits[0].status === 'LOANABLE') {
```
vlad-bm: Shouldn't we also check if hits[0] is truthy? Otherwise .status will be called on undefined.

equadon: Yes, you're right. Thanks for spotting that!

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/UserDetails/components/ItemsSearch/ItemsSearch.js
```diff
@@ -37,23 +38,25 @@ export default class ItemsSearch extends Component {
...
     let queryString = e.clipboardData.getData('Text');
     this.executeSearch(queryString).then(data => {
-      if (this.props.items.hits[0].status === 'LOANABLE') {
-        this.checkoutItem(this.props.items.hits[0], this.props.patron).then(
-          () => {
-            this.clearResults();
-            this.fetchUpdatedCurrentLoans(this.props.patron);
-          }
-        );
+      const { patron } = this.props;
```
kpsherva: I think you could assign in outside the `then` or as class attribute if you don't like the long `this.props.patron`, because in this case it gets assigned every time on paste, but the patron does not change.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/UserDetails/components/ItemsSearch/ItemsSearch.js
```diff
@@ -37,23 +38,25 @@ export default class ItemsSearch extends Component {
...
+      const { hits } = this.props.items;
+      if (!_isEmpty(hits) && hits[0].metadata.status === 'LOANABLE') {
```
kpsherva: actually now if think about it, we should make sure that it returns only one hit and if it returns more raise an error, or at least do not check out automatically, I know it is not likely as the barcodes are unique, but maybe it is worth it to check anyway. WDYT?

equadon: Can we do anything to guarantee that barcodes are unique in the database?

If we can't then yes I agree. Perhaps we can do a checkout if there is only one hit and show the search results if there are multiple hits?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/UserDetails/components/ItemsSearch/ItemsSearch.js
```diff
@@ -37,23 +38,25 @@ export default class ItemsSearch extends Component {
   };
 
   executeSearch = queryString => {
+    queryString = queryString || this.props.queryString;
+    this.setState({ prevSearchQuery: queryString });
+
     if (queryString) {
       return this.fetchItems(queryString);
```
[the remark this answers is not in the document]
equadon: I think this function needs to return a promise for the `onPasteHandler` but I will clean it up so it always returns a promise and that `executeSearch` is not called if it's empty.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/UserDetails/components/ItemsSearch/ItemsSearch.js
```diff
@@ -37,29 +38,31 @@ export default class ItemsSearch extends Component {
...
-            this.fetchUpdatedCurrentLoans(this.props.patron);
-          }
-        );
+
+    if (queryString) {
+      await this.executeSearch(queryString);
+
+      const { hits } = this.props.items;
+      const hasOneHit = !_isEmpty(hits) && hits.length === 1;
+      if (hasOneHit && hits[0].metadata.status === 'LOANABLE') {
```
ntarocco: warning, it will break because now it is `CAN_CIRCULATE`

equadon: Thanks!

ntarocco (dismissed):

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/UserDetails/components/ItemsSearch/ItemsSearch.js
kpsherva: yes I would go with that.
I checked, and we are not currently ensuring the unique constraint on the barcode, I've created an issue to solve it.

invenio_app_ils/ui/src/invenio_app_ils/common/api/urls.js
```diff
@@ -3,6 +3,7 @@ export const ApiURLS = {
     loan: '/circulation/loans/:loanPid',
     list: '/circulation/loans/',
     checkout: '/circulation/loans/checkout',
```
kpsherva: could you please remove the checkout link because it is not valid, I thought I removed it maybe it's a rebase leftover?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/ItemsSearch/ItemsSearch.js
```diff
@@ -37,29 +38,31 @@ export default class ItemsSearch extends Component {
...
-      return this.fetchItems(this.props.queryString);
-    }
+    queryString = queryString || this.props.queryString;
+    this.setState({ prevSearchQuery: queryString });
+
+    return this.fetchItems(queryString);
   };
 
-  _onPasteHandler = e => {
+  _onPasteHandler = async e => {
```
kpsherva: we should discuss if we use `Promise` or `await` - `async` or at least make sure to define cases in which we use one or another. 
BTW in this case what will happen if the `executeSearch` is not fulfilled?

equadon: Yes, I agree. Should we keep using promises until we have discussed this?

If `executeSearch` would fail it would not be handled and I think an exception will be raised. For the cases where we need to catch errors it might be cleaner to use the `Promise` directly instead of wrapping code in a `try..catch`. In this case it should be fine because `this.fetchItems` already catches errors.

ntarocco: I vote for `await/async` 👍 :)

ntarocco (dismissed):

kpsherva (dismissed):

kpsherva (dismissed):

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/ItemsCheckout/state/__tests__/actions.js
```diff
@@ -20,7 +20,7 @@ const item = { item_pid: '2' };
 const user_pid = '2';
 const shouldForceCheckout = false;
 const doForceCheckout = true;
-const postURL = ApiURLS.loans.request;
+const postURL = ApiURLS.loans.create;
```
topless: I would use `ApiURLS.loans.create` as is, instead of renaming it to postUrl.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/ItemsSearch/ItemsSearch.js
topless: Use a more useful name than `e` i.e. if it refers to an event, `event` is a good name for it :)

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/ItemsSearch/ItemsSearch.js
```diff
@@ -37,29 +38,31 @@ export default class ItemsSearch extends Component {
...
-            this.fetchUpdatedCurrentLoans(this.props.patron);
-          }
-        );
+
+    if (queryString) {
+      await this.executeSearch(queryString);
+
+      const { hits } = this.props.items;
+      const hasOneHit = !_isEmpty(hits) && hits.length === 1;
+      if (hasOneHit && hits[0].metadata.status === 'CAN_CIRCULATE') {
```
topless: You can do all the checks together for your one hit.

```
const hasOneHit = !_isEmpty(hits) && hits.length === 1 && hits[0].metadata.status === 'CAN_CIRCULATE';
if (hasOneHit) {
...
```

PR #284, opened 2019-04-23
backoffice: available items facet with ranges

### LINKED ISSUE #279: DocumentsSearch: Available items facet
The facet should be a boolean-like, filter by: has items available for loan or not, because in the current state there is no use case to filter by each number. It might be also by ranges 1-10 available items.

equadon:
* Added `keyed_range_filter`
* Fixes #279 

available-items-facet


ntarocco (dismissed):
@equadon Does it make sense to add this range filter to any invenio module?

equadon:
@ntarocco I think it would make sense. This range filter can be used anytime you need a facet with a keyed range, e.g. 0, 1-2, 2-4, 4+. Invenio-Records-REST already has `range_filter` but it only works on date ranges but we could call this `keyed_range_filter` or something. What do you think?

ntarocco (approved):

kpsherva (approved):

PR #294, opened 2019-04-25
patrons: prefix index name

zzacharo:
* Needed for be compliant with invenio-indexer
  and support ES on Demand

### DIFF
--- a/invenio_app_ils/indexer.py
+++ b/invenio_app_ils/indexer.py
@@ -189,10 +189,22 @@ class PatronsIndexer(RecordIndexer):
         data = record.dumps()
         return data
 
+    def _prefix_index(self, app, index):
+        """Prefixes the given index if needed.
+
+        :param app: Flask app to get the config from.
+        :param index: Name of the index to prefix.
+        :returns: A string with the new index name prefixed if needed.
+        """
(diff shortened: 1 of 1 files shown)


invenio_app_ils/indexer.py
```diff
@@ -189,10 +189,20 @@ def _prepare_record(self, record, index, doc_type):
         data = record.dumps()
         return data
 
+    def _prefix_index(self, app, index):
+        """Prefixes the given index if needed.
+        :param app: Flask app to get the config from.
+        :param index: Name of the index to prefix.
+        :returns: A string with the new index name prefixed if needed.
+        """
+        index_prefix = app.config['SEARCH_INDEX_PREFIX'] or ''
```
kpsherva, line 200: maybe we should add a line with that config variable with description to let people know it is possible to configure it ? WDYT?

kpsherva (dismissed):

ntarocco:
Would be nice to discuss this F2F! :)

invenio_app_ils/indexer.py
zzacharo: this is coming from `invenio-search` but I will add a line pointing there for ref :)

ntarocco (dismissed):

kpsherva (dismissed):

PR #299, opened 2019-04-29
item: fix date type in schema

### LINKED ISSUE #252: Backoffice ui: Circulation status not refreshed on the item details page

### LINKED ISSUE #282: Indexing of loan broken because of the end_date format
When the Loan is checked out out with the item (using circulation endpoint exposed in `loan.links.actions`) the start date and end date are changing format (check json in database). In effect the indexing fails because of parsing date problem. 
Please check other endpoints. Check if ui serializer of loan is not causing this (I've checked briefly, seems not)

kpsherva:
* previous date type was not aligned with invenio-circulation
* ( closes #282 )
* ( closes #252 )

invenio-circulation produces different date format, and when the output (loan) was inserted to circulation_status it causes mapping error in ES

### DIFF
--- a/invenio_app_ils/mappings/v6/items/item-v1.0.0.json
+++ b/invenio_app_ils/mappings/v6/items/item-v1.0.0.json
@@ -53,8 +53,7 @@
               "type": "keyword"
             },
             "transaction_date": {
-              "type": "date",
-              "format": "yyyy-MM-dd"
+              "type": "date"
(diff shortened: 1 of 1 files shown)


ntarocco (approved):
I think we should create an issue to review all the mappings, especially the resolved references, before going live...

equadon (approved):

PR #300, opened 2019-04-29
delete item

### LINKED ISSUE #266: backend: delete item
- Loan.links override the default invenio with actions when it should extend it with actios. Self link is required.
- required by #129

topless:
* delete item
* resolves #266


invenio_app_ils/records/api.py
```diff
@@ -144,6 +145,22 @@ def create(cls, data, id_=None, **kwargs):
         }
         return super(Item, cls).create(data, id_=id_, **kwargs)
 
+    def delete(self, **kwargs):
+        """Delete Item record."""
+        loan_search_res = search_by_pid(
+            item_pid=self[Item.pid_field],
+            filter_states=["ITEM_ON_LOAN"]
```
ntarocco: It is probably safer to use `current_app.config['CIRCULATION_STATES_LOAN_ACTIVE']`

ntarocco (commented):
Unfortunately, I think that for Item it is not that easy :(

Loan will always be there. If you check only for active loans, what will happen is that a completed loan will reference an item that has been deleted.
If you re-index that loan, it will try to resolve the item and most probably fail.

While it is probably correct to keep the deleted `item_pid` attached to the loan, I would say that you have to try to fix the json resolver for deleted the item when resolving from loan. It should probably just return an empty dict?
On top of that, we need to be sure that when displaying the loan details in the UI for a deleted item, the UI should not fail but probably show "DELETED ITEM" or similar.

I guess it requires that you test this user case.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoanDetails/components/LoanMetadata/LoanMetadata.js
```diff
@@ -19,6 +27,57 @@ export default class LoanMetadata extends Component {
     this.setState({ isAvailableItemsVisible: newState });
   };
 
+  _renderItem() {
+    const data = this.props.loanDetails;
+    return (
+      !_isEmpty(data.metadata.item) && (
```
zzacharo: I think this happening only when the loan is in pending state right? In that case I would say is better to check the condition that semantically is correct. WDYT?

invenio_app_ils/ui/src/invenio_app_ils/common/api/items/item.js
```diff
@@ -11,6 +11,11 @@ const get = itemPid => {
   });
 };
 
+const del = async itemPid => {
```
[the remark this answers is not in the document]
topless: @zzacharo its a reserved word for JavaScript.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/components/ItemMetadata/index.js
```diff
@@ -2,15 +2,16 @@ import { compose } from 'redux';
...
-import { fetchItemDetails } from '../../state/actions';
+import { deleteItem, fetchItemDetails } from '../../state/actions';
 
 const mapStateToProps = state => ({
   error: state.itemDetails.error,
   itemDetails: state.itemDetails.data,
 });
 
-const mapDispatchToProps = dispatch => ({
+const mapDispatchToProps = (dispatch, ownProps) => ({
```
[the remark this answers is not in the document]
topless: Its a redux convention, if you pass a second argument passes the props https://react-redux.js.org/api/connect#ownprops, pretty neat and useful for the case we want to redirect etc..

topless:
> Unfortunately, I think that for Item it is not that easy :(
> 
> Loan will always be there. If you check only for active loans, what will happen is that a completed loan will reference an item that has been deleted.
> If you re-index that loan, it will try to resolve the item and most probably fail.
> 
> While it is probably correct to keep the deleted `item_pid` attached to the loan, I would say that you have to try to fix the json resolver for deleted the item when resolving from loan. It should probably just return an empty dict?
> On top of that, we need to be sure that when displaying the loan details in the UI for a deleted item, the UI should not fail but probably show "DELETED ITEM" or similar.
> 
> I guess it requires that you test this user case.

This is exactly the intended behaviour for item delete. I will upload a gif soon demonstrating it.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoanDetails/components/LoanMetadata/LoanMetadata.js
topless: For the first question, I am under the impression that only documents can have loans on pending state. For the second one, I didn't understand what you are suggesting :/

zzacharo: You are right about the pending loans! But why do we need to check if `item` ref  is empty?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/state/actions.js
```diff
@@ -25,3 +37,36 @@ export const fetchItemDetails = itemPid => {
...
+          payload: { itemPid: itemPid },
+        });
+        setTimeout(() => {
+          dispatch(
+            sendSuccessNotification(
+              'Success!',
+              `The item ${itemPid} has been deleted.`
+            )
+          );
+          ownProps.history.push(BackOfficeURLS.itemsSearch);
```
zzacharo: Ok I see why ownprops was needed here, but I addressed this issue in the past that's why there is this [history.js](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/ui/src/invenio_app_ils/history.js). This prevents us by just using `withRouter` in order to redirect urls. See how it is used [here](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/components/ItemMetadata/components/CreateNewLoanModal/CreateNewLoanModal.js#L50). I think we need to refactor all the places that we do that as the actual purpose of `withRouter` is to expose the component in the router and be able to route to it.

topless:
item_with_active_loan
item_delete_with_loan

invenio_app_ils/records/jsonresolver/loan.py
```diff
@@ -16,9 +17,13 @@
...
-    del item["circulation_status"]
-    del item["document"]
+    try:
+        item = Item.get_record_by_pid(item_pid)
+        # remove `Document` and `circulation_status` fields
+        # to avoid circular deps.
+        del item["$schema"]
+        del item["circulation_status"]
+        del item["document"]
+    except PIDDeletedError as identifier:
```
equadon: Maybe remove this unused `identifier`.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/state/actions.js
```diff
@@ -25,3 +37,36 @@ export const fetchItemDetails = itemPid => {
...
+
+export const deleteItem = (itemPid, ownProps) => {
+  return async dispatch => {
+    dispatch({
+      type: DELETE_IS_LOADING,
+    });
+
+    await itemApi
+      .del(itemPid)
+      .then(response => {
```
equadon: Did we decide yet if we're gonna use `async` or promises API?

If we use `async` it's probably cleaner if we do: `const response = await itemApi.del(itemPid);`

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/components/ItemMetadata/ItemMetadata.js
```diff
@@ -1,16 +1,30 @@
...
+  constructor(props) {
+    super(props);
+    this.deleteItem = props.deleteItem;
+  }
+
+  _handleOnRefClick(loanPid) {
+    const navUrl = generatePath(BackOfficeURLS.loanDetails, {
+      loanPid: loanPid,
+    });
+    window.open(navUrl, `_target_${loanPid}`);
```
ntarocco: quick question, curiosity: why `_target_<loadnPid>` and not maybe `_blank`?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/components/ItemMetadata/ItemMetadata.js
```diff
@@ -32,6 +46,16 @@ export default class ItemMetadata extends Component {
...
             clickHandler={() => openRecordEditor(itemApi.url, item.item_pid)}
           />
+          <DeleteRecordModal
+            headerContent={`Are you sure you want to delete the Item
+            record with ID ${item.item_pid}?`}
+            onDelete={() => this.deleteItem(item.item_pid)}
+            refType={'Loan'}
+            onRefClick={this._handleOnRefClick}
+            checkRefs={() =>
+              loanApi.list(`item_pid:${item.item_pid} AND state:ITEM_ON_LOAN`)
```
ntarocco: `state:ITEM_ON_LOAN`, I think we should use the config for `loanActiveStates` and maybe `PENDING` too? This is to avoid to delete an item when we have a `PENDING` loan on it.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoanDetails/components/LoanMetadata/LoanMetadata.js
```diff
@@ -19,6 +27,57 @@ export default class LoanMetadata extends Component {
...
+        <Message
+          warning
+          icon="trash alternate"
+          header={`The Item ${
+            data.metadata.item_pid
+          } assigned to this loan has been deleted!`}
+          content={[
+            'If you need further assistance contact ',
+            <a
+              href={`mailto:cds-support@cern.ch?subject=Deleted Item ${
```
ntarocco: We cannot hardcode such email :)
You have to expose yet again another config from config.py, `SUPPORT_EMAIL`

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoanDetails/components/LoanMetadata/LoanMetadata.js
```diff
@@ -19,6 +27,57 @@ export default class LoanMetadata extends Component {
...
+                data.metadata.item_pid
+              }`}
+              key="cds-support"
```
ntarocco: We need to remove any `cds` from ILS :)

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoanDetails/components/LoanMetadata/LoanMetadata.js
```diff
@@ -19,6 +27,57 @@ export default class LoanMetadata extends Component {
...
+        />
+      )
+    );
+  }
+
+  _renderDeletedItem() {
+    const data = this.props.loanDetails;
+    return (
+      _isEmpty(data.metadata.item) &&
+      data.metadata.item_pid && (
```
ntarocco: I like it, thanks! I am just wondering if it is enough the assumption: empty `{}` = deleted item. But for now I think let's keep it like this!

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoanDetails/components/LoanMetadata/LoanMetadata.js
ntarocco: > For the first question, I am under the impression that only documents can have loans on pending state. For the second one, I didn't understand what you are suggesting :/

I wouldn't given for granted: you could have systems that allow both.
@zzacharo the check if in case the Item is deleted.
@topless maybe a comment or a named function could help for above and below code?
pseudo-code:
```
function isItemDeleted() {
...
}

return(!isItemDeleted(...))
```

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/components/ItemMetadata/ItemMetadata.js
topless: If you click more than one link they will open in different browser tabs instead of the same.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/state/actions.js
topless: We should also wrap it in a try catch

invenio_app_ils/ui/src/invenio_app_ils/common/config.js
```diff
@@ -9,3 +9,4 @@ export const invenioConfig = getInvenioConfig();
 // NOTE: the delay that is used when we fire a request with setTimeOut
 // in order to give some time to elastic search to index.
 export const ES_DELAY = 3000;
+export const SUPPORT_EMAIL = 'cds-support@cern.ch';
```
zzacharo: Here I think what @ntarocco was saying is to take this value from `invenio-config`, am I right?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/components/ItemMetadata/ItemMetadata.js
```diff
@@ -1,16 +1,30 @@
...
 import './ItemMetadata.scss';
 
 export default class ItemMetadata extends Component {
+  constructor(props) {
+    super(props);
+    this.deleteItem = props.deleteItem;
+  }
+
+  _handleOnRefClick(loanPid) {
+    const navUrl = generatePath(BackOfficeURLS.loanDetails, {
```
zzacharo: Also here because of @ntarocco refactoring with routes we should take care the refactoring of it as he has changed the scattered `generatePaths` in the code :)

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/components/ItemMetadata/ItemMetadata.js
zzacharo: Is this still valid? @ntarocco @topless

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/state/actions.js
```diff
@@ -1,14 +1,27 @@
...
+import history from '../../../../history';
 
 export const fetchItemDetails = itemPid => {
   return async dispatch => {
     dispatch({
       type: IS_LOADING,
     });
 
-    await item
+    await itemApi
```
zzacharo: We need to refactor that using try/catch and await no?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/state/actions.js
zzacharo: Also here :)

invenio_app_ils/ui/src/invenio_app_ils/common/api/items/item.js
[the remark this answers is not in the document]
topless: Still renamed it to delete when exposed from the Api

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/components/ItemMetadata/ItemMetadata.js
topless: @zzacharo this one will be fixed depending on the order of the merge, we are aware ;)

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/state/actions.js
topless: @zzacharo I was tempted to do these changes also since I changed the underlying api call, but the cool thing about it is that async still returns a promise so nothing breaks. I think it will be more preferable to change them all (including the tests) in a dedicated issue.

invenio_app_ils/ui/src/invenio_app_ils/common/config.js
topless: @ntarocco which config you were referring too? I think @zzacharo is right, it should probably be the `invenio-config`

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/components/ItemMetadata/ItemMetadata.js
topless: @zzacharo I think the safest way to treat this is to include pending loans also in the list of loans that prevent you to delete stuff.

zzacharo: No sorry I meant if is still valid to filter by `loanActiveStates` and not just `state:ITEM_ON_LOAN `

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/state/actions.js
zzacharo: @ntarocco was saying to refactor every time we introduce new code just not to have a huge change in the end(plus creating an issue ofc to refactor the existing codebase) that's why I mentioned it ;)

zzacharo (dismissed):

invenio_app_ils/config.py
```diff
@@ -776,6 +776,9 @@ def _(x):
 ILS_INDEXER_TASK_DELAY = timedelta(seconds=5)
 """Time delay that indexers spawning their asynchronous celery tasks."""
 
+ILS_SUPPORT_EMAIL = 'ils-support@test.org'
```
ntarocco: there is already a config for this `SUPPORT_EMAIL`. Can you please remove?

invenio_app_ils/ui/src/invenio_app_ils/common/api/items/item.js
```diff
@@ -4,11 +4,24 @@ import { prepareSumQuery } from '../utils';
 
 const itemURL = '/items/';
 
-const get = itemPid => {
-  return http.get(`${itemURL}${itemPid}`).then(response => {
-    response.data = serializer.fromJSON(response.data);
-    return response;
-  });
+const get = async itemPid => {
+  const response = await http.get(`${itemURL}${itemPid}`);
```
ntarocco, line 8: no `try/catch`?

invenio_app_ils/ui/src/invenio_app_ils/common/api/items/item.js
```diff
@@ -4,11 +4,24 @@ import { prepareSumQuery } from '../utils';
...
+  return response;
+};
+
+const del = async itemPid => {
+  const response = await http.delete(`${itemURL}${itemPid}`);
+  return response;
+};
+
+const list = async query => {
+  const response = await http.get(`${itemURL}?q=${query}`);
```
ntarocco, line 19: no `try/catch`?

ntarocco (changes requested):

invenio_app_ils/ui/src/invenio_app_ils/common/api/items/item.js
zzacharo: I think we need to implement a common mechanism for all these so not to bombard with `try/catch` everywhere...

zzacharo: @ntarocco @topless I think we should think something similar to this https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html. Promises that are used directly in the components probably need to be wrapped by an `ErrorBoundary` component and use try/catch in the redux actions as there we can dispatch diffenent actions. WDYT? I can create a quick prototype for that if you want.

invenio_app_ils/config.py
topless: Yes

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoanDetails/components/LoanMetadata/LoanMetadata.js
```diff
@@ -1,111 +1,170 @@
...
-                  </Table.Cell>
-                </Table.Row>
-                <Table.Row>
-                  <Table.Cell width={4}>State</Table.Cell>
-                  <Table.Cell width={12}>{data.metadata.state}</Table.Cell>
-                </Table.Row>
-              </Table.Body>
-            </Table>
-          </Grid.Column>
+        <Grid.Column>
```
kpsherva, line 114: just as info: there is a presentation component which renders metadata fields (`MetadataTable` in `backoffice/components` if someone would like to reuse.

kpsherva (dismissed):

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/ItemDetails/state/actions.js
```diff
@@ -1,27 +1,69 @@
...
-    await item
-      .get(itemPid)
-      .then(response => {
-        dispatch({
-          type: SUCCESS,
-          payload: response.data,
-        });
-      })
-      .catch(error => {
+    try {
```
topless, line 24: @ntarocco, replying to your previous comments, I have placed the try catch inside the actions, I think that if we move the try/catch to the `api` we will need something like @zzacharo approach, your thoughts?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LoanDetails/components/LoanMetadata/LoanMetadata.js
topless: @kprzerwa you are right I noticed it recently!

ntarocco (approved):

zzacharo (approved):

PR #303, opened 2019-04-30
Elastic search delay as constant

topless:
Suggestion for further refactor.

ResultsTable accepts `showMaxRows`. Across our code this property is mapped to another property in the wrapper component. A few examples, showMaxEntries, showItems, showPendingLoans, showPastLoans etc .. Despite their different name they all refer to the same thing, the amount of rows that will be displayed.

### DIFF
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/config.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/config.js
@@ -8,4 +8,4 @@ export const invenioConfig = getInvenioConfig();
 
 // NOTE: the delay that is used when we fire a request with setTimeOut
 // in order to give some time to elastic search to index.
-export const TIMEOUT_DELAY = 3000;
+export const ES_DELAY = 3000;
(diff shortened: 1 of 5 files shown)


zzacharo (approved):

invenio_app_ils/ui/src/invenio_app_ils/common/config.js
```diff
@@ -8,4 +8,4 @@ export const invenioConfig = getInvenioConfig();
 
 // NOTE: the delay that is used when we fire a request with setTimeOut
 // in order to give some time to elastic search to index.
-export const TIMEOUT_DELAY = 3000;
+export const ES_DELAY = 3000;
```
ntarocco, line 11: Thanks!
Just an idea: why not exposing the python config var `ILS_INDEXER_TASK_DELAY`? It is doing the same thing for the backend part. Or better to keep them separated?

equadon (approved):

invenio_app_ils/ui/src/invenio_app_ils/common/config.js
equadon: It would make sense to merge them but `ILS_INDEXER_TASK_DELAY` is currently set to 5 seconds while this is 3 seconds. 5 seconds might feel quite long in the UI.

topless: I believe that the value of `5` for the `ILS_INDEXER_TASK_DELAY` is an arbitrary choice, we can try to merge them and drop the default value to `3`. @ntarocco I somehow missed that config variable.

ntarocco: You are right that 5 secs is not acceptable for the UI, we should anyway decrease it to 2/3 secs. I am also not sure if it makes sense to merge, but in any case I would set this in a python config so we can override at deployment time.

PR #308, opened 2019-05-03
delete keyword

### LINKED ISSUE #269: backend: delete keyword
required by #129

topless:
* resolves #269

### DIFF
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -26,7 +26,7 @@ from invenio_records_rest.utils import deny_all
 from .facets import keyed_range_filter
 from .indexer import DocumentIndexer, ItemIndexer, LoanIndexer, LocationIndexer
 from .jwt import ils_jwt_create_token
-from .records.api import Document, InternalLocation, Item, Location
+from .records.api import Document, InternalLocation, Item, Keyword, Location
 from .records.jsonresolver.loan import item_resolver
 
 from .records.permissions import (  # isort:skip
@@ -40,6 +40,7 @@ from .search.api import (  # isort:skip
     DocumentSearch,
     InternalLocationSearch,
     ItemSearch,
+    KeywordSearch,
     LocationSearch,
     PatronsSearch
 )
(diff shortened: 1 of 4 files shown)


invenio_app_ils/config.py
```diff
@@ -429,6 +436,34 @@ def _(x):
...
+        ),
+        list_route="/keywords/",
+        default_media_type="application/json",
+        max_result_window=_RECORDS_REST_MAX_RESULT_WINDOW,
+        error_handlers=dict(),
+        list_permission_factory_imp=backoffice_permission,
+        read_permission_factory_imp=record_read_permission_factory,
+        create_permission_factory_imp=record_create_permission_factory,
+        update_permission_factory_imp=record_update_permission_factory,
+        delete_permission_factory_imp=record_delete_permission_factory,
```
ntarocco: I would disallow deletion here: `deny_all`

invenio_app_ils/config.py
```diff
@@ -429,6 +436,34 @@ def _(x):
...
+            _KEYID_CONVERTER
+        ),
+        list_route="/keywords/",
+        default_media_type="application/json",
+        max_result_window=_RECORDS_REST_MAX_RESULT_WINDOW,
+        error_handlers=dict(),
+        list_permission_factory_imp=backoffice_permission,
+        read_permission_factory_imp=record_read_permission_factory,
+        create_permission_factory_imp=record_create_permission_factory,
+        update_permission_factory_imp=record_update_permission_factory,
```
ntarocco: If you allow editing, you have to ensure that you re-index corresponding documents that use that keyword. `deny_all` here too?
Maybe adding a comment on both explaining why it is not allowed...

tests/api/test_record_delete.py
```diff
@@ -52,3 +52,16 @@ def test_delete_internal_location(client, users, json_headers, testdata):
...
+def test_delete_keyword(client, users, json_headers, testdata):
+    """Test DELETE existing keyword."""
+    login_user_via_session(
+        client,
+        email=User.query.get(users["admin"].id).email
+    )
+
+    keyword_pid = 'keyid-1'
+    url = url_for('invenio_records_rest.keyid_item', pid_value=keyword_pid)
+    res = client.delete(url, headers=json_headers)
```
ntarocco, line 66: If you put `deny_all`, then the test will fail. Just remove it IMHO.
You can test the delete method if you want.

```
k1 = Keywords.create({...})
k.delete()
```
But it would make more sense to test the deletion fail instead of the success case...

topless: Yes the crucial part is to test prevent of deletion for the record with references

ntarocco (approved):

equadon (approved):

PR #321, opened 2019-05-06
add series record

### LINKED ISSUE #255: Volumes and series structure and handling
- [ ] Design the solution of storing the volumes with taken into an account migration rules (check cds-dojson repository)
- [ ] Implement the solution with schema, and handling the volumes in the backend

### LINKED ISSUE #347: Volumes and series - backend
- [x] implement data model, update schemas, data validation
- [x] implement necessary resolvers

equadon:
Closes #347

### DIFF
--- a/invenio_app_ils/mappings/v6/documents/document-v1.0.0.json
+++ b/invenio_app_ils/mappings/v6/documents/document-v1.0.0.json
@@ -39,6 +39,26 @@
             }
           }
         },
+        "series": {
+          "type": "object",
+          "properties": {
+            "series_pid": {
+              "type": "keyword"
+            },
+            "issn": {
+              "type": "keyword"
+            },
+            "mode_of_issuance": {
+              "type": "keyword"
+            },
+            "title": {
+              "type": "keyword"
+            },
+            "volume": {
+              "type": "keyword"
+            }
+          }
+        },
         "_computed": {
           "type": "object",
           "properties": {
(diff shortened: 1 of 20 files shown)


equadon:
Right now `Document` and `Series` have many duplicate fields. Do you think it would be a good idea to store the common fields in a separate schema and include it in the document and series schema?

ntarocco:
> Right now `Document` and `MultiVolume` have many duplicate fields. Do you think it would be a good idea to store the common fields in a separate schema and include it in the document and multi volume schema?
> 
> I'm also wondering if we actually need fields like `languages`, `authors`, keywords, etc when that information could be gathered by going through the child documents inside the `MultiVolume`. What do you think?

We can gather all these questions and ask to librarians. I think that we should find the most complete MultiVolume in current CDS and see what fields are different.
About the schema split, well, it is a bit tricky with Invenio I think... can we brainstorm this together?

invenio_app_ils/mappings/v6/documents/document-v1.0.0.json
```diff
@@ -39,6 +39,26 @@
             }
           }
         },
+        "series": {
```
kpsherva, line 42: the `series_pids` are missing from the mapping

invenio_app_ils/mappings/v6/series/series-v1.0.0.json
```diff
@@ -0,0 +1,56 @@
+{
+  "mappings": {
```
kpsherva, line 2: we should ask Anne which fields might be in series

invenio_app_ils/mappings/v6/documents/document-v1.0.0.json
equadon: Do you think it's necessary to have it when the `series` object has the same information?

kpsherva (approved):

vlad-bm (approved):

PR #339, opened 2019-05-10
add selector component and modify keywords

equadon:
* Add/remove keywords from `Document`
* Component: Search and select hits from ElasticSearch
  * Single and multiple selection
  * Configurable search delay
  * Wildcard search
  * Supports search/select for:
    - `Document`
    - `EItem`
    - `InternalLocation`
    - `Items`
    - `Loans`
    - `Locations`
    - `Keyword`
    - `Patron`
    - `Series`

Multi-selection + update document keywords:
keywords_new_480

Single selection + search by author:
select-document_480

### DIFF
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESSelectorModal.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESSelectorModal.js
@@ -1,0 +1,67 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Button, Modal } from 'semantic-ui-react';
+import { ESSelector } from './';
+import './ESSelector.scss';
+
+export default class ESSelectorModal extends Component {
+  state = { visible: false };
+
+  constructor(props) {
+    super(props);
+    this.selectorRef = null;
+  }
+
+  toggle = () => this.setState({ visible: !this.state.visible });
+
+  save = () => {
+    const { onSave } = this.props;
+    if (onSave) {
+      onSave(this.props.selections);
+    }
+    this.toggle();
+  };
+
+  render() {
+    const { title } = this.props;
+    const trigger = React.cloneElement(this.props.trigger, {
+      onClick: this.toggle,
+    });
+
+    return (
+      <Modal
+        id="es-selector-modal"
+        open={this.state.visible}
+        trigger={trigger}
+        size="tiny"
+        centered={false}
+        onClose={this.toggle}
+      >
+        <Modal.Header>{title}</Modal.Header>
+        <Modal.Content>
+          <ESSelector {...this.props} />
+        </Modal.Content>
+        <Modal.Actions>
+          <Button color="black" onClick={this.toggle}>
+            Close
+          </Button>
+          <Button
+            positive
+            icon="checkmark"
+            labelPosition="right"
+            content="Save"
+            onClick={this.save}
+          />
+        </Modal.Actions>
+      </Modal>
+    );
+  }
+}
+
+ESSelectorModal.propTypes = {
+  alwaysWildcard: PropTypes.bool,
+  multiple: PropTypes.bool,
+  trigger: PropTypes.node.isRequired,
+  title: PropTypes.string,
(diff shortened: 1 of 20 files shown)


equadon:
Possible further use-cases for this component:
* Attach items to document
* Select document from item page
* Select authors for document
* Add documents to multi-volumes/series
* Select series for document

vlad-bm:
Regarding the search, do we want case-insensitive search? Because if we do, I think both elastic search and the front-end need to be modified to handle this. What do you think?

equadon:
> Regarding the search, do we want case-insensitive search? Because if we do, I think both elastic search and the front-end need to be modified to handle this. What do you think?

I agree! I created an issue with some proposed changes for the ES mappings: https://github.com/inveniosoftware/invenio-app-ils/issues/355

vlad-bm (dismissed):

invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESSelectorModal.js
```diff
@@ -0,0 +1,74 @@
...
+
+export default class ESSelectorModal extends Component {
+  state = { visible: false };
+
+  constructor(props) {
+    super(props);
+    this.selectorRef = null;
+  }
+
+  show = () => this.setState({ visible: true });
```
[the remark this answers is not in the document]
kpsherva: 👍

invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESSelectorModal.js
```diff
@@ -0,0 +1,74 @@
...
+
+  hide = () => this.setState({ visible: false });
+
+  toggle = () => this.setState({ visible: !this.state.visible });
+
+  save = () => {
+    const { multiple, onSave } = this.props;
+    if (onSave) {
+      if (multiple) {
```
kpsherva: what does the multiple objects to save change?

invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/serializer.js
```diff
@@ -0,0 +1,95 @@
...
+  description: metadata.name ? `Name: ${metadata.name}` : null,
+  extra: `ID: ${metadata.id}`,
+  metadata: metadata,
+});
+
+export const serializeHit = hit => {
+  const { metadata } = hit;
+
+  let result = {};
+  if (metadata.loan_pid) {
```
kpsherva: I think this way of identifying which type of a record to serialise it is a bit risky, especially that you identify the patron by the email. Maybe we can use the schemas name? or some param passed ?

invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESSelectorModal.js
equadon: The idea was that if you enable multi-selection the results returned to `onSave` will be an array with the selections. If you only allow a single selection it made more sense to return the selection as-is and not in an array of one item. But it might be confusing and perhaps it's better to always return an array even if only a single selection is allowed?

invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/serializer.js
equadon: I agree it's risky and will not be very stable if we make changes in the future. The schema name would work great except for patrons which don't have a schema but I suppose if we add a param to patrons we could easily identify those as well.

If we had access to the index name for each hit it would be great but I'm not sure ES supports returning those.

invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESSelectorModal.js
kpsherva: I would simplify it to have always the same structure, but just block in single mode possibility of adding more, like you proposed

PR #351, opened 2019-05-27
delete: modal accepts empty refProps

### LINKED ISSUE #343: DeleteRecordModal handle the case of empty `refProps`
Currently the modal assumes that you always need to fetch data from refs passed to it. We need to make it optional as some record types, e.g `eitems` can be deleted but they don't have any refs.

topless:
Till now the modal was expecting a list of references for the modal we want to delete. Now it has as a default an empty list, so for models without references we can just skip the refProps field.

closes #343

### DIFF
--- a/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/components/DeleteRecordModal/DeleteRecordModal.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/components/DeleteRecordModal/DeleteRecordModal.js
@@ -24,6 +24,13 @@ export default class DeleteRecordModal extends Component {
     this.toggleModal();
   }
 
+  handleOpen() {
+    const { refProps } = this.props;
+    if (!isEmpty(refProps)) {
+      this.fetchReferences(refProps.map(entry => entry.getRefData()));
+    }
+  }
+
   renderDeleteHeader = () => {
     const { deleteHeader } = this.props;
     return (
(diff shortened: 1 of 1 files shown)


equadon (dismissed):

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/components/DeleteRecordModal/DeleteRecordModal.js
```diff
@@ -89,7 +96,8 @@ export default class DeleteRecordModal extends Component {
 
   renderAll() {
     const { data, refProps } = this.props;
```
kpsherva, line 98: `data` missing from `propTypes` of this component (`isLoading` and `hasError` too)

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/components/DeleteRecordModal/DeleteRecordModal.js
```diff
@@ -89,7 +96,8 @@ export default class DeleteRecordModal extends Component {
...
-    const canDelete = sumBy(data, 'total') === 0;
+    const canDelete = isEmpty(refProps) || sumBy(data, 'total') === 0;
```
kpsherva, line 99: do we have a case that `refProps` are not empty but `data['total']` is?

kpsherva (approved):

PR #378, opened 2019-06-12
records: add support for related records

### LINKED ISSUE #256: Editions and languages structure and handling
- [x] implement a field for storing edition 
- [x] implement creating a record_relation between different editions of one book 
- [x] implement creating a record relation between different languages of a book
- [x] other relation types (see schema)

( duplicate of #359 )

### LINKED ISSUE #359: related_record management
We need to implement a way of linking records (manual and automatic in future harvests). 
There is 3 types of linking:
- [x] related _record field
- [x] different editions
- [x] series
Multipart monograph can have a relation.

equadon:
Changes:
* Related records is now implemented using https://github.com/inveniosoftware/invenio-pidrelations
* A new class `IlsRecordWithRelations` has been added to provide an API to manage related records
* The relations are dumped to `related_records`
* Two new API endpoints added:
    - GET `/api/{documents or series}/<pid>/relations`: Retrieve related records
    - POST `/api/{documents or series}/<pid>/relations`: Add/remove relations to record

In order to avoid duplicate entries (e.g. maintaining all relations both ways) when adding/removing related records it will try to see if the current record has a parent and add/remove it there. If it doesn't it will become the parent and add/remove to itself.

Closes #256
Closes #359


equadon:
Questions:
* Whenever we show the document/series details will we also display related records?
    - If yes, it might be better to always resolve them
    - If no, we could do an async call to a new api endpoint whenever we need it
* Where should we index related records?
    - In `documents`/`series` index?
    - Create a new index to index relations?

Decision after discussion: Dump relations in metadata

topless (dismissed):

invenio_app_ils/records/resolver/jsonresolver/related.py
```diff
@@ -0,0 +1,63 @@
...
+                    if key in record:
+                        del record[key]
+                record["relation"] = obj["relation"]
+                related.append(record)
+            return related
+        return resolver
+
+    url_map.add(
+        Rule(
+            "/api/resolver/documents/<pid>/related",
```
zzacharo: I could be wrong but is indexing not affected by this change? I mean if a document has a language relation to another one and the latter is updated the first shouldn't be reindexed? We have indexers to listen to another record type changes but not for the same record type...I think or I can be totally lost :P

invenio_app_ils/records/api.py
```diff
@@ -53,7 +74,112 @@ def create(cls, data, id_=None, **kwargs):
...
+            parent_node = self.node(relation_type)
+        parent_node.remove_child(record.pid)
+
+    def _parent_node(self, relation_type):
+        """Get the parent node with the given relation_type."""
+        node = self.node(relation_type)
+        parents = list(node.parents)
+        is_child = len(parents) > 0
+        if is_child:
+            return PIDNode(pid=parents[0], relation_type=relation_type)
```
zzacharo, line 239: Is the first element in the parents array the nearest parent? Also I see that there is also the option to have ordered pid relations, would this be helpful for multipart monographs?

invenio_app_ils/records/api.py
```diff
@@ -53,7 +74,112 @@ def create(cls, data, id_=None, **kwargs):
...
+        self.remove_related(
+            record,
+            relation,
+            parent_node=self._parent_node(relation)
+        )
+
+    def add_related(self, record, relation_type, parent_node=None):
+        """Add related record with a relation."""
+        record_node = record.node(relation_type)
+        if record_node.is_parent or record_node.is_child:
```
zzacharo, line 170: As I understand here if a record has already relations cannot be added as a relation to another one right?

zzacharo: Note these to two conditions are querying also db every time...

invenio_app_ils/config.py
```diff
@@ -1022,3 +1027,24 @@ def _(x):
...
+# ==============
+LANGUAGE_RELATION = RelationType(0, "language", "Language", "", "")
+EDITION_RELATION = RelationType(1, "edition", "Edition", "", "")
+
+PIDRELATIONS_RELATION_TYPES = [
+    LANGUAGE_RELATION,
+    EDITION_RELATION,
+]
+
+RECORD_TYPE_CLASS = {
```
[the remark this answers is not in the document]
equadon: Yes you're right, it's more of a constant. And now that I think of it this information is already available in `RECORDS_REST_ENDPOINTS` and `CIRCULATION_REST_ENDPOINTS` so probably should get it from there.

invenio_app_ils/records/api.py
```diff
@@ -37,11 +40,29 @@
...
+        )
+
     @classmethod
-    def get_record_by_pid(cls, pid, with_deleted=False):
+    def get_record_by_pid(cls, pid, with_deleted=False, pid_type=None):
         """Get ils record by pid value."""
+        if pid_type is None:
+            pid_type = cls._pid_type
+        else:
+            new_cls = current_app.config["RECORD_TYPE_CLASS"][pid_type]
```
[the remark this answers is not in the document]
equadon: Yes that would work as well. Honestly the only reason was it felt more convenient to pass the `pid_type` to `get_record_by_pid` and have it handled there.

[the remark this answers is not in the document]
zzacharo: But where do you use it by passing the pid_type? Also I saw that in the `_get_related` method you call it from the instance object...is this correct?

invenio_app_ils/records/api.py
```diff
@@ -53,7 +74,112 @@ def create(cls, data, id_=None, **kwargs):
...
+        if is_child:
+            return PIDNode(pid=parents[0], relation_type=relation_type)
+        return self.node(relation_type)
+
+    def _get_related(self, relation_type):
+        """Get all related records with a specific relation type."""
+        node = self._parent_node(relation_type)
+        children = [node.pid] + list(node.children)
+        return [
+            self.get_record_by_pid(child)
```
zzacharo: I mean here :)

invenio_app_ils/records/api.py
equadon: There should never be a case where there are more than 1 parent so it should probably raise an exception if that happens.

Good idea! Yes, the ordering could indeed be helpful for series.

invenio_app_ils/records/api.py
equadon: Correct! Not sure otherwise how we would handle adding a record as an edition that already has other editions. Would we add those editions to the list? Feels like it could be messy. What do you think?

Yes, unfortunately it does. I hadn't planned for this from the start so the number of db calls are more now. Fortunately this is quite easy to optimize by storing some metadata in the record.

invenio_app_ils/records/api.py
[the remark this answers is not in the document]
equadon: Ahh, yes that doesn't make sense. It should pass the `pid_type` otherwise it will only look for records with the same type.

invenio_app_ils/records/views.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+
+    endpoints = app.config.get("RECORDS_REST_ENDPOINTS", [])
+    pid_types = ("docid", "serid")
+
+    for pid_type in pid_types:
+        options = endpoints.get(pid_type, {})
+        rec_serializers = {
+            "application/json": (
+                "invenio_records_rest.serializers" ":json_v1_response"
```
ntarocco:
```suggestion
                "invenio_records_rest.serializers:json_v1_response"
```

invenio_app_ils/records/views.py
```diff
@@ -0,0 +1,102 @@
...
+def create_relations_blueprint(app):
+    """Add relations views to the blueprint."""
+    blueprint = Blueprint(
+        "invenio_app_ils_relations",
+        __name__,
+        url_prefix="",
+    )
+
+    endpoints = app.config.get("RECORDS_REST_ENDPOINTS", [])
+    pid_types = ("docid", "serid")
```
ntarocco: Use here `PID*` instead of hardcoding `docid` and `serid`?

invenio_app_ils/records/views.py
```diff
@@ -0,0 +1,102 @@
...
+                pid_type=obj["pid_type"]
+            )
+            obj["title"] = related.get("title", "")
+            obj["edition"] = related.get("edition", "")
+            obj["language"] = related.get("language", "")
+        pid = record[record.pid_field]
+        for key in list(record.keys()):
+            if key != "related_records":
+                del record[key]
+        return self.make_response(pid, record, 202)
```
ntarocco:
```suggestion
        return self.make_response(pid, record, 200)
```
any particular need for 202? Same below

invenio_app_ils/records/views.py
```diff
@@ -0,0 +1,102 @@
...
+        for obj in record["related_records"]:
+            related = IlsRecordWithRelations.get_record_by_pid(
+                obj["pid"],
+                pid_type=obj["pid_type"]
+            )
+            obj["title"] = related.get("title", "")
+            obj["edition"] = related.get("edition", "")
+            obj["language"] = related.get("language", "")
+        pid = record[record.pid_field]
+        for key in list(record.keys()):
```
ntarocco, line 79: Sorry, but it is not clear to me this for/if.
Isn't enough to do a:
```
record.get('related_records', {})
```
?

And also what about the `obj` dict?

invenio_app_ils/records/views.py
```diff
@@ -0,0 +1,102 @@
...
+        """Update relations."""
+        for obj in request.get_json():
+            related = IlsRecordWithRelations.get_record_by_pid(
+                obj["pid"],
+                pid_type=obj["pid_type"]
+            )
+            relation_type = IlsRecordWithRelations.get_relation_by_id(
+                obj["relation_type"]
+            )
+            action = obj["action"]
```
ntarocco:
```suggestion
            action = obj.get("action", "")
```

invenio_app_ils/records/api.py
```diff
@@ -65,7 +91,176 @@ def update(self, data):
...
+        """Remove related record with a relation."""
+        self._remove_related_record(
+            record,
+            relation_type,
+            parent_node=self._parent_node(relation_type)
+        )
+
+    def _add_related_record(self, record, relation_type, parent_node=None):
+        """Add related record with a relation."""
+        record_node = record.node(relation_type)
```
ntarocco: if the passed `record` is not of type `IlsRecordWithRelation`, here it will fail right? Does it make sense maybe raise before?

invenio_app_ils/records/api.py
```diff
@@ -38,11 +41,34 @@
 class IlsRecord(Record):
     """Ils record class."""
 
+    @property
+    def pid(self):
+        """Get the PersistentIdentifier for this record."""
+        return PersistentIdentifier.get(
+            pid_type=self._pid_type,
```
ntarocco, line 50: Now that I see this, I realize that _pid_type was not a great idea, since we can have multiple pids referencing a record.
Maybe we should have a "main" pid... anyway, nothing to change, just to keep in mind.

invenio_app_ils/records/api.py
```diff
@@ -65,7 +91,176 @@ def update(self, data):
...
+        parent_node.remove_child(record.pid)
+
+    def dump_related_add(self, record, relation, update_related=False):
+        """Add record to related records dump."""
+        obj = self.record_to_dump(record, relation)
+        related_records = self["related_records"]
+        if obj not in related_records:
+            related_records.append(obj)
+
+            if update_related:
```
ntarocco, line 195: not 100% sure about this extra If.
Shouldn't it happen all the time? Why `false` by default?

invenio_app_ils/records/api.py
```diff
@@ -53,7 +81,163 @@ def create(cls, data, id_=None, **kwargs):
...
+    def editions(self):
+        """Get related editions."""
+        return self._get_related(self.edition_relation())
+
+    @property
+    def languages(self):
+        """Get related languages."""
+        return self._get_related(self.language_relation())
+
+    def node(self, relation_type):
```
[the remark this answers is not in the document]
equadon: This is only a helper method to get the `PIDNode` of a record. I should make that more clear in the documentation.

invenio_app_ils/records/api.py
```diff
@@ -53,7 +81,163 @@ def create(cls, data, id_=None, **kwargs):
...
+        parent_node.remove_child(record.pid)
+
+    @classmethod
+    def record_to_dump(cls, record, relation_type):
+        """Convert a record to a dictionary dump."""
+        return dict(
+            pid=record[record.pid_field],
+            pid_type=record._pid_type,
+            relation_type=relation_type.id,
+            title=record.get("title", ""),
```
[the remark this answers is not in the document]
equadon: Yes, this was removed as we discussed. That is an outdated version. Only `pid`, `pid_type`, and `relation_type` are dumped now.

invenio_app_ils/records/views.py
equadon: Since I use the record serializer I needed to return a proper record, that's why I had to keep the record but remove everything except `related_records`. I cleaner option would be to create a new serializer but is it worth the effort?

The `obj` dict in the for loop before is just each entry in `related_records` which is extended with `title`, `edition`, and `language`.

invenio_app_ils/records/api.py
equadon: Yes, I think we should raise when doing validation. Initially I was gonna add the validation in this PR but it seemed easier to include it in the PR for the UI so unless you prefer to have it now I'll add it in the backoffice PR.

invenio_app_ils/records/api.py
equadon: Something that I would like to see changed related to pid type is renaming all `document_pid`, `keyword_pid`, ... to just `pid` and include `pid_type` in the metadata. This would make it a lot easier to identify the record type when dealing with multiple record types.

invenio_app_ils/records/api.py
equadon: When adding/removing relations it'll dump the current record then go to all its related records and dump those too with the new relation. If we don't set it to stop update it's related records it'll end up in an infinite loop.

There was no particular reason for it to be `false` by default, probably makes more sense to keep is as a positional argument.

invenio_app_ils/records/api.py
```diff
@@ -65,7 +91,176 @@ def update(self, data):
...
+            related_records.remove(obj)
+
+            if update_related:
+                for related in self._get_related(relation):
+                    self.updated_related_records.append(related)
+                    dump_record = record if related != record else self
+                    related.dump_related_remove(dump_record, relation)
+
+    @staticmethod
+    def compare_dump_obj(a, b):
```
kpsherva: I couldn't find where do you use it in this PR (is it used ?)

invenio_app_ils/records/api.py
```diff
@@ -65,7 +91,176 @@ def update(self, data):
...
+            relation_type=relation_type.id
+        )
+
+    def commit(self, commit_related=True, **kwargs):
+        """Store changes of the current record instance in the database."""
+        with db.session.begin_nested():
+            if commit_related:
+                for related in self.updated_related_records:
+                    related.commit(commit_related=False)
+            self.updated_related_records = []
```
kpsherva, line 230: this line should be outside of `db.session`

invenio_app_ils/records/api.py
ntarocco: Good, can we discuss this tomorrow? We did that at the beginning, but then we changed back. I need to remember why, maybe @kprzerwa @zzacharo @topless have better memory! :)

ntarocco:
Fantastic work! I would suggest to quickly chat about your comments in person so that we can speed up the merge!

invenio_app_ils/records/api.py
zzacharo: @equadon why don't you use the schema property to identify the record type? This is what is happening now in videos...I don't remember if there was an important reason not to have a common name...

invenio_app_ils/records/api.py
```diff
@@ -38,11 +41,34 @@
...
+            pid_value=self[self.pid_field],
+        )
+
+    @staticmethod
+    def pid_type_to_record_class(pid_type):
+        """Get the record class from the pid_type."""
+        if pid_type == "loanid":
+            endpoints_cfg = current_app.config["CIRCULATION_REST_ENDPOINTS"]
+        else:
+            endpoints_cfg = current_app.config["RECORDS_REST_ENDPOINTS"]
```
zzacharo: I think you could just rely on the current_app.config["RECORDS_REST_ENDPOINTS"] as [here](https://github.com/inveniosoftware/invenio-circulation/blob/master/invenio_circulation/ext.py#L46) we are merging the 2 configs :)

invenio_app_ils/records/api.py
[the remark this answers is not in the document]
zzacharo: Rename probably to something like `get_node`?

invenio_app_ils/records/api.py
zzacharo: Is this something that can be optimized eventually?

equadon: Yes, we could store this information in the dump so we wouldn't have to hit the db.

invenio_app_ils/records/api.py
equadon: @zzacharo That's what I'm doing now in the UI but it's not as clean in my opinion since the schema is versioned. It would be easier to just match a string to figure out the record type.

ntarocco:
Merging after our discussions

PR #381, opened 2019-06-13
backend: patron_exists implemented

### LINKED ISSUE #323: Implement `patron_exists` method
Currently the method returns `True`. We need:

- [x] Check if patron id provided is one of the users in our system

topless:
* implemented patron exists
* removed unused imports
* fixed a bug with undefined params
* closes #323

### DIFF
--- a/invenio_app_ils/api.py
+++ b/invenio_app_ils/api.py
@@ -9,6 +9,7 @@
 
 from __future__ import absolute_import, print_function
 
+from invenio_accounts.models import User
 from invenio_pidstore.errors import PersistentIdentifierError
 
 from .records.api import Item, Location
@@ -53,4 +54,4 @@ def item_exists(item_pid):
 
 def patron_exists(patron_pid):
     """Return True if the Patron exists given a PID."""
-    return True
+    return User.query.filter_by(id=patron_pid).first() is not None
--- a/tests/api/test_apis.py
+++ b/tests/api/test_apis.py
@@ -9,10 +9,13 @@
 
 from invenio_app_ils.api import (  # isort:skip
     get_document_pid_by_item_pid,
-    item_exists,
-    get_location_pid_by_item_pid,
     get_item_pids_by_document_pid,
+    get_location_pid_by_item_pid,
+    item_exists,
+    patron_exists,
 )
+from invenio_accounts.models import User
+
 from invenio_app_ils.records.api import Document, Item
 
 
@@ -46,3 +49,10 @@ def test_item_exists(testdata):
 def test_circulation_item_not_exist(testdata):
     """Test return False if item does not exist."""
     assert not item_exists("not-existing-item-pid")
+
+
+def test_patron_exists(users):
+    """Test return True if item exists."""
+    test_patron = User.query.all()[0]
+    assert patron_exists(test_patron.id)
+    assert not patron_exists('not-existing-patron-pid')
--- a/invenio_app_ils/circulation/api.py
+++ b/invenio_app_ils/circulation/api.py
@@ -19,8 +19,6 @@ from invenio_app_ils.errors import MissingRequiredParameterError, \
 from invenio_app_ils.proxies import current_app_ils_extension
 from invenio_app_ils.records.api import Item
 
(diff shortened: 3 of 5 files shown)


tests/api/test_apis.py
```diff
@@ -46,3 +47,9 @@ def test_item_exists(testdata):
 def test_circulation_item_not_exist(testdata):
     """Test return False if item does not exist."""
     assert not item_exists("not-existing-item-pid")
+
+
+def test_patron_exists(testdata):
+    """Test return True if item exists."""
+    first_patron_pid = testdata["loans"][0][Patron.pid_field]
```
ntarocco: Patron is not a real record, so I think `Patron.pid_field` attribute does not exist (unless you add it).

I would also add a test for the negative, when it does not exist.

invenio_app_ils/api.py
```diff
@@ -53,4 +54,10 @@ def item_exists(item_pid):
...
 def patron_exists(patron_pid):
     """Return True if the Patron exists given a PID."""
-    return True
+    try:
+        patron = User.query.filter_by(id=patron_pid).first()
+        if patron:
+            return True
+    except PersistentIdentifierError as ex:
+        return False
+    return False
```
zzacharo: Isn't this redundant? If it doesn't find something it will raise....

invenio_app_ils/api.py
```diff
@@ -53,4 +54,10 @@ def item_exists(item_pid):
 
 def patron_exists(patron_pid):
     """Return True if the Patron exists given a PID."""
-    return True
+    try:
+        patron = User.query.filter_by(id=patron_pid).first()
+        if patron:
+            return True
+    except PersistentIdentifierError as ex:
```
ntarocco: This exception will be never thrown. You might have an SQLAlchemy exception, but with `first()` it will return None if not found.

Maybe something similar to:
```
def patron_exists(patron_pid):
    return User.query.filter_by(id=patron_pid).first() != None
```

ntarocco (changes requested):

invenio_app_ils/api.py
zzacharo: This resolves my comment 👍

tests/api/test_apis.py
zzacharo: Yes you are right :)

topless: All done thanks

invenio_app_ils/api.py
topless: 👍

topless:
Tests fail also on travis as they do locally.
`tests/api/conftest.py:144:`

`DetachedInstanceError: Instance <Transaction at 0x7fcfeeb34b10> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/bhk3)`

I took the initiative and removed unused imports from `test_mails.py`, it seems if I revert them the tests are passing, @ntarocco @zzacharo @equadon @vladbmihaescu anyone any thoughts or ideas? I don't think its all the dependencies but there is some magic going on there, I think the import of `current_app` is affecting it, will investigate a bit to see whats going on.

So it is the `from invenio_mail.tasks import send_email` that needs to be imported in test_mails so test_apis do not fail.

zzacharo (approved):

invenio_app_ils/api.py
```diff
@@ -53,4 +54,4 @@ def item_exists(item_pid):
 
 def patron_exists(patron_pid):
     """Return True if the Patron exists given a PID."""
-    return True
+    return User.query.filter_by(id=patron_pid).first() is not None
```
equadon, line 57: Should we refer to the patron ID as a PID even if it's not a record?

equadon (approved):

invenio_app_ils/api.py
ntarocco: Agree with you, the param name comes from `invenio-circulation` that does not know if patron is a record or not. Given that, I would probably not change the "signature" of the method. WDYT?

ntarocco (approved):

Issue #136, opened 2019-01-25
loaders: fine tune loaders schema


zzacharo:
Finalize all loaders schema to be more strict. Currently we extend https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/schemas/json.py#L77 which allow all unknown fields to pass through the request in the record data. 

Also we need to override this https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/schemas/json.py#L83 to enable customizing this https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/schemas/json.py#L88 per record type as now is set only for Item.

PR #383, opened 2019-06-14
search: filter loans based on permissions

### LINKED ISSUE #25: Filter the loans by permission (in circulation_state)
Circulation api endpoint for item state returns loan object, if loaned. This should be filtered by permissions, as not every user has permissions to see all the field of a loan (for example, the `patron_pid` and dates if the loan is on another user).

**Tasks**

* Implement a custom serializer in `invenio-circulation` (or in `invenio-app-ils`?) for `Loan` to show/hide/filter loan fields depending on the current logged in user, or None/fail if anonymous
* Implement a custom serializer in `invenio-app-ils` for `Item` to use the `Loan` serializer and show/hide/filter item/loan fields depending on the current logged in user, or None/fail if anonymous

equadon:
* Filter `circulation_status` for resolved items
* Filter `circulation_status` for item searches

Closes #25

### DIFF
--- a/invenio_app_ils/cli.py
+++ b/invenio_app_ils/cli.py
@@ -287,16 +287,18 @@ class DocumentGenerator(Generator):
 
         def random_series():
             data = []
-            for pid in random.sample(multipart_pids, randint(0, len(multipart_pids)-1)):
-                data.append(dict(
-                    pid=pid,
-                    volume=str(randint(1, 100))
-                ))
-            for pid in random.sample(serial_pids, randint(0, len(serial_pids)-1)):
-                data.append(dict(
-                    pid=pid,
-                    volume=str(randint(1, 100))
-                ))
+            if multipart_pids:
+                for pid in random.sample(multipart_pids, randint(0, len(multipart_pids) - 1)):
+                    data.append(dict(
+                        pid=pid,
+                        volume=str(randint(1, 100))
+                    ))
+            if serial_pids:
+                for pid in random.sample(serial_pids, randint(0, len(serial_pids) - 1)):
+                    data.append(dict(
+                        pid=pid,
+                        volume=str(randint(1, 100))
+                    ))
             return data
 
         objs = [{
@@ -374,7 +376,7 @@ class LoanGenerator(Generator):
         for i in range(1, size):
             item = self._get_item_can_circulate(items)
             status = self._get_valid_status(item, items_on_loans)
-            patron_id = randint(1, len(patrons_pids)-1)
+            patron_id = random.choice(patrons_pids)
             transaction_date = datetime(
                 current_year, randint(1, 12), randint(1, 28)
             )
@@ -495,7 +497,7 @@ def data(n_docs, n_items, n_eitems, n_loans, n_keywords, n_intlocs, n_series):
     indexer = RecordIndexer()
 
     holder = Holder(
-        patrons_pids=["1", "2"],
+        patrons_pids=["1", "2", "5", "6"],
         librarian_pid="4",
         total_intloc=n_intlocs,
         total_keywords=n_keywords,
--- a/invenio_app_ils/permissions.py
+++ b/invenio_app_ils/permissions.py
@@ -59,3 +59,8 @@ def views_permissions_factory(action):
         return backoffice_permission()
     else:
         return deny_all()
+
+
+def circulation_status_permission(patron_pid):
+    """Return circulation status permission for a patron."""
+    return Permission(UserNeed(int(patron_pid)), backoffice_access_action)
--- a/invenio_app_ils/records/serializers/__init__.py
+++ b/invenio_app_ils/records/serializers/__init__.py
@@ -9,10 +9,22 @@
 
 from invenio_records_rest.schemas import RecordSchemaJSONV1
 from invenio_records_rest.serializers.json import JSONSerializer
-from invenio_records_rest.serializers.response import record_responsify
+from invenio_records_rest.serializers.response import record_responsify, \
+    search_responsify
+
+from invenio_app_ils.records.serializers.item import ItemJSONSerializer
 
 json_v1 = JSONSerializer(RecordSchemaJSONV1, replace_refs=True)
 """JSON v1 serializer."""
 
+item_v1 = ItemJSONSerializer(RecordSchemaJSONV1, replace_refs=True)
+"""Item v1 serializer."""
+
 json_v1_response = record_responsify(json_v1, 'application/json')
 """JSON response builder that uses the JSON v1 serializer."""
+
+item_v1_response = record_responsify(item_v1, 'application/json')
+"""JSON response builder that filters item circulation status."""
+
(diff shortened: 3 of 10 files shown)


invenio_app_ils/permissions.py
```diff
@@ -59,3 +61,26 @@ def views_permissions_factory(action):
         return backoffice_permission()
     else:
         return deny_all()
+
+
+def filter_circulation_status(circulation_status):
```
equadon: Is this an appropriate place to put this filtering function or would it make more sense to put it somewhere else?

invenio_app_ils/circulation/utils.py
```diff
@@ -51,3 +53,29 @@ def circulation_is_loan_duration_valid(loan):
...
+
+    filter_keys = [
+        "loan_pid",
+        "patron_pid",
+        "start_date",
+        "end_date",
+        "request_expire_date",
+    ]
+
+    if has_request_context():
```
topless:
```suggestion
    allowed = True
    if has_request_context():
        allowed = loan_owner_permission(patron_pid).can()
```

invenio_app_ils/cli.py
```diff
@@ -287,16 +287,18 @@ def generate(self):
...
-                    pid=pid,
-                    volume=str(randint(1, 100))
-                ))
-            for pid in random.sample(serial_pids, randint(0, len(serial_pids)-1)):
-                data.append(dict(
-                    pid=pid,
-                    volume=str(randint(1, 100))
-                ))
+            if multipart_pids:
+                for pid in random.sample(multipart_pids, randint(0, len(multipart_pids)-1)):
```
topless:
```suggestion
                for pid in random.sample(multipart_pids, randint(0, len(multipart_pids) - 1)):
```

invenio_app_ils/cli.py
```diff
@@ -287,16 +287,18 @@ def generate(self):
...
+                    data.append(dict(
+                        pid=pid,
+                        volume=str(randint(1, 100))
+                    ))
+            if serial_pids:
+                for pid in random.sample(serial_pids, randint(0, len(serial_pids)-1)):
```
topless:
```suggestion
                for pid in random.sample(serial_pids, randint(0, len(serial_pids) - 1)):
```

topless (dismissed):

invenio_app_ils/permissions.py
```diff
@@ -38,10 +38,12 @@ def backoffice_permission(*args, **kwargs):
 class LoanOwnerPermission(Permission):
     """Return Permission to evaluate if the current user owns the loan."""
 
-    def __init__(self, record):
+    def __init__(self, record, patron_pid=None):
         """Constructor."""
+        if record:
```
kpsherva: did you have a case that record is not passed here other than the newly created factory? The class was specifically meant for Loan's protection while the circulation_status is a resolved reference. I think this should fail when the record is not provided, especially because we use it in the config `read_permission_factory_imp=LoanOwnerPermission`, instead I would just write simple permission factory just for the `circulation_status`.

invenio_app_ils/circulation/utils.py
```diff
@@ -51,3 +53,29 @@ def circulation_is_loan_duration_valid(loan):
...
+        "request_expire_date",
+    ]
+
+    allowed = True
+    if has_request_context():
+        allowed = loan_owner_permission(patron_pid).can()
+
+    for key in filter_keys:
+        if key in circulation_status and not allowed:
+            del circulation_status[key]
```
kpsherva: I would check if the key exists in the `circulation_status`, or maybe use `.pop()`. For example the `extension_count` is not there in some cases.

kpsherva (changes requested):

invenio_app_ils/permissions.py
equadon: No, it's only the new one that uses it. I wanted to re-use this since I needed a permission check  for loan ownership and this one already does it.

`LoanOwnerPermission` doesn't use `record` besides for getting the patron pid so what do you think about changing the constructor to take the patron pid instead of the record?

invenio_app_ils/records/serializers/item.py
```diff
@@ -0,0 +1,67 @@
...
+    def filter_circulation_status(self, data):
+        """Filter circulation status depending on user permissions."""
+        if "circulation_status" in data["metadata"]:
+            circulation_status = data["metadata"]["circulation_status"]
+            patron_pid = circulation_status.get("patron_pid", None)
+            if not patron_pid:
+                return circulation_status
+
+            allowed = True
+            if has_request_context():
```
ntarocco: I would say that you don't need `has_request_context`, you should always be in a request context, no?

invenio_app_ils/records/serializers/item.py
```diff
@@ -0,0 +1,67 @@
...
+                allowed = loan_owner_permission(patron_pid).can()
+
+            for key in self.FILTER_KEYS:
+                if key in circulation_status and not allowed:
```
ntarocco: would it be better to check `if allowed` before the loop?

ntarocco (dismissed):
Looks good to me! Good job!

invenio_app_ils/records/serializers/item.py
zzacharo: This was a fix for the cli commands as I remember :)

equadon: @ntarocco Yes, it looks like you're right. I removed it and no issues and I'm not sure in what situation that could actually happen.

ntarocco (dismissed):

invenio_app_ils/records/serializers/__init__.py
```diff
@@ -9,10 +9,22 @@
...
-from invenio_records_rest.serializers.response import record_responsify
+from invenio_records_rest.serializers.response import record_responsify, \
+    search_responsify
+
+from invenio_app_ils.records.serializers.item import ItemJSONSerializer
 
 json_v1 = JSONSerializer(RecordSchemaJSONV1, replace_refs=True)
 """JSON v1 serializer."""
 
+item_v1 = ItemJSONSerializer(RecordSchemaJSONV1, replace_refs=True)
```
ntarocco, line 20: I think that we should not be loose using with [RecordSchemaJSONV1](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/schemas/json.py#L39) probably strict and use [StrictKeysMixin](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/schemas/json.py#L20).
This will ensure that we control exactly what fields we output.

WDYT?
@zzacharo WDYT?

zzacharo: Yes that was my intention but back then everything was optional and undecided. You mean this no? https://github.com/inveniosoftware/invenio-app-ils/issues/136

kpsherva (dismissed):

kpsherva (approved):

PR #391, opened 2019-06-19
api: fix wrong Loan indexer

ntarocco:


ntarocco:
@topless @equadon @zzacharo tests will probably fail but this is the fix to the tests in case you need to use it locally until the PR is ready.

invenio_app_ils/ext.py
```diff
@@ -54,7 +55,7 @@ def item_indexer(self):
     def loan_indexer(self):
         """Return a loan indexer instance."""
         endpoints = self.app.config.get('RECORDS_REST_ENDPOINTS', [])
-        pid_type = DOCUMENT_PID_TYPE
+        pid_type = CIRCULATION_LOAN_PID_TYPE
```
zzacharo, line 58: @ntarocco this was the issue?

ntarocco: I guess so, in any case it was wrong. Plus, I noticed another thing: we have created a LoanIndexer that will index Item after a loan changes, but we also registered the signal. So, we were re-indexing multiple times.
Basically, I have the feeling that with this bug, everything was working because of the signal, and not because of the LoanIndexer implementation :)

zzacharo: I think the signal is there to connect to ils with loan change in circulation...The LoanIndexer is used from us in ils to properly index loans and their side effects...So I think we need both no?

topless (approved):

equadon (commented):
This indeed seems to be the issue. It's possible to run individual test files again. Nice find!

equadon (approved):

ntarocco:
I am honestly a bit surprised that tests are passing.

PR #392, opened 2019-06-21
cli: add ils setup command

equadon:
Runs the setup commands using the `CliRunner` instead of through bash to avoid loading Invenio after every CLI call.

```
Usage: ils setup [OPTIONS]

  ILS setup command.

Options:
  --skip-db-destroy  Skip destroying DB.
  --skip-demo-data   Skip creating demo data.
  --skip-patrons     Skip creating patrons.
  --verbose          Verbose output.
```

### DIFF
--- a/invenio_app_ils/cli.py
+++ b/invenio_app_ils/cli.py
@@ -619,3 +619,69 @@ def index():
     for pat in patrons:
         patron = Patron(pat.id)
         indexer.index(patron)
+
+
+@click.command()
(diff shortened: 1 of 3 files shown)


ntarocco (dismissed):

kpsherva (dismissed):

topless (approved):
Great idea!

kpsherva (approved):

PR #404, opened 2019-06-27
api: add mechanism to validate patches

zzacharo:


invenio_app_ils/records/api.py
```diff
@@ -53,6 +53,16 @@ def create(cls, data, id_=None, **kwargs):
         data["$schema"] = current_jsonschemas.path_to_url(cls._schema)
         return super(IlsRecord, cls).create(data, id_=id_, **kwargs)
 
+    def patch(self, patch):
+        """Validate patch metadata."""
+        getattr(self, 'validate_patch', lambda x: pass)(patch)
```
equadon: Another option could be to validate with decorators so that we might be able to reuse them in other places.

We should also considering having a `ValidationError` exception we can raise so the UI has a unified way of dealing with server-side validation responses. What do you think?

ntarocco (commented):
LGTM! I would add also for the `create` method?

ntarocco:
@zzacharo 
```
getattr(self, 'validate_patch', lambda x: pass)(patch)
SyntaxError: invalid syntax
```

invenio_app_ils/records/api.py
zzacharo: For the reusable part when you create these methods in your api class you can use helpers to validate, so these can be used also in other places. I think that way is cleaner than messing with complex validators. For the error handling for sure we need to create special error classes for these errors I agree!

kpsherva (approved):

PR #405, opened 2019-06-28
backoffice: replace document on item

### LINKED ISSUE #310: As a librarian I should be able to replace document for an item
Attach document by PID, title

On the item details page only.
reuse the keyword attachment component
backend: validate if document exists

zzacharo:
closes #310


equadon (dismissed):

ntarocco (dismissed):
Failing because of sorting

PR #410, opened 2019-07-04
Backoffice: Related records

### LINKED ISSUE #373: Related records: backoffice UI
Implement backoffice UI for selecting related records. We'll need a way to select a related record and its relation to the record it's added to.

- [x] implement type of relation to select with the drop down (edition/translation)
- [x] implement radio button to choose the endpoint of search (document / series)
- [x] searchbox (reuse existing component)

! to be implemented both in the document details page and series detail page
ensure proper error handling

TO DO what is parent relation type and other?

### LINKED ISSUE #406: goTo doesn't work properly when navigating to a different record
Problem: `goTo(path)` doesn't reload the details page when going from e.g. `/backoffice/documents/63` to `/backoffice/documents/64`. It works fine if we navigate to a different record type, e.g. `/backoffice/series/80`.

equadon:
Adds support for related records for documents and series.

* Refactored related records
* Improved `ESSelectorModal` to render custom selectors
* `ResultTable`'s `rowActionClickHandler` now passes the entire row not just `ID`
* Linking to a record details page with a different PID now works (closes #406)

Closes #373 

List related records (document & series details page):
details-tabs

Select related records:
select-tabs


equadon:
The error message when trying to remove the parent is now:

```ILS: RelatedRecordError - Failed to remove language relation for document 1 because it is the parent. Please move related documents and series to another parent.```

File: invenio_app_ils/records/related/__init__.py [hunk 1]
```diff
@@ -0,0 +1,195 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
equadon: I agree, it belongs in `api`. The only concern I have is `api.py` is becoming quite big so might be worth splitting if we add more to it in the future.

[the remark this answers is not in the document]
zzacharo: @equadon I meant in an `api.py` file inside the related module :) But yes I agree that we need to monitor if it becomes really big :)

invenio_app_ils/ui/src/invenio_app_ils/common/api/utils.js
```diff
@@ -29,3 +29,14 @@ export const prepareSumQuery = param => {
     return param;
   }
 };
+
+export const recordToPid = record => {
+  const schema = record.metadata['$schema'];
+  if (schema.includes('/documents/document-')) {
```
ntarocco, line 35: after the refactoring of `pid_type`, please remember to refactor this

invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESRelatedSelector.js
```diff
@@ -0,0 +1,194 @@
...
+import './ESRelatedSelector.scss';
+import {
+  LanguageRelation,
+  getRelationTypes,
+  getRelationTypeByName,
+  getIconByRelation,
+  getRelationTypeById,
+} from '../RelatedRecords/config';
+
+export default class ESRelatedSelector extends Component {
```
ntarocco, line 16: Is it important to prefix with `ES`? Me as developer, do I need to know that it is showing results from ES and not DB?
The whole point to use redux is to have the UI **not** to know about the backend.

invenio_app_ils/records/related/__init__.py
```diff
@@ -0,0 +1,199 @@
...
+        parent = self.parent
+        is_parent_equal = parent_pid.pid_value == child[child.pid_field] and \
+            parent_pid.pid_type == child._pid_type
+        if is_parent_equal:
+            # Trying to remove parent
+            # TODO: remove old relations and copy over relations to self.parent
+            records_cfg = current_app.config["RECORDS_REST_ENDPOINTS"]
+            record_name = records_cfg[child._pid_type]["record_class"].__name__
+            raise RelatedRecordError(
+                ("Failed to remove {} relation for {} {} because it is the "
```
ntarocco: I think this message should be maybe more a log.
The user does not really now about parent/child and IMHO is going to be confused.
What about something like (please, fix my english):
`Cannot remove the language relation. Please, remove this relation by editing Document 1`

invenio_app_ils/records/related/__init__.py
```diff
@@ -0,0 +1,199 @@
...
+        )
+
+    def _add_record(self, record, relation, parent_node=None):
+        """Add related record."""
+        if record == self.parent:
+            raise RelatedRecordError("Cannot add itself as a related record.")
+        record_node = get_node(record, relation)
+        if record_node.is_parent or record_node.is_child:
+            raise RelatedRecordError(
+                "Failed to add related record because it already has relations"
```
ntarocco: Can you please add more details to **ALL** error message or logs?
I would like to see the type and the PID of the parent and child, and the relation details. If we have an exception, we need to have the info to be able to fix it! :)

invenio_app_ils/records/related/__init__.py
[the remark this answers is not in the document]
equadon: Ah okay, I misunderstood :)

invenio_app_ils/ui/src/invenio_app_ils/common/api/utils.js
```diff
@@ -29,3 +29,14 @@ export const prepareSumQuery = param => {
     return param;
```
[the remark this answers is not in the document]
equadon: Yes, it became a lot better. Thanks for the suggestion! :)

invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESRelatedSelector.js
equadon: No, I agree. The reason for the naming was that the selector only supports ElasticSearch results but this should probably be refactored to have an ElasticSearch "backend" and then pass the search backend you want to use as a prop to the selector.

ntarocco (approved):

zzacharo (approved):

PR #411, opened 2019-07-04
frontsite: patron profile page

### LINKED ISSUE #375: user own profile page
- [x] user account page implementation
- [x] patron current loan requests list
- [x] patron past loans list
- [x] paginate results of the lists above (we just show all no pagination)
- [ ] test `invenio-accounts-rest` endpoint `/users/` is blocked except for GET method, a separate issue was created for this one #408 
- [x] ensure the users can't see other user's pages

topless:
* implemented profile page component and routing
* fixed sticky footer bug that was scrolling more that the content height
* moved PatronCurrentLoans state to common so it can be shared among backoffice and frontsite
* fixed wrong propType for patronPid
* integrated PatronCurrentLoans, PatronPastLoans, PatronPendingLoans in frontsite
* removed time from start/end date
* removed Authorization header since its not used
* cleaned up and updated the INSTALL.rst
* fixed existing tests and snapshots
* tests for PatronPastLoans state
* withSize added to the loan api query builder
* constant ES_MAX_SIZE defined in common/config
* ability to override the default query for fetchPatronPendingLoans, fetchPatronPastLoans, fetchPatronCurrentLoans
* show all button which does exactly that, shows all, no pagination
* for active loans used the config array of valid active loan states
* closes #375

profile-page


invenio_app_ils/ui/src/invenio_app_ils/common/state/PatronCurrentLoans/__tests__/actions.js
```diff
@@ -64,7 +68,7 @@ describe('Patron current loans tests', () => {
 
       store.dispatch(actions.fetchPatronCurrentLoans(2)).then(() => {
         expect(mockFetchPatronCurrentLoans).toHaveBeenCalledWith(
-          '(patron_pid:2 AND state:ITEM_ON_LOAN)&sort=-mostrecent'
+          '(patron_pid:2 AND state:(ITEM_ON_LOAN))&sort=-mostrecent'
```
zzacharo, line 71: Why the parentheses here? I think was also correct before no?

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/Profile/ProfileContainer.js
```diff
@@ -0,0 +1,28 @@
...
+  PatronPendingLoans,
+} from './components';
+import { sessionManager } from '../../../authentication/services';
+
+export default class ProfileContainer extends Component {
+  render() {
+    const currentUser = sessionManager.user;
+    return (
+      <Container>
+        <PatronMetadata username={currentUser.username} />
```
zzacharo: I don't know if it's better to pass the prop down or just take it directly to each component. These components most probably will not be reused so I don't know if it makes sense to take the user prop.

topless: For the specific case, we can take it in the component directly, you are right, but in every other case the container has to pass down the data to the presentation components. Whatever you prefer for this one.

invenio_app_ils/ui/src/invenio_app_ils/common/state/PatronCurrentLoans/__tests__/actions.js
topless: For current loans we had hardcoded the state in the query, I switched it to the `invenioConfig.circulation.loanActiveStates` which is a list of states, so the form of our query changes, thats why the tests where updated with parentheses.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/ItemsSearch/ItemsSearch.js
```diff
@@ -175,7 +175,7 @@ ItemsSearch.propTypes = {
   fetchUpdatedCurrentLoans: PropTypes.func.isRequired,
   clearResults: PropTypes.func.isRequired,
   checkoutItem: PropTypes.func.isRequired,
-  patron: PropTypes.number.isRequired,
+  patron: PropTypes.string.isRequired,
```
equadon, line 178: I think the patron id is an integer in the database so maybe it makes sense to keep it like that here too?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/PatronDetails/components/PatronCurrentLoans/PatronCurrentLoans.js
```diff
@@ -77,7 +77,7 @@ export default class PatronCurrentLoans extends Component {
 }
 
 PatronCurrentLoans.propTypes = {
-  patron: PropTypes.number,
+  patron: PropTypes.string,
```
equadon, line 80: Same as above!

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/Profile/ProfileContainer.js
kpsherva: the `currentUser` is very specific and as we can read it any time anywhere for logged user I think we should take it directly

topless: 👍

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/Profile/ProfileContainer.js
```diff
@@ -0,0 +1,28 @@
...
+        <PatronPendingLoans patronPid={currentUser.id} />
+        <PatronCurrentLoans patronPid={currentUser.id} />
```
equadon: The prop for `PatronCurrentLoans` is called `patron` but should probably be renamed like the others.

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/Profile/ProfileContainer.js
topless: Well, an update on this one, this would work only for `PatronMetadata` for the rest of the components `PatronPendingLoans`, `PatronCurrentLoans`, `PatronPastLoans` since they are used elsewhere we need to provide the patronPid to maintain reusability.

invenio_app_ils/ui/src/invenio_app_ils/routes/urls.js
```diff
@@ -5,6 +5,7 @@ const FrontSiteBase = '/';
 
 const FrontSiteRoutesList = {
   home: FrontSiteBase,
+  profile: `${FrontSiteBase}profile`,
```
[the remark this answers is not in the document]
topless: For the key or for the value of the object? I thought a bit about it, and I couldn't find who else is going to have a profile apart from patrons. On the other argument, it also matches the link we provide the page and url. I am more in favour of `profile`.

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/Profile/ProfileContainer.js
zzacharo: @topless But they are not as you created new ones no? The backoffice ones are different components. If it's confusing I would say to rename this to `CurrentPatronPendingLoans` or something that indicates that this is for the current user.

invenio_app_ils/ui/src/invenio_app_ils/routes/urls.js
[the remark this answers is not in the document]
kpsherva: for the key.  I'm ok in general with `profile` just when you read the code `patronProfile` is more to the point.

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/Profile/ProfileContainer.js
equadon: I kept it as-is but made sure `patronPid` is a required prop for the components. I think it's fine to pass the prop even if it's always the current user. Generally makes it more testable to pass the prop.

invenio_app_ils/ui/src/invenio_app_ils/routes/urls.js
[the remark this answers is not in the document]
equadon: I changed the key to `patronProfile` but kept the url as `/profile`. I can't find anything else that requires a profile page but explicitly naming it this way leaves no confusion.

zzacharo (approved):

kpsherva (approved):

PR #436, opened 2019-07-16
global: update series schemas and display

### LINKED ISSUE #438: Update schemas - series

kpsherva:
( closes #438 )

### DIFF
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/formatters.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/formatters.js
@@ -119,10 +119,15 @@ function formatSeriesToTableView(series) {
     Updated: toShortDate(fromISO(series.updated)),
   };
   if (!isEmpty(series.metadata)) {
-    serialized['Title'] = series.metadata.title;
-    serialized['Authors'] = series.metadata.authors;
+    serialized['Title'] = series.metadata.title.title;
     serialized['Mode of Issuance'] = series.metadata.mode_of_issuance;
+    if (series.metadata.authors && !isEmpty(series.metadata.authors)) {
+      serialized['Authors'] = series.metadata.authors;
+    } else {
+      serialized['Authors'] = [];
+    }
   }
+
   return serialized;
 }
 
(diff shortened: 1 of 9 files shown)


ntarocco (dismissed):
We will need to keep an eye now to any schema change, so that we don't brake migration given that the schemas are duplicated

invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/formatters.js
```diff
@@ -119,10 +119,15 @@ function formatSeriesToTableView(series) {
     Updated: toShortDate(fromISO(series.updated)),
   };
   if (!isEmpty(series.metadata)) {
-    serialized['Title'] = series.metadata.title;
-    serialized['Authors'] = series.metadata.authors;
+    serialized['Title'] = series.metadata.title.title;
```
vlad-bm, line 122: A bit nitpicky, but we could also check for title existence using has from lodash:
`if (has(series.metadata, "title.title"))`

vlad-bm (dismissed):

invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/formatters.js
kpsherva: In general title is required in the schema so it should not go through validation if there is no title in title, and I would leave it for now because of the migration

kpsherva:
> We will need to keep an eye now to any schema change, so that we don't brake migration given that the schemas are duplicated

@ntarocco we could have one schema stored on the legacy like it was done on videos or a test checking schema with master of cds-dojson. In general we could have a final source of truth for schemas at this point

ntarocco (approved):

topless:
This merge broke master branch and specifically navigation to DocumentDetails in FrontSite, I believe it should be reopened or create a new issue about it.

On a side note, `title.title` doesn't make a lot of sense, if we need it to be and object, it could be something like `title.value`?

We should introduce tests that will validate all our navigation event to avoid issues like this slipping in our master.

WDYT @ntarocco, @kprzerwa, @equadon, @vladbmihaescu, @zzacharo?

ntarocco:
Discussed during stand-up. You can create an issue or just fix it in the next one.

PR #441, opened 2019-07-19
frontsite: pagination for ResultsTable

vlad-bm:
Added previous/next page buttons on the ResultsTable footer and remove the Show All button.
(Closes #440)

PR #445, opened 2019-08-08
frontsite: pagination for ResultsTable

vlad-bm:
(Closes #441)
Screenshot from 2019-08-08 17-13-53


invenio_app_ils/config.py
```diff
@@ -777,6 +777,9 @@ def _(x):
 
 # RECORDS REST facets
 # =========================
+#: Number of records to fetch by default
+RECORDS_REST_DEFAULT_RESULTS_SIZE = 10
```
ntarocco, line 781: We don't need this because it is already defined in `invenio-records-rest`, unless we change the value

ntarocco (commented):
We need to change the sentence `Showing first 10 of 25` to something like `Page 1 of 10` if possible or if it is too complicated just remove it

equadon:
Pagination message in the footer is now: `Showing entries 1-5 of 17`

invenio_app_ils/config.py
zzacharo: We need it if we dont want to retrieve it from `current_app.wsgi_app.mounts['/api'].config`!

equadon: Yes, we discussed it and decided to keep it so we don't have to do that. But it would be nice to have all config in one place. :)

zzacharo: You could expose the whole api config in the current_app['api_config'] probably :) Or most probably better to just create a function that makes the retrieval of this config a bit more readable :P

ntarocco (approved):

PR #446, opened 2019-08-08
frontsite: export csv files

vlad-bm:
(Closes #437)


invenio_app_ils/ui/src/invenio_app_ils/common/api/eitems/eitem.js
```diff
@@ -46,10 +46,29 @@ const queryBuilder = () => {
   return new QueryBuilder();
 };
 
+const count = query => {
```
topless: This block of code gets repeated for every query builder and is identical

invenio_app_ils/ui/src/invenio_app_ils/common/api/eitems/eitem.js
```diff
@@ -46,10 +46,29 @@ const queryBuilder = () => {
...
+  return http.get(`${eitemURL}?q=${query}`).then(response => {
+    response.data = response.data.hits.total;
```
topless: Instead of overriding the data of the response, its is more preferable to handle it by looking for total wherever you use it, or just return the number, thats what I would expect from a count method.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/components/ExportSearchResults/ExportSearchResults.js
```diff
@@ -0,0 +1,206 @@
...
+        key: 'json',
+        text: 'JSON',
+        value: 'json',
+      },
+    ];
+
+    this.resultsOptions = Array.from(Array(totalPages).keys()).map(i => {
+      const page = i + 1;
+      // build dropdown options, e.g.:
+      // 1 - 9999, 10000 - 19999, ...
```
equadon, line 36: If the display limit is `10000` shouldn't it say `1-10000` instead of `1-9999`? Or write `0-9999` then `10000-19999`?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/LocationList.js
```diff
@@ -68,13 +63,7 @@ export default class LocationList extends Component {
 
   renderResults(data) {
     const rows = this.prepareData(data);
-    const headerActionComponent = (
-      <NewButton
-        clickHandler={() => {
-          openRecordEditor(locationApi.url);
```
zzacharo, line 74: @ntarocco I would propose to just make the `openRecordEditor` function dummy so after we implement the forms to be able to find easily all the buttons needed to point to the corresponding form...WDYT?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/components/InternalLocationList/InternalLocationList.js
```diff
@@ -24,7 +20,7 @@ export default class InternalLocationList extends Component {
   }
 
   handleOnRefClick(itemPid) {
-    openRecordEditor(itemApi.url, itemPid);
+    // edit record
```
zzacharo: Or leave a comment like this one in the places you removed the `openRecordEditor()` :)

invenio_app_ils/ui/src/invenio_app_ils/common/api/eitems/eitem.js
ntarocco: As for get or similar methods, I think we will have to live with that, unless you have suggestions! :)

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/components/ExportSearchResults/ExportSearchResults.js
ntarocco: Nice catch! The thing is that I believe there is an issue in `invenio-records-rest`: if I set a size of `10000`, it fails throwing an error that the limit has been reached.
I think there is a < 10000 instead of <= 10000.
So for now I have to keep it like this unless we manage to fix that issue.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/LocationList.js
ntarocco: I will change as suggested by adding a comment!

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/LocationList/components/InternalLocationList/InternalLocationList.js
kpsherva: or TODO, some editors are easily detecting it

kpsherva (dismissed):

PR #452, opened 2019-08-14
Disable limiting from ping

zzacharo:


invenio_app_ils/ext.py
```diff
@@ -83,6 +83,18 @@ def init_app(self, app):
         # disable warnings being logged to Sentry
         logging.getLogger("py.warnings").propagate = False
 
+        # add /ping endpoint in the flask limiter exempt so we
```
diegodelemos, line 100: Wouldn't it be better to use the config [RATELIMIT_PER_ENDPOINT](https://github.com/inveniosoftware/invenio-app/blob/faa48f7eb194a3c2f81b6ef80ff00fcaeffb1953/invenio_app/config.py#L69) to set a very high rate-limiting for the ping endpoint rather than rewriting it here?

zzacharo: The problem is with `uwsgi_curl` that doesn't pass the correct headers so the rate limiting is failing [here](https://github.com/inveniosoftware/invenio-app/blob/faa48f7eb194a3c2f81b6ef80ff00fcaeffb1953/invenio_app/limiter.py#L19) because `remote_addr` is None. But after discussion with @slint he has an issue opened https://github.com/inveniosoftware/invenio-app/issues/49 to address this, so I just did this as a quick solution now because it doesn't make sense by default to have rate limiting on the `/ping` endpoint

lnielsen:
~~Can you fix this centrally in Invneio as it will be a problem for everyone :-)~~

Sorry didn't see the open issue already :-)

equadon (approved):

ntarocco:
@zzacharo thanks for this, but why not fixing in `invenio-app` then?

zzacharo:
@ntarocco needed something quick in order to fix our deployments. But there is an issue open about! But I can create the PR for that in `invenio-app` if you want?

ntarocco:
@zzacharo yes please, that would be nice, and then remove the custom code here :) thanks!

PR #454, opened 2019-08-16
global: document metadata refactor

### LINKED ISSUE #324: Restructure document and item properties
It has been agreed that all the fields internal for the system (not the part of record description) should be encapsulated in the `_computed` field (f.e `circulation` field in the `Document`) 
Comment: I think the field should be name `_system`

**UPDATE**
After a meeting with the team the issue has been discussed and decided to be addressed in a different fashion.

- [x] move `eitems` one level up from `_computed` and remove `_computed`
- [x] ~create `items` for `document` in the same fashion as `eitems`~
- [x] move properties of `circulation` that are relevant to `items` to the newly created `items`
- [x] align `circulation` and `circulation_status` for `document` and `item` respectively

topless:
- removed  one level nesting of `_computed`, and moved `eitems` on the root
- eitems with `total` number of eitems and `hits` with the actual eitems
- `has_eitems` removed from `circulation` and filters and aggregations point to `eitems.total`
- `has_items` removed from `circulation` filters and aggregations point to `items.total`
- `items` `circulation_status` property renamed to `circulation` in alignment with `document`
- fixes book series page on frontsite
- closes #324


invenio_app_ils/records/resolver/jsonresolver/document_item.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Resolve the EItem referenced in the Document."""
```
equadon: Docs and comments in this file need to be updated with `Item` instead of `EItem`.

invenio_app_ils/schemas/documents/document-v1.0.0.json
```diff
@@ -341,10 +341,24 @@
       "type": "array",
       "uniqueItems": true
     },
-    "_computed": {
+    "eitems": {
       "type": "object",
       "properties": {
-        "eitems": {
+        "total": { "type": "integer" },
+        "hits": {
```
equadon: Do we consider only `hits` as a `ref` or shouldn't `total` also be a part of the resolved ref?

Same applies to `items`.

invenio_app_ils/records/resolver/jsonresolver/document_item.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@jsonresolver.hookimpl
+def jsonresolver_loader(url_map):
+    """Resolve the referred EItems for a Document record."""
+    from flask import current_app
+
+    def items_resolver(document_pid):
+        """Search and return the total number of items."""
+        # NOTE: in comparison to eitems we don't return the actual items in
```
ntarocco: Actually, I was wrong, there should be many. It is Documents attached to Serial that can be too many.
It should be safe to resolve items here.
To add it with a few fields.

invenio_app_ils/schemas/documents/document-v1.0.0.json
ntarocco: yes, the entire `items` and `eitems` should be a $ref.

ntarocco:
@equadon I have rebased, but could you please check if everything is right? I could not load the react app correctly, I didn't have the time to check if it is on my machine or there is any other issue. Thanks!

PR #468, opened 2019-08-27
stats: average number of loans per document item

### LINKED ISSUE #432: Document Stats: add average
In the document stats, add a new column after `past loans` and `renewals` to display the average amount of loans per physical copy.
Simply, calculate `past loans / n. of items`, without taking into account the number of copies in the given range of time.

topless:
* implemented past loans to items ratio for document statistics
* tests for the average presentation
* renamed `number_of_past_loans` to `past_loans_count` to match rest of the codebase
* closes #432 

Screenshot 2019-08-27 at 18 03 50


invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/DocumentDetails/components/DocumentStats/DocumentStats.js
```diff
@@ -28,8 +28,15 @@ export default class DocumentStats extends Component {
...
   renderStats() {
-    const { isLoading, error, data } = this.props;
-    let renewalCount = sumBy(data.hits, loan => loan.metadata.extension_count);
+    const { isLoading, error, data, document } = this.props;
+    const pastLoans = data.total;
+    const renewalCount = sumBy(
+      data.hits,
+      loan => loan.metadata.extension_count
+    );
+    let avgLoans = pastLoans / document.metadata.items.total;
```
ntarocco: It took me a while to understand the need of `isFinite` and `Math.round` :)

What about:
```
const pastLoans = data.total || 0;
const avgLoans = document.metadata.items.total ? (pastLoans / document.metadata.items.total).toFixed(2)  : "-"
```

ntarocco (commented):
Can you please add a small "info" button as in the screenshot close to `average` table header label to explain that it is an approximation counting all items and not only the the one in the range date?

Screenshot 2019-08-28 at 08 43 57

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/DocumentDetails/components/DocumentStats/DocumentStats.js
ntarocco: To be honest, I would be tempted to improve the search query and instead of using `document.metadata.items.total` I would probably calculate:

```
itemApi
      .list(
        itemApi
          .query()
          .withStatus(invenioConfig.items.available.status)
          .qs()
      )
```

equadon: Maybe even one decimal is enough for this value?

topless: @ntarocco to clarify a bit this point for my better understanding. `document.metadata.items.total` is the result of the query `item_search.search_by_document_pid(document_pid).count()`. If I understood correctly you are suggesting here to re-request the number of items, what would be the difference?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/DocumentDetails/components/DocumentStats/DocumentStats.js
```diff
@@ -38,12 +53,21 @@ export default class DocumentStats extends Component {
               <Table.Row>
                 <Table.HeaderCell>past loans</Table.HeaderCell>
                 <Table.HeaderCell>renewals</Table.HeaderCell>
+                <Table.HeaderCell>
+                  average{' '}
+                  <Popup
+                    position="top right"
+                    content="This is the average number of loans per document item"
```
ntarocco: I think we should be more explicit here. The point is to say that it is not exact, because the number of items can vary.
`Note that this average is computed with the current number of items ({itemsCount}) for this book and not with the number of items in the selected range dates.`
@equadon @kprzerwa if you have better suggestions with english, please help.

@topless please show this only if the average is more than 0.

topless: @ntarocco I have rephrased it a bit. I found the negation of range dates a bit confusing when I read it for first time. 

*This average is computed with the number of past loans on the selected range of dates, and the current number of items (${itemsCount}) of the document.*

kpsherva (approved):

PR #469, opened 2019-08-28
Schema ui fixes

kpsherva:
Added:
testData dependant on the test files used for backend. It works like a fake integration test, if something changes in the schema on the backend (or frontend) the frontend tests will fail immediately (snapshots) and it should help us to spot the missing changes quicker.


invenio_app_ils/mappings/v6/documents/document-v1.0.0.json
```diff
@@ -18,29 +46,55 @@
...
+                "curated_relation": {
+                  "type": "boolean"
+                },
+                "value": {
+                  "type": "keyword"
+                }
+              }
+            },
+            "alternative_names": {
+              "type": "keyword"
```
topless, line 75: There are a lot of "type": "keywords", i.e. `volume`, `note` etc, which I think they should be `text`.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/DocumentDetails/components/DocumentMetadata/DocumentMetadata.js
```diff
@@ -317,7 +329,7 @@ export default class DocumentMetadata extends Component {
             <Grid.Column>
               <Container>
                 <Header as="h3">Abstract</Header>
-                <p>{document.metadata.abstracts}</p>
+                <p>{document.metadata.abstracts[0].value}</p>
```
topless, line 332: We can iterate on the abstracts here and display each one in a `p` element.

topless:
I have a more general comment for the schemas in general. It seems that the schemas start to grow wildly especially the document one. I have tried in the past to use `#definitions` but I didn't manage to make it work. Maybe we should consider breaking them down to smaller more reusable ones. 

Before that, an easy win we can do is to use a utility like [sort-json](https://github.com/richie5um/vscode-sort-json) for vscode, and sort alphabetically the properties.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/DocumentDetails/components/DocumentMetadata/DocumentMetadata.js
ntarocco: UX is gonna come especially when we have some real data :)

ntarocco (approved):

PR #490, opened 2019-09-12
Backend: DocumentRequest record type

equadon:
Add new record type for document requests.

* Custom search factory to filter document requests so patrons only see their own requests

Closes #462

PR #493, opened 2019-09-16
Backoffice: Document requests

### LINKED ISSUE #472: Request form for books not existing in the catalogue - backoffice
- [x] backoffice search page and details page for the book requests

https://codimd.web.cern.ch/0p-ZhZDnTAycgWGNULBXBQ#User-makes-a-request
https://cds.cern.ch/ill/article_request_step1

equadon:
Add backoffice UI for `DocumentRequest`.

Requires https://github.com/inveniosoftware/invenio-app-ils/pull/490
    
* Search and details page
* Document details now link to the request if it has a request
* Cancel request implemented
* Moved `CancelLoanModal` to `common/components/CancelModal` which is now used by both loans and document requests
* Delete document will fail if it came from a request that's not deleted

Search page:
doc-req-search

Details page (fulfilled):
doc-req-details-fulfilled

Details page actions:
doc-req-details-actions
    
Closes #472


invenio_app_ils/config.py
```diff
@@ -603,6 +613,51 @@ def _(x):
...
+        record_serializers={
+            "application/json": (
+                "invenio_app_ils.records.serializers:json_v1_response"
+            ),
+        },
+        search_serializers={
+            "application/json": (
+                "invenio_records_rest.serializers:json_v1_search"
+            ),
+            "text/csv": ("invenio_app_ils.records.serializers:csv_v1_search"),
```
ntarocco, line 642: If we add this `csv`, I would add also the `Export` button then to the Search page! :)

invenio_app_ils/mappings/v6/document_requests/document_request-v1.0.0.json
```diff
@@ -0,0 +1,61 @@
...
+          "type": "keyword"
+        },
+        "state": {
+          "type": "keyword"
+        },
+        "title": {
+          "type": "text"
+        },
+        "volume": {
+          "type": "keyword"
```
ntarocco: I would go for text here, I would say users will probably input something like `Volume 1`

invenio_app_ils/records/loaders/schemas/json/document_request.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018 CERN.
```
ntarocco: 2019

invenio_app_ils/schemas/document_requests/document_request-v1.0.0.json
```diff
@@ -0,0 +1,68 @@
...
+    },
+    "cancel_reason": {
+      "type": "string",
+      "title": "Reason for cancelling the document request"
+    },
+    "document_pid": {
+      "title": "The persistent identifier of the created Document",
+      "type": "string"
+    },
+    "isbn": {
```
ntarocco: I think it is missing the `document` `$ref` field, in the mappings too.

invenio_app_ils/records/loaders/schemas/json/document_request.py
```diff
@@ -0,0 +1,30 @@
...
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""DocumentRequest schema for marshmallow loader."""
+
+from invenio_records_rest.schemas.fields import PersistentIdentifier
+from marshmallow import Schema, fields
+
+
```
ntarocco: A few comments here:
* maybe I am wrong, but it is missing the `metadata` field with inside all these fields, no? to be discussed maybe face to face :)
* I would maybe inherit from `RecordMetadataSchemaJSONV1` and not directly `Schema` so you already get the `pid` field.
* I would change `document_pid = fields.Str()` to `document_pid = fields.Str(dump_only=True)`, WDYT?
* I would add `document = fields.Raw(dump_only=True)`

Can we sanitize these strings somehow? In `invenio-records-rest` we have `SanitizedHTML` and `SanitizedUnicode`, I am not sure it can be useful...

invenio_app_ils/records/loaders/schemas/json/document_request.py
```diff
@@ -0,0 +1,30 @@
...
+class DocumentRequestSchemaV1(Schema):
+    """Document Request schema."""
+
+    pid = PersistentIdentifier()
+    state = fields.Str(required=True)
+    patron_pid = fields.Str(required=True)
```
ntarocco: for the `patron_pid`, I think we have to create a specific validator to validate that if it is the user creating the request, it must be him, pseudo-code:

```
    @validates('patron_pid')
    def validate_patron_pid(self, value):
        if not current_user == admin or librarian:
            return patron_pid ==  current_user
```

invenio_app_ils/config.py
```diff
@@ -603,6 +613,51 @@ def _(x):
...
+        },
+        list_route="/document-requests/",
+        item_route="/document-requests/<{0}:pid_value>".format(
+            _DREQID_CONVERTER
+        ),
+        default_media_type="application/json",
+        max_result_window=_RECORDS_REST_MAX_RESULT_WINDOW,
+        error_handlers=dict(),
+        read_permission_factory_imp=DocumentRequestOwnerPermission,
+        list_permission_factory_imp=authenticated_user_permission,
```
ntarocco: just to confirm, `authenticated_user_permission` is also checking that I can see only my requests using `search_factory_imp`, right?

invenio_app_ils/config.py
equadon: It's already there and I tested it and works great! :)

invenio_app_ils/config.py
equadon: Yes, the custom search factory takes care of that. I believe this permission is only for the actual list endpoint.

invenio_app_ils/cli.py
```diff
@@ -661,6 +667,47 @@ def persist(self):
         return self.holder.related_records['objs']
 
 
+class DocumentRequestGenerator(Generator):
+    """Document requests generator."""
+
+    STATES = ["CANCELLED", "PENDING", "FULFILLED"]
```
ntarocco: can you please move this to the `api` class? It is important that these states are defined there.

invenio_app_ils/records/api.py
```diff
@@ -525,3 +550,80 @@ def create(cls, data, id_=None, **kwargs):
...
+        """Validate document request data."""
+        for op in ops:
+            if "state" in op["path"] and op["value"] == "FULFILLED":
+                # Ensure there is also a document_pid in the ops
+                found_document_pid = False
+                for op in ops:
+                    if "document_pid" in op["path"]:
+                        self._validate_fulfilled(document_pid=op["value"])
+                        found_document_pid = True
+                if not found_document_pid:
```
ntarocco: I am not sure I understand fully this case. Is it happening when it is already in FULFILLED state?

ntarocco (dismissed):

invenio_app_ils/records/api.py
equadon: This case would happen if you were to patch the `state` and change it to `FULFILLED` but not provide a `document_pid`. In that case, we validate it it with `document_pid=None` which will raise an exception.

ntarocco: Can I suggest something, again for next PR, just to improve readibility? (even if I am not sure it does improve :))

```
fulfilledStateOp = next(iter(filter(ops, lambda op: op.path == 'state' and op.value == 'FULFILLED'), None))
documentPidOp = next(iter(filter(ops, lambda op: path == 'document_pid'), None))
if fulfilledStateOp:
    self._validate_fulfilled(document_pid=documentPidOp["value"])
```

I am not a big fan of `next(iter))` but it is one line instruction to get the element (not a list) or `None`.
WDYT?

Maybe you can find a better solution too...

equadon: For me that looks good!

Another suggestion:
```python
fulfilled_state = any(op['path'] == '/state' and op['value'] == 'FULFILLED' for op in ops)
if fulfilled_state:
    document_pid = next((op['value'] for op in ops if op['path'] == '/document_pid'), None)
    self._validate_fulfilled(document_pid=document_pid)
```

invenio_app_ils/cli.py
zzacharo: Is this still relevant @equadon ?

zzacharo (dismissed):

invenio_app_ils/cli.py
equadon: It's been handled in the frontsite PR.

ntarocco (approved):

PR #494, opened 2019-09-16
fix: all no-unused-vars cause by for loops

topless:
The main reason for all these warnings is demonstrated below.

The following code does not consider usage of the variable `some`, which is defined in the loop scope, since it is just an index to our array, but no interaction with the actual variable.
```
for (let some in something) {
    someArray[some] = 'thing'
}
```

As a counter strategy for it, it is suggested to use `forEach`. I am not the biggest fan of forEach but I prefer it from warnings.

### DIFF
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESRelatedSelector.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/components/ESSelector/ESRelatedSelector.js
@@ -56,11 +56,11 @@ export default class ESRelatedSelector extends Component {
 
     result.metadata.extraFields = result.metadata.extraFields || {};
     if (!isEmpty(this.extraRefs)) {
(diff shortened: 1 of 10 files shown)


equadon:
Should we really update the code to adapt to a linter that reports incorrectly? Wouldn't it be better to pin the `eslint` version to before it got messed up? Next time they might create other false-positives and we have to update our code again.

ntarocco:
@topless @equadon it looks like that the fix (introduced in eslint) will come with the next version of [create-react-app 3.1.2](https://github.com/facebook/create-react-app/issues/7566).

I would say to close this PR and wait for the release.
WDYT?

topless:
@ntarocco sure thing, i believe the list impact can be done by switching `let` to `var` like its done here, but this will hide other issues, on multiple loops on a component etc. Lets leave it as is, more simple.

PR #512, opened 2019-09-23
bugs: results table

### LINKED ISSUE #337: refactor: ResultsTable component
[Here](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/ResultsTableBody.js#L33) we need to check if the value is boolean and handle it differently as if the value of the column is false it returns `-` and also when is true the value is not displayed correctly.

topless:
* fix in eitems now points to the right document_pid
* fix display of boolean values on table
* closes #337

Screenshot 2019-09-23 at 15 40 52

### DIFF
--- a/invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/ResultsTableBody.js
+++ b/invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/ResultsTableBody.js
@@ -1,50 +1,59 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { Button, Table } from 'semantic-ui-react';
+import isBoolean from 'lodash/isBoolean';
+import isNumber from 'lodash/isNumber';
 
 export default class ResultsTableBody extends Component {
-  renderCell = (cell, column, id, colIndex) => {
-    return (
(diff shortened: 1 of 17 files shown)


invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/ResultsTableBody.js
```diff
@@ -1,50 +1,59 @@
...
+import isNumber from 'lodash/isNumber';
 
 export default class ResultsTableBody extends Component {
-  renderCell = (cell, column, id, colIndex) => {
-    return (
-      <Table.Cell key={colIndex + '-' + id} data-test={column + '-' + id}>
-        {cell}
-      </Table.Cell>
-    );
+  renderValue = value => {
```
ntarocco, line 8: Why not having the formatters knowing how to format the values? Until now, the ResultsTable was not "caring" about how to render a value, but it was the responsability of the formatter.
I think, IMHO, it was more correct before with the formatter setting Yes or No (or maybe in the future we want a green tick or red cross)...

WDYT?

Issue #464, opened 2019-08-21
Create / update forms


kpsherva:
Backoffice UI:
Implement create/update forms for:

- [ ] ~~document~~ moved to another issue 
- [ ] ~~item~~ moved to another issue 
- [x] eitem
- [x] series
- [x] locations https://github.com/inveniosoftware/invenio-app-ils/issues/492
- [ ] keyword (tag) 
  - https://github.com/inveniosoftware/invenio-app-ils/issues/203
  - https://github.com/inveniosoftware/invenio-app-ils/issues/536
  - https://github.com/inveniosoftware/invenio-app-ils/issues/304

make sure the user has proper access rights for the operations. Both backend/frontend. The task is mainly frontend, most of the functionality on the backend should be handled by invenio-records-rest
not for loans

kpsherva:
task divided to smaller issues, closing

PR #515, opened 2019-09-24
backoffice: add series editor form

zzacharo:
addresses #464


invenio_app_ils/records/loaders/schemas/json/series.py
```diff
@@ -7,15 +7,25 @@
...
+class SeriesSchemaV1(RecordMetadataSchemaJSONV1, StrictKeysMixin):
     """Series schema."""
 
+    def get_pid_field(self):
+        """Return pid_field value."""
+        return "pid"
+
+    title = fields.Str(required=True)
     pid = PersistentIdentifier()
     mode_of_issuance = fields.Str()
```
ntarocco: `mode_of_issuance` is required too. @kprzerwa maybe you can help with the mandatory fields.
I think that the loader should then match the required fields in json schema.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/BackofficeRecordEditor/BackofficeRecordEditor.js
```diff
@@ -0,0 +1,37 @@
+// EditUserDialog.js
```
ntarocco: IMHO, I would get rid of this completely and just have a component inside Backoffice -> Series -> `NewSeries` or `EditSeries`, and they internally render a `SeriesForm` that is filled with values when editing.

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/BackofficeRecordEditor/series/SeriesFormContainer.js
```diff
@@ -0,0 +1,48 @@
+// EditUserDialog.js
```
ntarocco: Same here, I would move it with series.

Didn't we say to "kill" the name`Container`?

invenio_app_ils/ui/src/invenio_app_ils/pages/backoffice/BackofficeRecordEditor/series/SeriesFormContainer.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  renderEditForm = () => {
+    const { isLoading, error, data } = this.props;
+    return (
+      <Loader isLoading={isLoading}>
+        <Error error={error}>
+          <SeriesForm
+            data={data}
+            title="Edit serie"
```
ntarocco: I think `series` is always plural :)

invenio_app_ils/ui/src/invenio_app_ils/routes/urls.js
```diff
@@ -46,6 +46,10 @@ const BackOfficeRoutesList = {
   stats: {
     home: `${BackOfficeBase}/stats`,
   },
+  editor: {
```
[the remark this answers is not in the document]
zzacharo: Do you want to enlist everything? Or a common name?

invenio_app_ils/webpack.py
```diff
@@ -18,10 +18,12 @@
     dependencies={
         "axios": "^0.19.0",
         "formik": "^1.5.8",
+        "iso-639-1": "^2.1.0",
```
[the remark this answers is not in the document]
zzacharo: @ntarocco Yes I thought for now, so the form can be complete, to use something. Definitely an entrypoint serving the enums is the best solution. Should I remove it?

invenio_app_ils/records/api.py
```diff
@@ -104,6 +104,16 @@ def relations(self):
         """Get record relations."""
         return self._relations
 
+    def clear(self):
+        """Clear IlsRecordWithRelations record."""
+        relations = self["relations"]
```
ntarocco: you only need `relations_metadata`, `relation` field is normally a JSON resolver no?

See `class Series` `create` method:
* `relations` are added in the `create` method
* `tags` are actually not added, a but?

I think that resolver fields should be added in the `create` + `update` methods to ensure that they are there.
WDYT?

invenio_app_ils/records/loaders/schemas/json/series.py
ntarocco: I am not sure this has been fixed

invenio_app_ils/ui/src/invenio_app_ils/common/api/series/series.js
```diff
@@ -139,9 +151,11 @@ const count = query => {
 
 export const series = {
   url: apiURL,
+  create: create,
   get: get,
   delete: del,
   patch: patch,
+  put: put,
```
ntarocco: I would expose the `HTTP verb`: it is better to expose APIs that are abstracting the internal implementation/action: `put` -> `update/edit`?

Probably you can now remove `patch` both here and in the config.py `RECORDS_REST_ENDPOINTS`

invenio_app_ils/records/api.py
zzacharo: @ntarocco `Relation` is indeed a jsonresolver. I am just thinking that if we move it in the update then you could not be able to override the field programmatically as it would always keep the first assigned value when it was created...I suppose this is fine. 

In series we have `tags`???

invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/formatters.js
```diff
@@ -180,10 +180,7 @@ function formatRelatedToTableView(related, relation) {
   const obj = {
     ID: related.pid,
     Type: formatPidTypeToName(related.pid_type),
-    Title:
-      related.title && related.title.title
-        ? related.title.title
-        : related.title,
+    Title: related.title && related.title ? related.title : related.title,
```
kpsherva: 🤣

PR #516, opened 2019-09-24
337 refactor results table

topless:


ntarocco:
is this still relevant?

topless:
@ntarocco It is relevant but probably will not come as one pull request, but partially as we are simplifying the tables while we iterate over them. The ResultLists are tightly coupled with the tests which made them a little bit trickier to throw away.

PR #529, opened 2019-09-30
global: minor fixes

ntarocco:


ntarocco:
@kprzerwa we will have to check if and while is still happening that you have the namedtuple as dict before merging this.

zzacharo (approved):

topless (approved):

PR #537, opened 2019-10-06
Sidebar for backoffice

topless:
According to your yesterday conversation there are a couple of things missing from this pr.

1) The `actions` section in the sidebar (request a loan, checkin, checkout) due to the fact that I am not sure where to point them.
2) Sidebar currently cannot be extended. In my head is not so trivial just to provide an array of elements and render them in the sidebar. I think its easier to support a sidebar replacement (provide your own sidebar implementation) rather than trying to extend this one.

Cheers,

sidebar_640


ntarocco:
Nice job! It looks good now! :) 👍 
I am not 100% sure about the order of the menu items (e.g. Locations before Catalogue) but I think we will iterate on this after some feedback from library/others!

invenio_app_ils/ui/src/invenio_app_ils/common/components/ResultsTable/formatters.js
```diff
@@ -6,8 +6,9 @@ import assign from 'lodash/assign';
 import { formatPidTypeToName } from '../ManageRelationsButton/utils';
 import { loan as loanApi } from '../../api';
 import { BackOfficeRoutes, FrontSiteRoutes } from '../../../routes/urls';
-import { invenioConfig } from '../../config';
+import { invenioConfig, TRUNCATE_LENGTH } from '../../config';
```
ntarocco, line 9: If you do this, then I would say that for backoffice and frontsite it will be different.
I guess it will be better to have a small config for backoffice only then...

ntarocco (approved):

zzacharo (approved):

invenio_app_ils/ui/src/invenio_app_ils/routes/components/BackOffice/BackOffice.js
```diff
@@ -1,181 +1,42 @@
...
+          </Grid.Column>
+
+          {/* Content */}
+          <Grid.Column
+            stretched
+            width={13}
+            style={contentStyle}
+            verticalAlign={'top'}
+          >
+            <BackOfficeContent />
```
kpsherva: just minor naming: I would name it with `BackOfficeRoutingSwitch` or something similar because IMO the component is not having content inside

topless: `BackOfficeRoutesSwitch` it is

PR #553, opened 2019-10-14
ui: prepare ui to be deployed as spa

zzacharo:


kpsherva (approved):

invenio_app_ils/ui/.env
```diff
@@ -0,0 +1,6 @@
+REACT_APP_JWT_TOKEN='token'
```
ntarocco, line 1: You can remove `REACT_APP_JWT_TOKEN` and `REACT_APP_JWT_TOKEN_EXPIRATION`, I think they are not used at all... WDYT?

invenio_app_ils/ui/.env
```diff
@@ -0,0 +1,6 @@
...
+REACT_APP_USER_ID=1
+REACT_APP_LOCATION_ID=1
+REACT_APP_USER_ROLE=librarian
+REACT_APP_JWT_TOKEN_EXPIRATION=11111111111111111111111
+REACT_APP_JWT_USERNAME=librarian
```
ntarocco, line 6: I am not sure we have an username...

invenio_app_ils/ui/src/invenio_app_ils/common/config/invenioConfig.js
```diff
@@ -1,7 +1,200 @@
...
+            field: 'bestmatch',
+            order: 2,
+            title: 'Best match',
+          },
+        ],
+      },
+      sortOrder: ['asc', 'desc'],
+    },
+  },
+  support_email: 'info@inveniosoftware.org',
```
ntarocco, line 199: this `support_email` might be something you will want to inject from the env or a js when deploying in a specific env. Did you figure out a way to do that?

setup.py
```diff
@@ -65,7 +65,7 @@
     "invenio-records>=1.3.0,<1.4.0",
     # upgraded packages
     "invenio-accounts-rest>=1.0.0a4,<1.1.0",
-    "invenio-app>=1.2.1,<1.3.0",
+    "invenio-app>=1.2.3,<1.3.0",
```
ntarocco, line 68: This is great, but I don't see that in the config.py you override the flask static vars. I think they should be here, so that we can reproduce locally the prod env via docker-compose full

invenio_app_ils/ui/.env
zzacharo: I think we had added it in some point but the backend code is not there anymore...I just left whatever was there just not to break UI for now...This file will go away once we implement the oauth workflow during the sprint.

invenio_app_ils/ui/.env
zzacharo: I left it for the same reason as below :)

ntarocco (approved):

setup.py
zzacharo: After discussion we will not put them in the config as invenio expects them from the environment.

PR #565, opened 2019-10-18
fix: card image size

topless:


ntarocco:
For reviewers: I merged this because master was broken without.

PR #578, opened 2019-10-23
docker: refactor docker-compose.full.yml to latest spa deployment

### LINKED ISSUE #535: Deploy UI as SPA
* move UI config in the React app
* modify the `docker-compose` to serve the SPA in the "proxy" container (nginx) and the invenio app as before.

See whiteboard steps
See: https://create-react-app.dev/docs/deployment

zzacharo:
closes #535

### DIFF
--- a/docker/nginx/conf.d/default.conf
+++ b/docker/nginx/conf.d/default.conf
@@ -9,13 +9,13 @@
 # 443. All traffix on port 80 is redirect to port 443 on SSL.
 #
 # Nginx proxies all requests on port 443 to upstream the application server
-# which is expected to be running on port 5000/5001.
+# which is expected to be running on port 5000.
 
 upstream ui_server {
   server web-ui:5000 fail_timeout=0;
 }
 upstream api_server {
-  server web-api:5001 fail_timeout=0;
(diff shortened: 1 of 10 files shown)


docker/nginx/conf.d/default.conf
```diff
@@ -57,16 +58,23 @@ server {
...
-  proxy_request_buffering off;
+
+  # proxy_request_buffering off;
   # Sets the HTTP protocol v1.1 for proxying in order to not use the buffer
   # in case of chunked transfer encoding
-  proxy_http_version 1.1;
+
+  # proxy_http_version 1.1;
+
+  location = /ping {
```
ntarocco: I think we should remove this from here and pass it below in the `location ~ ^/(login|`?

ntarocco (approved):

docker/nginx/conf.d/default.conf
zzacharo: I had troubles with the uwsgi and the ping that's why I put it there. I need to re test if this is fine.

zzacharo: Changed it!

equadon (approved):

PR #585, opened 2019-10-28
convert functional to class components

topless:


ntarocco (approved):
It is a bit hard to understand what it has been changed in the UI, if all parts are there, but I guess there is no feature change here, just a refactoring.

equadon (approved):

topless:
> It is a bit hard to understand what it has been changed in the UI, if all parts are there, but I guess there is no feature change here, just a refactoring.

Well nothing changed in the ui, and nothing changed in general, just instead of functions we use classes for the components.

PR #595, opened 2019-10-31
stats: record-views

### LINKED ISSUE #167: [Req] Statistics: record views
Track any view to the frontsite document details page

Backend:
- [x] integrate `invenio-stats`
- [x] possible approach: create a new backend endpoint to "record" the view
- [x] in the view, send a signal to `invenio-stats`

Frontend:
- [x] display record views and unique views in a component

topless:
- requires https://github.com/inveniosoftware/invenio-stats/pull/100
- closes #167 

If the document has `metadata.eitems.hits` the downloads stats appear.

Screenshot 2019-11-06 at 13 45 31

### DIFF
--- a/invenio_app_ils/records/views.py
+++ b/invenio_app_ils/records/views.py
@@ -24,10 +24,10 @@ from invenio_app_ils.proxies import current_app_ils_extension
 from invenio_app_ils.signals import record_viewed
 
 
-def create_record_stats_blueprint(app):
-    """Add record stats endpoint to the blueprint."""
+def create_document_stats_blueprint(app):
+    """Create document stats blueprint."""
     blueprint = Blueprint(
-        "invenio_app_ils_record_stats", __name__, url_prefix=""
+        "ils_document_stats", __name__, url_prefix=""
     )
     endpoints = app.config.get("RECORDS_REST_ENDPOINTS", [])
     options = endpoints.get(DOCUMENT_PID_TYPE, {})
@@ -37,45 +37,47 @@ def create_record_stats_blueprint(app):
         mime: obj_or_import_string(func)
         for mime, func in rec_serializers.items()
     }
+
+    stats_view = DocumentStatsResource.as_view(
+        DocumentStatsResource.view_name.format(DOCUMENT_PID_TYPE),
+        serializers=serializers,
+        default_media_type=default_media_type,
+    )
     blueprint.add_url_rule(
-        "{0}/record-stats".format(options['item_route']),
-        view_func=RecordStatsResource.as_view(
-            RecordStatsResource.view_name,
-            serializers=serializers,
-            default_media_type=default_media_type,
-        ),
+        "{0}/stats".format(options['item_route']),
+        view_func=stats_view,
         methods=["POST"],
     )
     return blueprint
 
 
-class RecordStatsResource(ContentNegotiatedMethodView):
-    """Endpoint to trigger record stats signal."""
+class DocumentStatsResource(ContentNegotiatedMethodView):
+    """Document stats resource."""
 
-    view_name = "record_stats"
+    view_name = "{}_stats"
 
     @pass_record
     def post(self, pid, record, **kwargs):
         """Send a signal to count record view for the record stats."""
         data = request.get_json()
-        if data.get('stat') == 'record-view':
+        if data.get('event') == 'record-view':
             record_viewed.send(
                 current_app._get_current_object(),
                 pid=pid,
                 record=record,
             )
             return self.make_response(pid, record, 202)
-        return abort(400)
+        return DocumentRequestError("Invalid record view request")
 
 
 def create_document_request_action_blueprint(app):
-    """Add document request action views to the blueprint."""
+    """Create document request action blueprint."""
     blueprint = Blueprint(
-        "invenio_app_ils_document_request", __name__, url_prefix=""
+        "ils_document_request", __name__, url_prefix=""
     )
 
     endpoints = app.config.get("RECORDS_REST_ENDPOINTS", [])
-    options = endpoints.get("dreqid", {})
+    options = endpoints.get(DOCUMENT_REQUEST_PID_TYPE, {})
     default_media_type = options.get("default_media_type", "")
     rec_serializers = options.get("record_serializers", {})
     serializers = {
@@ -87,7 +89,6 @@ def create_document_request_action_blueprint(app):
         AcceptRequestResource.view_name.format(DOCUMENT_REQUEST_PID_TYPE),
         serializers=serializers,
         default_media_type=default_media_type,
-        ctx=dict(),
     )
     blueprint.add_url_rule(
         "{0}/accept".format(options["item_route"]),
@@ -99,7 +100,6 @@ def create_document_request_action_blueprint(app):
         RejectRequestResource.view_name.format(DOCUMENT_REQUEST_PID_TYPE),
         serializers=serializers,
         default_media_type=default_media_type,
-        ctx=dict(),
     )
     blueprint.add_url_rule(
         "{0}/reject".format(options["item_route"]),
@@ -110,17 +110,7 @@ def create_document_request_action_blueprint(app):
     return blueprint
 
 
-class DocumentRequestActionResource(ContentNegotiatedMethodView):
-    """Document request action resource."""
-
-    def __init__(self, serializers, ctx, *args, **kwargs):
-        """Constructor."""
-        super(DocumentRequestActionResource, self).__init__(
-            serializers, *args, **kwargs
-        )
-
-
-class AcceptRequestResource(DocumentRequestActionResource):
+class AcceptRequestResource(ContentNegotiatedMethodView):
     """Accept document request action resource."""
 
     view_name = "{}_accept_request"
@@ -132,7 +122,7 @@ class AcceptRequestResource(DocumentRequestActionResource):
         raise DocumentRequestError("Accept request is not implemented")
 
 
-class RejectRequestResource(DocumentRequestActionResource):
+class RejectRequestResource(ContentNegotiatedMethodView):
     """Reject request resource."""
 
     view_name = "{}_reject_request"
--- a/invenio_app_ils/records/views.py
+++ b/invenio_app_ils/records/views.py
@@ -9,7 +9,7 @@
 
 from __future__ import absolute_import, print_function
 
-from flask import Blueprint, request
+from flask import Blueprint, abort, current_app, request
 from invenio_db import db
 from invenio_records_rest.utils import obj_or_import_string
 from invenio_records_rest.views import pass_record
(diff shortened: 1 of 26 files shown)


topless:
There are a couple of files that were formatted automatically from my editor onLostFocus because I opened them, sorry for that.

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/Documents/DocumentsDetails/DocumentMetadata/components/DocumentStats.js
```diff
@@ -0,0 +1,84 @@
...
+    const request = {
+      views: {
+        stat: 'record-view',
+        params: {
+          pid_value: document.pid,
+        },
+      },
+    };
+    try {
+      const response = await http.post('/stats', request);
```
kpsherva: can you add this to URLs list so it's not hardcoded?

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/Documents/DocumentsDetails/DocumentMetadata/components/DocumentStats.js
```diff
@@ -0,0 +1,84 @@
...
+  };
+
+  render() {
+    const { document } = this.props;
+    const { data, error } = this.state;
+    const hasItems = document.metadata.items.hits.length || false;
+    const hasEitems = document.metadata.eitems.hits.length || false;
+    return (
+      <>
+        {(hasItems || hasEitems) && (
```
kpsherva: what if it had items and eitems but they were deleted? also those are views for the document so i'm not sure why do you base it on existence of items but maybe I'm missing something

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/components/BookCard/BookCard.js
```diff
@@ -29,29 +29,30 @@ export class BookCard extends Component {
...
+          {data.metadata.document_type}
+        </Card.Meta>
         <Image
           centered
           src={getCover(data.metadata.pid)}
           size={'small'}
           onError={e => (e.target.style.display = 'none')}
         />
-        <Card.Content centered={'true'}>
+        <Card.Content>
```
kpsherva, line 41: why not centered anymore?

invenio_app_ils/records/views.py
```diff
@@ -18,8 +18,53 @@
...
+    endpoints = app.config.get("RECORDS_REST_ENDPOINTS", [])
+    options = endpoints.get(DOCUMENT_PID_TYPE, {})
+    default_media_type = options.get("default_media_type", "")
+    rec_serializers = options.get("record_serializers", {})
+    serializers = {
+        mime: obj_or_import_string(func)
+        for mime, func in rec_serializers.items()
+    }
+    blueprint.add_url_rule(
+        "{0}/stats/record-view".format(options['item_route']),
```
equadon: Will there be more types of stats recorded other than views? If so, perhaps it might be quicker to have one API endpoint for recording stats and just passing the name of the event `record-view` with the `POST` request and handling it in the same class but different methods.

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/components/BookCard/BookCard.js
topless: `centered` is not a property of Card.Content. In the console there was this `Warning: Received `true` for a non-boolean attribute ...` that's why I removed it. They are still centered from the prop of Card.

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/Documents/DocumentsDetails/DocumentMetadata/components/DocumentStats.js
topless: I was looking for a way to show/hide the downloads thing and got carried away, I will make the download stats available only if there are eitems present.

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/components/Document/DocumentStats.js
```diff
@@ -0,0 +1,77 @@
...
+    const request = {
+      views: {
+        stat: 'record-view',
+        params: {
+          pid_value: document.pid,
+        },
+      },
+    };
+    try {
+      const response = await http.post(FrontSiteRoutes.statsUrl, request);
```
ntarocco: Just reading this, I think it should be: `documentApis.statsUrl` no?

setup.py
```diff
@@ -84,10 +84,13 @@
     "invenio-userprofiles>=1.0.1,<1.1.0",
     # extra
     "invenio-circulation>=1.0.0a19,<1.1.0",
+    "invenio-stats>=1.0.0a12",
     # until flask-sqlalchemy is fixed
     "SQLAlchemy>=1.2.16,<1.3.0",
     "invenio-pidrelations>=1.0.0a6,<1.1.0",
     "sentry-sdk>=0.10.2",
+    # kombu 4.6.5 requires amqp==2.5.1
+    "amqp==2.5.1",
```
ntarocco: don't need anymore?

ntarocco (commented):
In general, is `stats` too generic? Do we want something more specific around `record-views`? Even if there will be downloads too later on...

invenio_app_ils/signals.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018 CERN.
```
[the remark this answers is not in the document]
topless: fixed it but its everywhere, year updates can be part of the clean up tasks we run on the codebase, super unimportant :)

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/components/Document/DocumentStats.js
topless: I wasn't really sure where to put it, thanks!

setup.py
topless: I will remove it, but maybe this is the reason it was passing, we ll see

topless:
> In general, is `stats` too generic? Do we want something more specific around `record-views`? Even if there will be downloads too later on...

It is indeed, and it comes in conflict with circulation which also has stats. The view is also very explicit and can mislead, maybe RecordStats can make some sense, since these stats are also records.

invenio_app_ils/records/views.py
```diff
@@ -18,8 +18,55 @@
...
+        """Send a signal to count record view for the record stats."""
+        data = request.get_json()
+        if data.get('stat') == 'record-view':
+            record_viewed.send(
+                current_app._get_current_object(),
+                pid=pid,
+                record=record,
+            )
+            return self.make_response(pid, record, 202)
+        return abort(400, "The proprety stat: `{}` is not supported".format(
```
ntarocco: Typo `proprety`. I would say, just return 400 without much details, what do you think?
I was thinking to the 400 just because if it happens we will have a Sentry exception and we can check...

invenio_app_ils/ui/src/invenio_app_ils/routes/urls.js
```diff
@@ -110,3 +110,13 @@ export const BackOfficeRoutes = {
   ...BackOfficeRoutesList,
   ...BackOfficeRouteGenerators,
 };
+
+const documentDetailsRecordViewStats = `/documents/:documentPid/record-stats`;
+
+export const DocumentApis = {
```
ntarocco: Sorry, I was not clear, I meant to add the action in the `document APIs` as for all other `REST` calls:
https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/ui/src/invenio_app_ils/common/api/documents/document.js#L180

ntarocco (commented):
@topless please review [my second commit](https://github.com/inveniosoftware/invenio-app-ils/pull/595/commits/3d60ab86687e96f54a507d0dc21ae2bc7620de0d), I had to change a few things and I will merge, we take any comment in another PR:
1. the submit stats was not working for me, because you were sending to `/record-stats` the request to get the stats, but `invenio-stats` is registering it under `/stats`, I had `404`.
2. when sending a "view" event, you have to add `pid_type` and not only `pid_value` to avoid any nasty bug in case we add other stats
3. for the ui app, I refactored the URLs as I was suggesting in the comments: REST endpoints should in the `api` module and not in the `urls.js` using the views urls.

invenio_app_ils/ui/src/invenio_app_ils/common/api/stats/stats.js
```diff
@@ -41,8 +41,22 @@ const getMostLoanedDocuments = async (fromDate, toDate) => {
...
+  const data = {
+    views: {
+      stat: 'record-view',
+      params: {
+        pid_type: pidType,
+        pid_value: pidValue,
+      },
+    },
+  };
+  return await http.post('/stats', data);
```
kpsherva, line 54: which url is the correct one ? `circulation/stats/` or `/stats`?

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/components/Document/DocumentStats.js
```diff
@@ -19,59 +16,54 @@ export class DocumentStats extends Component {
...
-      this.setState({ data: response.data });
+      const response = await stats.recordStats(pidType, document.pid);
+      const views = _get(response.data, 'views', {
+        count: '-',
+        unique_count: '-',
+      });
+      this.setState({ views: views });
     } catch (error) {
-      this.setState({ error: error });
+      console.warn(error);
```
kpsherva, line 28: just to understand: why we don't show the error to the user?

kpsherva (approved):

invenio_app_ils/ui/src/invenio_app_ils/common/api/stats/stats.js
ntarocco: Well, it might be not super clear :)
This `stats.js` contains multiple endpoints for stats, not only `circulation` stats. Or we split it in 2, e.g. `circulation_stats.js` and `records_stats.js` or keep everything there.
For now, I put record stats together with circulation stats, as they are stats :)
So, for `most-loaned`, it is `/circulation/stats` (maybe we should have `/circulation/loans/stats`?) and for record, `invenio-stats` registers the views under `/stats`.

invenio_app_ils/ui/src/invenio_app_ils/pages/frontsite/components/Document/DocumentStats.js
ntarocco: I removed the error because it was very weird: after loading the page, suddenly at the bottom you have a yellow box "Something went wrong" and I have no idea what is there or what went wrong!
I think the error should be displayed after an user action, not randomly. In this case it is not an user action, so we should not scare the user. I think it is better to just show "-" instead of the number and fix the issue when happens on our own.
WDYT?

kpsherva: true, It also does not make sense to block the whole page because of the error in statistics

invenio_app_ils/ui/src/invenio_app_ils/common/api/stats/stats.js
kpsherva: I see, maybe just a small comment would help to tell which is which (but it is NIT)

ntarocco: Probably it is better to split then... I will leave it to Zach :)

PR #600, opened 2019-11-04
[On hold]Fix backoffice sidebar

kpsherva:


ntarocco (commented):
The issue with the Grid, and that's why I have removed it, is that it has not a fixed width col.
So the `{2}` is too small, the `{3}` is too big.
On top of that I did not want to play with negative margins :)
I am happy to discuss if you want to demo it.

kpsherva:
decided to go with another solution

Issue #222, opened 2019-03-27
Documentation: describe the system


ntarocco:
Here some other tools to check features and documentation:
https://www.goodfirms.co/blog/best-free-open-source-library-management-software-solutions

PR #610, opened 2019-11-08
Vocabularies

### LINKED ISSUE #543: Vocabularies
For many of the fields in the different records, we need to implement vocabularies so that the user can choose a value from a list via a simple `<select>` component. Examples could be: list of languages, list of countries, list of subjects...

**Requirements**
* as a sys admin, I need to be able to import a new vocabulary (set of values) for a given field.
* as a sys admin, I need to easily delete/edit/add a new item to a vocabulary list
* as a librarian, I need to be able to select a value from a vocabulary for a given record field

**Tasks**
* identify all the fields that need a vocabulary in ILS and identify any potential exception
* design and implement the solution, backend
Proposal: have a CLI command to import/sync a JSON for a given vocabulary. JSON structure:
```
[{
    type: language,
    id: "en",
    value: "en",
    text: "English"
}, ...]
```
Index this list in a dedicated ES index accessible from Invenio, e.g. `/api/vocabularies/?q=type:language`. No records needed. Similar to Patron implementation.
* for these fields, add the dropdowns with search and autocompletion in the frontend app

equadon:
* Backend: added `/api/vocabularies` endpoint
* Backoffice: updated forms to use vocabularies
* Upgraded NPM packages

Closes #543 

Followup issues:
* Delete tag record: https://github.com/inveniosoftware/invenio-app-ils/issues/611
* Add document licenses: https://github.com/inveniosoftware/invenio-app-ils/issues/613


invenio_app_ils/ui/src/invenio_app_ils/forms/components/CountryField.js
```diff
@@ -0,0 +1,329 @@
...
+    return (
+      <div className={classes} role="alert" aria-live="polite" aria-atomic>
+        {_flag}
+        {_text}
+      </div>
+    );
+  };
+}
+
+export const FLAG_NAMES = [
```
kpsherva: is there any chance we can take them from a vocabulary?

invenio_app_ils/ui/src/invenio_app_ils/forms/core/VocabularyField.js
```diff
@@ -0,0 +1,130 @@
...
+      fieldPath,
+      label,
+      multiple,
+      serializer,
+      type,
+      ...uiProps
+    } = this.props;
+    const { isLoading, options } = this.state;
+    const noResultsMessage = isLoading
+      ? 'Loading vocabularies...'
```
kpsherva: for the user's perspective I would display `Loading options...` I don't think they will know what is a vocabulary

invenio_app_ils/ui/src/semantic-ui/site/collections/form.overrides
```diff
@@ -2,6 +2,12 @@
          Site Overrides
 *******************************/
 
+.field .ui.dropdown {
+  .ui.label.error {
+    background-color: @dropdownMissingColor;
```
kpsherva: I suggest to rename it to `@dropdownErrorColor`

invenio_app_ils/cli.py
```diff
@@ -845,9 +846,15 @@ def create_userprofile_for(email, username, full_name):
     "--skip-demo-data", is_flag=True, help="Skip creating demo data."
 )
 @click.option("--skip-patrons", is_flag=True, help="Skip creating patrons.")
+@click.option(
+    "--skip-vocabularies",
+    is_flag=True,
```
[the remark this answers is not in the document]
equadon: No, it will import the demo vocabularies by default and only ignore it if you provide `--skip-vocabularies`.

invenio_app_ils/ui/src/invenio_app_ils/forms/components/CountryField.js
equadon: The problem is that semantic UI throws an exception if it gets an incorrect flag name so I need to compare with the list of valid names. I just saw that semantic exports the list of valid flag names so I'll use that instead.

**Edit**: I've tried to import the available flag names from https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Flag/Flag.js but with no success. Any ideas?

INSTALL.rst
```diff
@@ -170,3 +170,37 @@ In addition to the normal ``docker-compose.yml``, this one will start:
 - UWSGI (application container)
 - Celery (background task worker)
 - Flower (Celery monitoring)
+
+
+Vocabularies
+------------
+Vocabularies are indexed using Elasticsearch and can be indexed from a JSON
```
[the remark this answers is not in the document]
equadon: Yes, I agree but it probably shouldn't be in `INSTALL.rst`. It would make more sense to put that general terminology documentation somewhere else.

invenio_app_ils/ui/src/invenio_app_ils/forms/components/CountryField.js
ntarocco: If you don't like it, we can get rid of the flags! :)

invenio_app_ils/vocabularies/cli.py
```diff
@@ -0,0 +1,122 @@
...
+    """Index JSON-based vocabularies in Elasticsearch."""
+    if not force:
+        click.confirm(
+            "Are you sure you want to index the vocabularies?",
+            abort=True
+        )
+    source = "json"
+    index_count = 0
+    for filename in filenames:
+        print('indexing vocabularies in {}...'.format(filename))
```
ntarocco: nit: Is `print` ok to be used? Isn't it better to use `click.echo`?

invenio_app_ils/vocabularies/data/author_affiliation_identifier_schemes.json
```diff
@@ -0,0 +1,107 @@
...
+  },
+  {
+    "type": "author_affiliation_identifier_scheme",
+    "key": "Editor",
+    "text": "Editor"
+  },
+  {
+    "type": "author_affiliation_identifier_scheme",
+    "key": "HostingInstitution",
+    "text": "Hosting Institution"
```
ntarocco: isn't this one the `author type`? https://schema.datacite.org/meta/kernel-4.3/include/datacite-contributorType-v4.xsd

invenio_app_ils/vocabularies/data/author_types.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "type": "author_type",
```
ntarocco, line 3: this one maybe just `affiliation_type`?

invenio_app_ils/vocabularies/sources/base.py
```diff
@@ -0,0 +1,47 @@
...
+    def __init__(self, source_type):
+        """Initialize vocabulary source."""
+        self.source_type = source_type
+
+    @staticmethod
+    def validate(vocabulary):
+        """Validate vocabulary."""
+        if not isinstance(vocabulary, Vocabulary):
+            raise VocabularyError(
+                "{!r} is not a vocabulary".format(vocabulary))
```
ntarocco: Didn't know about the `!r`! :)

equadon: Completely useless now since I didn't include the `__str__` and `__repr__` implementations.. :)

INSTALL.rst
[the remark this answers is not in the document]
ntarocco: Good point! I propose that we will move it when we will write the documentation for it.
https://github.com/inveniosoftware/invenio-app-ils/issues/222

invenio_app_ils/config.py
```diff
@@ -678,12 +678,12 @@ def _(x):
         record_class=Vocabulary,
         record_serializers={
             "application/json": (
-                "invenio_records_rest.serializers" ":json_v1_response"
+                "invenio_app_ils.records.serializers:json_v1_response"
```
ntarocco, line 681: @equadon I had to change the serializers otherwise it was not working, 500 when fetching the vocabularies... was it working with you with default serializers?
Python 3 in your venv?

invenio_app_ils/ui/src/invenio_app_ils/forms/components/CountryField.js
```diff
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { VocabularyField } from '../core';
 import { Dropdown, Flag } from 'semantic-ui-react';
+import { names as FLAG_NAMES } from 'semantic-ui-react/dist/commonjs/elements/Flag/Flag';
```
ntarocco, line 5: @equadon @kprzerwa this is the only way I found to avoid the list.
Given that there are no future releases (kind of dead project), I guess it is pretty safe... And also if the path is not found, there is a build error...
They exported values in TypeScript but in JS :(

What do you think?

invenio_app_ils/vocabularies/data/alternative_title_types.json
```diff
@@ -1,22 +1,22 @@
 [
   {
     "type": "alternative_title_type",
-    "key": "AlternativeTitle",
+    "key": "ALTERNATIVE_TITLE",
```
ntarocco, line 4: @equadon @kprzerwa what do you thing to use the uppercase instead? The advantage it that checking the record metadata we can guess what is vocabulary value and what is user input...

invenio_app_ils/ui/src/invenio_app_ils/forms/core/VocabularyField.js
```diff
@@ -37,19 +35,11 @@ export class VocabularyField extends React.Component {
...
-        options.push({
-          key: '',
-          value: '',
-          text: '-',
-        });
-      }
       for (let hit of response.data.hits) {
         options.push(serializer(hit));
       }
-      if (this._mounted) {
```
ntarocco: @equadon why did you need this `mounted`?

invenio_app_ils/config.py
equadon: Hm strange. Yes, everything was working when I was running it but this change seems good.

invenio_app_ils/ui/src/invenio_app_ils/forms/components/CountryField.js
equadon: Thank you! I tried to make that work but wasn't successful. It's not optimal but given how the exports work we don't have many choices.

invenio_app_ils/vocabularies/data/alternative_title_types.json
equadon: We use uppercase elsewhere for constants in ILS so we should continue doing that I think.

Edit: I also updated the language keys to be uppercase to keep it consistent.

invenio_app_ils/ui/src/invenio_app_ils/forms/core/VocabularyField.js
equadon: Now that accordion fields are optimized to only render if they are open it might be that you open a vocabulary field inside an accordion and if you close it before the vocabulary field has called `this.setState` you'll get an error that you can't update the state for an unmounted component.

PR #635, opened 2019-12-02
refactor: deleted tag record

topless:
Tag is not a record any longer, but its loaded from a vocabulary. We will have to address way it looks in the facets which I think aligns nicely with what @kprzerwa has done!


invenio_app_ils/cli.py
```diff
@@ -267,58 +239,56 @@ def persist(self):
...
 
 
+
+with open('invenio_app_ils/vocabularies/data/tags.json') as f:
+    TAGS = json.loads(f.read())
+
+
+def get_random_tags():
+    random.shuffle(TAGS)
+    return [tag["key"] for tag in TAGS[0:randint(1, len(TAGS) - 1)]]
```
equadon: Another way of doing this:
```python
return [tag["key"] for tag in random.sample(TAGS, randint(1, len(TAGS) - 1))]
```

equadon (approved):

ntarocco (commented):
From a quick search , I still see some traces of tags:
* `document.test.js`: `.withTag({ name: 'tag1' })`
* `document.js`: `withTagQuery.push(`tags.name:"${tag.name}"`);`
* `series js`: `tags.name`
or maybe I missed the deletion?

topless:
> From a quick search , I still see some traces of tags:
> 
> * `document.test.js`: `.withTag({ name: 'tag1' })`
> * `document.js`: `withTagQuery.push(`tags.name:"${tag.name}"`);`
> * `series js`: `tags.name`
>   or maybe I missed the deletion?

You are right I missed these

invenio_app_ils/cli.py
```diff
@@ -669,11 +627,14 @@ def data(
 
     indexer = RecordIndexer()
 
+    with open('invenio_app_ils/vocabularies/data/tags.json') as f:
```
ntarocco: this will not work when deployed, the path will be wrong. See https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/cli.py#L925

ntarocco (commented):
Occurrences I still see in the code and should be removed, but to be checked with care:
* `ui/src//api/tags` JS
* `utils.js` -> `recordToPidType`
* found a `tag%3A${tag}`, but I thing it should `tags%3A${tag}`... to be checked
* `https://127.0.0.1:5000/schemas/tags/tag-v1.0.0.json`
* `serializeTag`
* `Create new tag`
* `series` JS code, but it is not in the schema nor in the mappings. I guess, to be removed from JS if not needed, to be added in the schema/mappings and tested if to be kept (probably the team can help to answer the question).

I would recommend to search for all `tag` occurrences and remove the leftovers.

topless:
- [x] `ui/src/api/tags` JS
- [x] `utils.js` -> `recordToPidType`
- ~~[ ] found a `tag%3A${tag}`, but I thing it should `tags%3A${tag}`... to be checked~~ 
in RECORDS_REST_FACTS `tag=dict(terms=dict(field="tags", size=FACET_TAG_LIMIT))` so it has to remain `tag`
- [x] `https://127.0.0.1:5000/schemas/tags/tag-v1.0.0.json`
- [x] `serializeTag`
- [x] `Create new tag`
- [x] `series` JS code, but it is not in the schema nor in the mappings. I guess, to be removed from JS if not needed, to be added in the schema/mappings and tested if to be kept (probably the team can help to answer the question).

> Along with `tags` in the cli I have also added languages to be loaded from the corresponding vocabulary as per @equadon request.

ntarocco (approved):

ntarocco:
@topless 👏 it now looks complete! :) 👍

PR #638, opened 2019-12-04
oauth integration

zzacharo:

PR #639, opened 2019-12-04
File permissions

### LINKED ISSUE #628: Files: permissions
For e-times, implement the Files-REST permission factory to take into account the field `open_access`:
  1. create a new action to download files
  1. when trying to access bucket/file, fetch the record, fetch the field `open_access`
  2. is the field is `false`, then check if the user has the specific `Need/Action`

equadon:
* Permissions for files
* Removed language multi-field mapping for series
* UI fixes from comments during demo

Closes #628 

TODOs:
1. The test to download a file with an anonymous user now has a `sleep` otherwise it fails
2. Sometimes after uploading a file to an E-Item, the document is re-indexed but not with the last file uploaded

### DIFF
--- a/invenio_app_ils/permissions.py
+++ b/invenio_app_ils/permissions.py
@@ -9,14 +9,36 @@
 
 from __future__ import absolute_import, print_function
 
-from flask import abort
+from functools import wraps
+
+from flask import abort, current_app
 from flask_login import current_user
 from flask_principal import UserNeed
 from invenio_access import action_factory
 from invenio_access.permissions import Permission, authenticated_user
 from invenio_records_rest.utils import allow_all, deny_all
 
+from invenio_app_ils.proxies import current_app_ils
+
 backoffice_access_action = action_factory("ils-backoffice-access")
+
+
+def need_permissions(action):
+    """View decorator to check permissions for the given action or abort.
+
+    :param action: The action needed.
+    """
+    def decorator_builder(f):
+        @wraps(f)
+        def decorate(*args, **kwargs):
+            check_permission(
+                current_app.config["ILS_VIEWS_PERMISSIONS_FACTORY"](action)
+            )
+            return f(*args, **kwargs)
+
+        return decorate
+
+    return decorator_builder
 
 
 def check_permission(permission):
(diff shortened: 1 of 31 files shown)


invenio_app_ils/permissions.py
```diff
@@ -35,6 +37,22 @@ def backoffice_permission(*args, **kwargs):
     return Permission(backoffice_access_action)
 
 
+def file_download_permission(obj):
+    """File download permissions."""
+    search = current_app_ils.eitem_search
+    for hit in search.search_by_bucket_id(str(obj.bucket_id)).scan():
+        if not hit.to_dict().get("open_access", False):
```
zzacharo: @equadon This should be checked against ES6, no?

ui/src/pages/backoffice/EItem/EItemDetails/components/EItemFiles/EItemFiles.js
```diff
@@ -0,0 +1,142 @@
...
+    this.maxFiles = invenioConfig.eitems.maxFiles;
+  }
+
+  onDelete = row => {
+    this.props.deleteFile(row.bucket, row.key);
+  };
+
+  onSelectFile = file => {
+    if (this.props.files.length >= this.maxFiles) {
+      this.props.addNotification(
```
zzacharo: @equadon nit: I am doing also something similar in the [oauth PR](https://github.com/inveniosoftware/invenio-app-ils/pull/638/files#diff-df7011847aeaa5e03ddbfc1cfa49b48cR12) if you like to adopt :)

ui/src/pages/backoffice/components/buttons/DownloadButton/DownloadButton.js
```diff
@@ -0,0 +1,36 @@
...
+    return (
+      <Button
+        icon
+        primary
+        as="a"
+        size="small"
+        labelPosition="left"
+        href={to}
+        {...(disabled ? { disabled: true } : {})}
+        {...(fluid ? { fluid: true } : {})}
```
zzacharo: Why not just `<Button fluid={fluid ? true : false}`? Do I miss something?

ui/src/pages/backoffice/components/buttons/DownloadButton/DownloadButton.js
```diff
@@ -0,0 +1,36 @@
...
+
+DownloadButton.propTypes = {
+  text: PropTypes.string,
+  fluid: PropTypes.bool,
+  disabled: PropTypes.bool,
+};
+
+DownloadButton.defaultProps = {
+  text: 'Download',
+  fluid: false,
```
zzacharo: Actually you don't even need the check as you have already a default prop for the `fluid` property ;)

invenio_app_ils/permissions.py
equadon: The hit object didn't change much AFAIK from 6 to 7 so it should be okay but we'll see when the ES6 tests are running.

ui/src/pages/backoffice/EItem/EItemDetails/components/EItemFiles/EItemFiles.js
equadon: You approach is cleaner, I'll use that too. :)

invenio_app_ils/permissions.py
equadon: I tested it both in ES6 and ES7 and `to_dict()` is supported but I'm fine with using the DB approach instead.

equadon: I just remembered that we don't use Records-Files and only rely on Files-REST so we can't do this.

ntarocco: True! Let's change it as discussed together :)

ntarocco (approved):

PR #644, opened 2019-12-08
ill: data model and refactoring indexer

ntarocco:
* add `ill` module: it is not completely self-contained because it uses some utilities from ILS, but it has his own app and config.
* refactoring of indexer to be more readable and usable
* started the refactoring of the project by splitting in modules per record types. The complete refactoring will come in the future


invenio_app_ils/ill/config.py
```diff
@@ -0,0 +1,97 @@
...
+        ),
+        default_media_type="application/json",
+        max_result_window=10000,
+        error_handlers=dict(),
+        read_permission_factory_imp=backoffice_permission,
+        create_permission_factory_imp=backoffice_permission,
+        update_permission_factory_imp=backoffice_permission,
+        delete_permission_factory_imp=backoffice_permission,
+    ),
+    illlid=dict(
```
zzacharo, line 63: is this intented? the 3 `l`s? :P

zzacharo (approved):

invenio_app_ils/ill/search.py
```diff
@@ -0,0 +1,45 @@
...
+        doc_types = None
+
+
+class BorrowingRequestsSearch(RecordsSearch):
+    """Search for ILL borrowing requests."""
+
+    class Meta:
+        """Search only on borrowing requests index."""
+
+        index = "ill-borrowing-requests"
```
equadon: I believe we use underscores to separate words in other schemas but it should be fine either way. The only thing we'll need to do is convert the `RECORDS_REST_SORT_OPTIONS` config to be instantiated with `{}` instead of `dict()` since we can't use dashes in the keys there.

equadon (approved):

invenio_app_ils/circulation/indexer.py
```diff
@@ -0,0 +1,51 @@
...
+    indexer.index(indexed, referenced)
+
+
+class LoanIndexer(RecordIndexer):
+    """Indexer class for Loan record."""
+
+    def index(self, loan, arguments=None, **kwargs):
+        """Index an Loan."""
+        super().index(loan)
+        eta = datetime.utcnow() + current_app.config["ILS_INDEXER_TASK_DELAY"]
```
kpsherva, line 50: nit: are we suppose to use arrow instead of datetime?

invenio_app_ils/ill/config.py
kpsherva: iLL Library id :)

invenio_app_ils/search/api.py
```diff
@@ -32,32 +32,6 @@ def search_by_pid(self, *pids):
         """Retrieve documents with the given pid(s)."""
         return self.filter("terms", pid=pids)
 
-    def search_by_series_pid(self, series_pid=None):
```
kpsherva, line 35: we will not search by series pid?

kpsherva (approved):

invenio_app_ils/ill/schemas/ill-borrowing-requests/borrowing-request-v1.0.0.json
```diff
@@ -0,0 +1,37 @@
...
+    },
+    "pid": {
+      "title": "The persistent identifier of ILL borrowing request",
+      "type": "string"
+    },
+    "status": {
+      "title": "Status of the ILL borrowing request",
+      "type": "string"
+    },
+    "cancel_reason": {
```
zzacharo: Is this the same as the `reject_reason` in document-requests?

invenio_app_ils/ill/schemas/ill-borrowing-requests/borrowing-request-v1.0.0.json
```diff
@@ -0,0 +1,37 @@
...
+      "title": "Reason for cancelling the ILL borrowing request",
+      "type": "string"
+    },
+    "library_pid": {
```
zzacharo: Should we have some info about the library as well? I am thinking in the details page of the Ill request maybe the library name is needed or a link would be enough?

invenio_app_ils/ill/schemas/ill-borrowing-requests/borrowing-request-v1.0.0.json
ntarocco: Yes!

invenio_app_ils/ill/schemas/ill-borrowing-requests/borrowing-request-v1.0.0.json
ntarocco: There is already the resolver for that. Is this what you mean?

invenio_app_ils/circulation/indexer.py
ntarocco: good question! :) No I don't think so, this is just to delay the execution of the celery task, it is like `timedelta(2 secs)`

invenio_app_ils/search/api.py
ntarocco: It is not needed any more normally. This was a leftover from a resolver, that was resolving series from docs. But now they are handled with relations, so dead code

invenio_app_ils/ill/schemas/ill-borrowing-requests/borrowing-request-v1.0.0.json
ntarocco: Added to the schema!

PR #648, opened 2019-12-09
acquisition: vendor create and update

### LINKED ISSUE #646: acquisition: create update/create forms for vendors
### Vendors

[Vendor schema](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/acquisition/schemas/acq-vendors/vendor-v1.0.0.json)

- [x] Form to create a new vendor
- [x] Form  to edit a vendor

### Notes

- [Core](https://github.com/inveniosoftware/invenio-app-ils/tree/master/ui/src/forms/core) components needed for the forms (e.g string fields -> `StringField` or `TextField`)
- [Example](https://github.com/inveniosoftware/invenio-app-ils/tree/master/ui/src/pages/backoffice/Document/DocumentForms) structure that could be followed.

topless:
- form for vendor create
- form for vendor edit 
- upgrade formik to `2.0.6`
- closes #646

### DIFF
--- a/ui/src/pages/backoffice/Acquisition/Vendor/VendorEditor/components/VendorForm/VendorForm.js
+++ b/ui/src/pages/backoffice/Acquisition/Vendor/VendorEditor/components/VendorForm/VendorForm.js
@@ -1,0 +1,58 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import pick from 'lodash/pick';
+import { getIn } from 'formik';
+import { StringField, TextField } from '@forms';
+import { vendor as vendorApi } from '@api/acquisition/vendor';
+import { AcquisitionRoutes } from '@routes/urls';
+import { goTo } from '@history';
+import { BaseForm } from '@forms';
+
+export class VendorForm extends Component {
+  prepareData = data => {
+    return pick(data, ['name', 'address', 'email', 'phone', 'notes']);
+  };
+
+  createVendor = data => {
+    return vendorApi.create(data);
+  };
+
+  updateVendor = (pid, data) => {
+    return vendorApi.update(pid, data);
+  };
+
+  successCallback = response => {
+    goTo(
+      AcquisitionRoutes.vendorDetailsFor(getIn(response, 'data.metadata.pid'))
+    );
+  };
+
+  render() {
+    return (
+      <BaseForm
+        initialValues={
+          this.props.data ? this.prepareData(this.props.data.metadata) : {}
+        }
+        editApiMethod={this.updateVendor}
+        createApiMethod={this.createVendor}
+        successCallback={this.successCallback}
+        successSubmitMessage={this.props.successSubmitMessage}
+        title={this.props.title}
+        pid={this.props.pid ? this.props.pid : undefined}
+      >
+        <StringField label="Name" fieldPath="name" required />
+        <StringField label="Address" fieldPath="address" />
+        <StringField label="Email" fieldPath="email" />
+        <StringField label="Phone" fieldPath="phone" />
+        <TextField label="Notes" fieldPath="notes" rows={5} />
+      </BaseForm>
+    );
+  }
+}
+
+VendorForm.propTypes = {
+  data: PropTypes.object,
+  successSubmitMessage: PropTypes.string,
+  title: PropTypes.string,
+  pid: PropTypes.string,
+};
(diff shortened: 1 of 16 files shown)


ui/src/pages/backoffice/Acquisition/Vendor/VendorEditor/components/VendorForm/VendorForm.js
```diff
@@ -0,0 +1,53 @@
...
+
+  create = data => {
+    return vendorApi.create(data);
+  };
+
+  update = (pid, data) => {
+    return vendorApi.update(pid, data);
+  };
+
+  successCallback = () => goTo(AcquisitionRoutes.vendorList);
```
equadon: It should go to the details page of the newly created/updated vendor, right?

ui/src/pages/backoffice/Acquisition/Vendor/VendorEditor/components/VendorForm/VendorForm.js
```diff
@@ -0,0 +1,53 @@
...
+import pick from 'lodash/pick';
+import { StringField, TextField } from '@forms';
+import { vendor as vendorApi } from '@api/acquisition/vendor';
+import { AcquisitionRoutes } from '@routes/urls';
+import { goTo } from '@history';
+import { BaseForm } from '@forms';
+
+export class VendorForm extends Component {
+  prepareData = data => {
+    return pick(data, ['name', 'address', 'email', 'phone', 'notes']);
```
zzacharo, line 13: I would say that the VendorForm just renders the data it receives and any modifications to happen `VendorEditor`. But why do we need to pick only these fields in the first place?If it's about to restrict some fields from being submitted, until now we were using the prop `submitSerializer` which was stripping some fields before submitting...

ui/src/pages/backoffice/Acquisition/Vendor/VendorEditor/components/VendorForm/VendorForm.js
```diff
@@ -0,0 +1,53 @@
...
+  };
+
+  create = data => {
```
zzacharo: Rename potentially to `createVendor`, just to keep the name convention as in the other forms?

ui/src/pages/backoffice/Acquisition/Vendor/VendorEditor/components/VendorForm/VendorForm.js
```diff
@@ -0,0 +1,53 @@
...
+    return vendorApi.create(data);
+  };
+
+  update = (pid, data) => {
```
zzacharo: Same here -> `updateVendor`.

ui/src/pages/backoffice/Acquisition/Vendor/VendorEditor/components/VendorForm/VendorForm.js
topless: I noticed in some places is explicit and some others not, and my thinking was I am inside the vendor form so obviously will create and update vendor, by having a look around I see that pretty in all the `*Form.js` files these methods are repeated and are the same. Maybe it would make sense to move the to BaseForm.js and provide the relevant api module as a prop.

zzacharo: If there inconsistencies we could normalise them as well, but honestly I wouldn't move them from here as they are just simple functions and I am not sure it's worth the extra boilerplate for this...Even if "software engineering" speaking is tempting :)

topless: I agree, it's more straightforward/clear to see where it comes from rather than start looking for who supplied the prop and what prop. I like the shorter names though .. :)

ui/src/pages/backoffice/Acquisition/Vendor/VendorEditor/components/VendorForm/VendorForm.js
topless: Not everywhere, I checked the wrong example probably `LocationForm.js` (so does `ItemForm`, `EItemForm` and maybe more) is setup in the same fashion and happens to have the same fields 🎉

PR #649, opened 2019-12-09
Acquisition: Refactor backend

equadon:
* Implemented delete for vendors
* Added simple loaders
* Refactored module

Requires #644


invenio_app_ils/acquisition/jsonresolvers/order_vendor.py
```diff
@@ -0,0 +1,49 @@
...
+        vendor_pid = get_field_value(Order, order_pid, "vendor_pid")
+        vendor = Vendor.get_record_by_pid(vendor_pid)
+        return pick(
+            vendor,
+            "pid",
+            "name",
+            "address",
+            "email",
+            "phone",
+            "notes"
```
ntarocco, line 40: 👍

invenio_app_ils/acquisition/mappings/v6/acq_orders/order-v1.0.0.json
```diff
@@ -163,6 +163,34 @@
...
+        },
+        "vendor": {
+          "type": "object",
+          "properties": {
+            "pid": {
+              "type": "keyword"
+            },
+            "name": {
+              "type": "text",
+              "fields": {
```
ntarocco, line 175: do we need the `name` field as keyword?

invenio_app_ils/cli.py
```diff
@@ -607,6 +613,130 @@ def persist(self):
         return recs
 
 
+class VendorGenerator(Generator):
```
ntarocco, line 616: This `cli.py` is becoming massive too, we might need to break it in blocks eventually... :)

invenio_app_ils/config.py
```diff
@@ -1009,6 +939,58 @@ def _(x):
...
+            default_order="desc",
+            order=3,
+        ),
+        expected_delivery_date=dict(
+            fields=["expected_delivery_date"],
+            title="Expected delivery date",
+            default_order="desc",
+            order=4,
+        ),
+        # order_line_count=dict(
```
ntarocco, line 967: remove?

invenio_app_ils/config.py
```diff
@@ -1009,6 +939,58 @@ def _(x):
             order=2,
         ),
     ),
+    acq_orders=dict(  # OrderSearch.Meta.index
```
ntarocco, line 942: Do you think we should move inside the module `config.py`?

invenio_app_ils/jsonresolver/api.py
```diff
@@ -31,3 +31,8 @@ def _inner(*args, **kwargs):
                 return default_value
         return _inner
     return decorator
+
+
+def pick(obj, *keys):
```
ntarocco, line 36: 👍

ntarocco (commented):
Needs rebase, FYI I squashed my 2 commits into 1.

invenio_app_ils/acquisition/mappings/v6/acq_orders/order-v1.0.0.json
equadon: Yes, we need it for the vendor facet on the order search page.

invenio_app_ils/cli.py
equadon: Completely agree. `cli.py` should only have the commands imo and extract the functionality into api. Would be a good candidate for the refactor that was planned. :)

invenio_app_ils/config.py
equadon: The reasoning behind putting it in ILS `config.py` was that these sorting options are based on our needs so if it were to become its own module it would make more sense to have it in ILS. However, I don't see any harm in providing these sort options in acquisition `config.py` as an example for others. Do you have any preference?

ntarocco (approved):

PR #651, opened 2019-12-09
Acquisition: Backoffice search pages

equadon:
Added search pages for acquisition vendors and orders.

Requires #649 

Vendor search:
acq-vendor-search-new1

Order search (updated):
acq-order-search-updated

### DIFF
--- a/ui/src/api/utils.js
+++ b/ui/src/api/utils.js
@@ -1,3 +1,5 @@
+import { invenioConfig } from "@config";
+
 export const prepareDateQuery = (field, date, dateFrom, dateTo) => {
   if (
     (!date || typeof date !== 'string') &&
@@ -79,4 +81,13 @@ export const siblingRelationPayload = (relationType, extra, second) => {
     relation_type: relationType,
     ...extra,
   };
+};
+
+export const formatPrice = price => {
+  if (!price) return null;
+
+  return new Intl.NumberFormat(invenioConfig.i18n.priceLocale, {
+    style: 'currency',
+    currency: price.currency,
+  }).format(price.value);
 };
--- a/ui/src/config/__mocks__/invenioConfig.js
+++ b/ui/src/config/__mocks__/invenioConfig.js
@@ -10,6 +10,22 @@ export const invenioConfig = {
     deliveryMethods: { DELIVERY: '', 'PICK UP': '' },
     requestDuration: 60,
   },
+  orders: {
+    currencies: [
+      { value: 'EUR', text: 'Euro' },
+      { value: 'CHF', text: 'Swiss Francs' },
+    ],
+    paymentModes: [
+      { value: 'CREDIT_CARD', text: 'Credit Card' },
+      { value: 'CASH', text: 'Cash' },
+    ],
+    statuses: [
+      { value: 'CANCELLED', text: 'Cancelled' },
+      { value: 'RECEIVED', text: 'Received' },
+      { value: 'ORDERED', text: 'Ordered' },
+      { value: 'PENDING', text: 'Pending' },
+    ],
+  },
   relationTypes: [
     {
       id: 0,
--- a/ui/src/config/invenioConfig.js
+++ b/ui/src/config/invenioConfig.js
@@ -27,6 +27,9 @@ export const invenioConfig = {
   eitems: {
     maxFiles: 5,
   },
+  i18n: {
+    priceLocale: 'en-GB',
+  },
   items: {
     canCirculateStates: ['CAN_CIRCULATE'],
     circulationRestrictions: [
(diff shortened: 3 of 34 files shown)


ui/src/api/orders/order.js
```diff
@@ -0,0 +1,113 @@
+import { http, apiConfig } from '../base';
```
ntarocco: Can I propose to have the folder structure:
`/api/acquisitions/orders.js`
? WDYT?

ui/src/api/orders/order.js
```diff
@@ -0,0 +1,113 @@
...
+};
+
+const update = async (pid, data) => {
+  const response = await http.put(`${orderURL}${pid}`, data);
+  response.data = serializer.fromJSON(response.data);
+  return response;
+};
+
+const del = async pid => {
+  return await http.delete(`${orderURL}${pid}`);
```
ntarocco: Do we need the delete in the end?

ui/src/api/utils.js
```diff
@@ -80,3 +82,27 @@ export const siblingRelationPayload = (relationType, extra, second) => {
     ...extra,
   };
 };
+
+export const formatPrice = price => {
+  if (!price) return null;
+
+  const value = price.value.toFixed(2);
+
+  switch (price.currency) {
```
ntarocco: Maybe evaluate https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat instead?

ui/src/api/vendors/vendor.js
```diff
@@ -0,0 +1,104 @@
+import { http, apiConfig } from '../base';
```
ntarocco: we did renaming proposed above, you will have `/api/acquisitions/vendor.js`

ui/src/config/__mocks__/invenioConfig.js
```diff
@@ -10,6 +10,23 @@ export const invenioConfig = {
...
   },
+  orders: {
+    currencies: [
+      { value: 'EUR', text: 'Euro' },
+      { value: 'CHF', text: 'Swiss Francs' },
+    ],
+    paymentModes: [
+      { value: 'CREDIT_CARD', text: 'Credit Card' },
+      { value: 'CASH', text: 'Cash' },
+      { value: 'BUDGET_CODE', text: 'Budget Code' },
```
ntarocco: Can you please remove the budget code from ILS?

ui/src/config/invenioConfig.js
```diff
@@ -51,15 +51,33 @@ export const invenioConfig = {
...
+  orders: {
+    currencies: [
+      { value: 'EUR', text: 'Euro' },
+      { value: 'CHF', text: 'Swiss Francs' },
+    ],
+    maxShowOrderLines: 3,
+    paymentModes: [
+      { value: 'CREDIT_CARD', text: 'Credit Card' },
+      { value: 'CASH', text: 'Cash' },
+      { value: 'BUDGET_CODE', text: 'Budget Code' },
```
ntarocco: same here

ui/src/config/searchConfig.js
```diff
@@ -261,6 +261,102 @@ const searchConfig = {
...
+          order: 3,
+          title: 'Delivery date',
+        },
+        {
+          default_order: 'desc',
+          field: 'expected_delivery_date',
+          order: 4,
+          title: 'Expected delivery date',
+        },
+        // {
```
ntarocco: remove if removed in the backend?

ui/src/api/orders/order.js
equadon: The only issue is where do we put the serializer and tests? Perhaps `/api/acquisitions/orders/order.js` to follow the same structure as the other apis?

ui/src/api/vendors/vendor.js
equadon: Renamed them as proposed.

ui/src/api/utils.js
equadon: It seems to be a better solution than rolling our own. All the major browsers have had support for it since at least 2016 so for me it looks good.

ntarocco:
Thanks for the screenshots, a few comments (ping @kprzerwa):
**VENDORS**:
* isn't too "empty/white" in full-width? Probably because there are no facets...

**ACQUISITIONS**:
* Nice! 👏 
* why the `paper` part in `Document (paper)`? Is it the medium?
* I am bit confused by the double numbering in the list of order lines `1 Orderline #1...`, I would go for something easier... e.g.:
  * `2 x Document (paper) for patron 1 - 20.00 CHF`
  * `2 x "The lord of the r..." for patron 1 - 20.00 CHF` (if you have the title)
  basically `quantity x title for patron - total price`. The orderline number might be not really relevant
* I would remove the currency facet because I am not sure when I select a currency in which field I am fitering: grand total? In any case, not sure it is super useful...
* `delivery_date` is it the date when delivered? if yes, I would change to `Delivered the` or `Delivered: `

ntarocco:
BTW, in general, the `Show x results per page` looks globally broken...

equadon:
@ntarocco I agree the vendor search is very empty. We couldn't come up with any useful facets but we could have a ghost facet column just to reduce the space to make it more compact.

Yes, *paper* is the medium. Unfortunately, we don't have the titles, only the document PID so it's hard to print the order lines in a meaningful way.

I like having the quantity in there though instead of the numbers since they don't mean anything.

ui/src/config/__mocks__/invenioConfig.js
```diff
@@ -10,6 +10,22 @@ export const invenioConfig = {
     deliveryMethods: { DELIVERY: '', 'PICK UP': '' },
     requestDuration: 60,
   },
+  orders: {
+    currencies: [
+      { value: 'EUR', text: 'Euro' },
```
ntarocco, line 15: to remove for next PR?

ntarocco (approved):

PR #656, opened 2019-12-13
Backoffice: eitem search page

kpsherva:
Screenshot 2019-12-13 at 14 32 11
Screenshot 2019-12-13 at 14 32 19

### DIFF
--- a/ui/src/components/SearchControls/components/SearchAggregations/SearchAggregationsCards.js
+++ b/ui/src/components/SearchControls/components/SearchAggregations/SearchAggregationsCards.js
@@ -14,30 +14,31 @@ import { getSearchConfig } from '../../../../config';
 export default class SearchAggregationsCards extends Component {
   searchConfig = getSearchConfig(this.props.modelName);
 
-  _renderValueElement = (
-    bucket,
-    isSelected,
-    onFilterClicked,
-    getChildAggCmps
-  ) => {
-    const childAggCmps = getChildAggCmps(bucket);
-    return (
-      <List.Item key={bucket.key}>
-        <List.Content floated="right">
-          <Label>{bucket.doc_count}</Label>
-        </List.Content>
-        <List.Content>
-          <Checkbox
-            label={`${bucket.key}`}
-            value={bucket.key}
-            onClick={() => onFilterClicked(bucket.key)}
-            checked={isSelected}
-          />
-          {childAggCmps}
-        </List.Content>
-      </List.Item>
-    );
-  };
+   _renderValueElement = (
+      bucket,
+      isSelected,
+      onFilterClicked,
+      getChildAggCmps
+    ) => {
+      const childAggCmps = getChildAggCmps(bucket);
+      const key = bucket.key_as_string ? bucket.key_as_string : bucket.key;
+      return (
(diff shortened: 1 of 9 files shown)


ui/src/components/SearchControls/components/SearchAggregations/SearchAggregationsCards.js
```diff
@@ -21,17 +21,18 @@ export default class SearchAggregationsCards extends Component {
       getChildAggCmps
     ) => {
       const childAggCmps = getChildAggCmps(bucket);
+      const key = bucket.key_as_string ? bucket.key_as_string : bucket.key;
```
ntarocco, line 24: What about changing the values `true/false` to `Open Access/Closed Access`?

ui/src/pages/backoffice/EItem/EItemSearch/EitemList.js
```diff
@@ -0,0 +1,125 @@
...
+                        {eitem.metadata.files.length} file(s) attached{' '}
+                        <Icon name="file" />
+                      </List.Content>
+                    )}
+                  </List.Item>
+                  <List.Item>
+                    <List.Content>
+                      {!isEmpty(eitem.metadata.urls) && (
+                        <List.Content>
+                          {eitem.metadata.urls.length} external providers <Icon name="external" />
```
ntarocco: External provider might be valid for CERN, but for others might be that e-books are served inside the organization.
I would just put something like `Link to electronic item` or similar

ui/src/pages/backoffice/EItem/EItemSearch/EitemList.js
```diff
@@ -0,0 +1,125 @@
...
+                    </List.Content>
+                  </List.Item>
+                </List>
+              </Item.Meta>
+            </Grid.Column>
+            <Grid.Column computer={4} largeScreen={4}>
+              <List>
+                <List.Item>
+                  <List.Content>
+                    {eitem.metadata.open_access && (
```
ntarocco, line 67: Shall we display a `closed access` label in case it is not open?

kpsherva: the question is which one will be less likely to happen, so the label does not appear everywhere, I assumed the open will be less frequent, do you think it might be an opposite case?

ui/src/pages/backoffice/EItem/EItemSearch/EitemList.js
kpsherva: true, for me `external` meant external from the ILS system. but it might be understood also as external for the organisation

ui/src/components/SearchControls/components/SearchAggregations/SearchAggregationsCards.js
kpsherva: do you think it might have any significant performance influence for this case if we index string/keyword instead of boolean?

equadon: You could just add a label for the UI for true/false, right?

ntarocco (approved):

PR #664, opened 2019-12-17
acquisition: vendor details page

equadon:
Re-index orders if a document in an order line was updated.

Vendor details page:
acq-vendor-details-v3

Updated order details page:
acq-order-details-v3

### DIFF
--- a/invenio_app_ils/acquisition/ext.py
+++ b/invenio_app_ils/acquisition/ext.py
@@ -30,16 +30,19 @@ class _InvenioIlsAcquisitionState(object):
         self.app = app
 
     def record_class_by_pid_type(self, pid_type):
+        """Return a record class by pid type."""
         endpoints = self.app.config.get('RECORDS_REST_ENDPOINTS', [])
         return endpoints[pid_type]['record_class']
 
-    def search_by_pid_type(self, pid_type):
+    def search_class_by_pid_type(self, pid_type):
+        """Return a search class by pid type."""
         endpoints = self.app.config.get("RECORDS_REST_ENDPOINTS", [])
         return endpoints[pid_type]["search_class"]
 
     def indexer_by_pid_type(self, pid_type):
+        """Return an indexer instance by pid type."""
         endpoints = self.app.config.get("RECORDS_REST_ENDPOINTS", [])
-        return endpoints[pid_type]["indexer_class"]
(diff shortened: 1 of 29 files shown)


invenio_app_ils/acquisition/ext.py
```diff
@@ -33,11 +33,11 @@ def record_class_by_pid_type(self, pid_type):
         endpoints = self.app.config.get('RECORDS_REST_ENDPOINTS', [])
         return endpoints[pid_type]['record_class']
 
-    def search_by_pid_type(self, pid_type):
+    def search_class_by_pid_type(self, pid_type):
         endpoints = self.app.config.get("RECORDS_REST_ENDPOINTS", [])
         return endpoints[pid_type]["search_class"]
 
-    def indexer_by_pid_type(self, pid_type):
+    def indexer_class_by_pid_type(self, pid_type):
```
ntarocco: just make sure that you will always instantiate the cls when using it: in the main `ext.py`, we return an instance and not a class if I am not mistaken. Check the usage of it.

ui/src/pages/backoffice/Acquisition/Order/OrderDetails/PaymentInformation.js
```diff
@@ -7,18 +7,24 @@ export class PaymentInformation extends React.Component {
...
-    if (!payment) return null;
     const leftTable = [
       {
         key: `Total (${order.grand_total_main_currency.currency})`,
         value: formatPrice(order.grand_total_main_currency),
       },
-      {
+    ];
+    if (
+      order.grand_total.currency !== order.grand_total_main_currency.currency
```
ntarocco: It might be that librarian insert/misuse the totals and insert both with the same currency. Does it make sense to change this if and check if `grand_total` is defined instead?

ntarocco (approved):
Good job! I would just add some space between the red and blue buttons on the right navigation panel.

kpsherva (approved):

ui/src/pages/backoffice/Acquisition/Order/OrderDetails/PaymentInformation.js
equadon: I kept both and indicated that the "main" one is the local currency with a popup explanation.

PR #677, opened 2019-12-19
backoffice: cancel document request

### LINKED ISSUE #667: backoffice: Cancel document request
The document request details page in the backoffice should have a few actions/buttons:
* `Cancel request`: remove the cancel reason for DocumentRequest. The cancel should be of 3 types (3 buttons/a button with 3 options?)
  * `User cancelled`: the user cancelled for any reason. It can be triggered by the user in the Profile page, or by the librarian because the user asked her/him to do it.
  * `Librarian cancelled because document already in the catalog`: the librarian cancelled because the document is in the catalog but the user did not find it. The librarian will probably then create a loan out of it.
  * `Librarian cancelled because document not found`: the librarian cancelled because the document has not been found in any of the providers/external libraries.
  * An e-mail should be sent to the user in all cases

More info: https://github.com/inveniosoftware/invenio-app-ils/issues/634

equadon:
* Reject reason filter for document request search page
* ILS mail module
* Structured email logging

Closes #667 

details-reject

reject-modal

Show reject reason and document on doc req search page:
search-list


zzacharo (approved):

invenio_app_ils/mail/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+        message_id=data["id"],
+        task_id=request.get("id"),
+        task=request.get("task"),
+        exception=repr(exc),
+        data=data,
+    )
+    celery_logger.exception(
+        json.dumps(error, sort_keys=True),
+        exc_info=exc
+    )
```
ntarocco, line 78: silly question, were you able to test both cases?

ntarocco (approved):

invenio_app_ils/mail/tasks.py
equadon: Yes, both success and errors were working. We might have to tweak the data we log but with the unique message id we should be able to track emails if they fail/succeed or fails before the task is sent.

kpsherva (approved):

PR #679, opened 2019-12-20
forms: order improvement

topless:
- fixed order line patron
- fixed order line mediums
- little re-style and reorganize order form

Screenshot 2019-12-20 at 10 43 21

### DIFF
--- a/ui/src/pages/backoffice/Acquisition/Order/OrderEditor/components/OrderForm/OrderForm.js
+++ b/ui/src/pages/backoffice/Acquisition/Order/OrderEditor/components/OrderForm/OrderForm.js
@@ -1,22 +1,13 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { Header } from 'semantic-ui-react';
+import { Header, Segment, Grid } from 'semantic-ui-react';
 import { getIn } from 'formik';
-import { SelectorField, TextField } from '@forms';
 import { Loader } from '@components';
-import { order as orderApi, vendor as vendorApi } from '@api/acquisition';
+import { order as orderApi } from '@api/acquisition';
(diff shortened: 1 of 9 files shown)


ntarocco:
That's a nice starting for styled forms! Thanks!

ui/src/pages/backoffice/Acquisition/Order/OrderEditor/components/OrderForm/OrderForm.js
```diff
@@ -34,6 +25,11 @@ const orderSubmitSerializer = (values, newRecord) => {
         ? line.document.id
         : line.document.pid;
     }
+    if (line.patron) {
+      line.patron_pid = _has(line.patron, 'id')
+        ? line.patron.id
+        : line.patron.pid;
```
ntarocco, line 31: If sometimes we have `id` and sometimes `pid`, we might have to fix this... did you notice that we have both or it is just to be safe?

ntarocco (approved):

equadon (approved):

ui/src/pages/backoffice/Acquisition/Order/OrderEditor/components/OrderForm/OrderForm.js
topless: So when we use the `ESSelector` it uses this [serializer](https://github.com/inveniosoftware/invenio-app-ils/blob/master/ui/src/components/ESSelector/serializer.js) which uses `id` because not everything searchable is a record. The existing rendered data are not getting serialized (so they have pids) only the search results. We should address it in an application level. Thanks to @equadon for helping with the struggle on this one particularly.

kpsherva (approved):

PR #685, opened 2020-01-06
acq: orders sorting

### LINKED ISSUE #672: backoffice: purchase order search sort broken
http://localhost:3000/backoffice/acquisition/orders?q=&sort=received_date&order=desc&p=1&s=10 the sort is broken

Different sort options (asc/desc)

- [x] Order Date
- [x] Grand Total (in main currency)
- [x] ~Delivery Date~ -> Received Date
- [x] Expected Delivery Date
- [x] Best Match  (I am not sure this is needed)

topless:
- changed naming of delivery date to received
- changed results per page to match default Elastic results (10) so SearchResultsPerPage isn't broken in first render
- cleaned unused imports
- fixed sorting for grand total in main currency, also added currency indication to make it more clear
- display grand total on an alternative currency only if its available
- defaultResultSize inside Pagination and ResultTable components instead of explicit default in every instance
- defaultResultSize for QueryBuildHelper, which fixes the ResulstPerPage components
- closes #672

### DIFF
--- a/ui/src/config/searchConfig.js
+++ b/ui/src/config/searchConfig.js
@@ -353,15 +353,15 @@ const searchConfig = {
         },
         {
           default_order: 'desc',
-          field: 'grand_total_main_currency.value',
+          field: 'grand_total',
           order: 2,
-          title: 'Grand total',
+          title: `Total (${invenioConfig.orders.defaultCurrency})`,
         },
         {
           default_order: 'desc',
           field: 'received_date',
           order: 3,
-          title: 'Delivery date',
+          title: 'Received date',
         },
         {
           default_order: 'desc',
--- a/ui/src/pages/backoffice/components/OrderList/OrderListEntry.js
+++ b/ui/src/pages/backoffice/components/OrderList/OrderListEntry.js
@@ -31,7 +31,12 @@ export default class OrderListEntry extends Component {
           </Link>
         </Item.Description>
         <Item.Description>
-          <label>total</label> {formatPrice(order.metadata.grand_total)}
+          <label>total</label>
+          {formatPrice(order.metadata.grand_total_main_currency)}
+          {order.metadata.grand_total.currency !==
+          order.metadata.grand_total_main_currency.currency
+            ? `(${formatPrice(order.metadata.grand_total)})`
+            : ''}
         </Item.Description>
       </>
     );
@@ -106,7 +111,7 @@ export default class OrderListEntry extends Component {
             <List.Content floated="right">
               <strong>{toShortDateTime(received_date)}</strong>
             </List.Content>
-            <List.Content>delivered</List.Content>
+            <List.Content>received</List.Content>
           </List.Item>
         )}
         {expected_delivery_date && (
@@ -114,7 +119,7 @@ export default class OrderListEntry extends Component {
             <List.Content floated="right">
               <strong>{toShortDateTime(expected_delivery_date)}</strong>
             </List.Content>
-            <List.Content>expected delivery date</List.Content>
+            <List.Content>expected</List.Content>
           </List.Item>
         )}
       </List>
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -970,13 +970,13 @@ RECORDS_REST_SORT_OPTIONS = dict(
         ),
         grand_total=dict(
             fields=["grand_total_main_currency.value"],
-            title="Grand total",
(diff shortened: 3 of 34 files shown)


ntarocco (approved):

ui/src/config/searchConfig.js
```diff
@@ -4,16 +4,16 @@ import merge from 'lodash/merge';
...
-    text: '30',
-    value: 30,
+    text: '25',
+    value: 15,
   },
   {
-    text: '60',
-    value: 60,
+    text: '50',
+    value: 30,
```
equadon: The text and the value should match

equadon (approved):

ui/src/pages/backoffice/components/OrderList/OrderListEntry.js
```diff
@@ -30,8 +30,14 @@ export default class OrderListEntry extends Component {
...
           </Link>
         </Item.Description>
+        {order.metadata.grand_total.currency !== order.metadata.grand_total_main_currency.currency && (
+          <Item.Description>
+            <label>grand total</label> {formatPrice(order.metadata.grand_total)}
+          </Item.Description>
+        )}
         <Item.Description>
-          <label>total</label> {formatPrice(order.metadata.grand_total)}
+          <label>grand total</label>{' '}
```
zzacharo: `{' '}` is this prettier thing or is it intended?

ui/src/pages/backoffice/components/OrderList/OrderListEntry.js
```diff
@@ -30,8 +30,14 @@ export default class OrderListEntry extends Component {
             {order.metadata.vendor.name}
           </Link>
         </Item.Description>
+        {order.metadata.grand_total.currency !== order.metadata.grand_total_main_currency.currency && (
```
zzacharo: Just to understand, is this meant to be `if...else` conditional rendering or just render the below label in case the condition is satisfied? I am asking because the label for both `<Item.Description>` elements is `grand total`

ui/src/pages/backoffice/components/OrderList/OrderListEntry.js
topless: @zzacharo this is prettier, we can use `&nbsp;` if you prefer

ui/src/pages/backoffice/components/OrderList/OrderListEntry.js
topless: @zzacharo this is conditional display, show it only if the currencies are different. Indeed the label is the same, but the currencies differentiate them and looks like this
Screenshot 2020-01-07 13 30 09

ui/src/pages/backoffice/components/OrderList/OrderListEntry.js
zzacharo: No but I think is fine to remove it, no?

ui/src/pages/backoffice/components/OrderList/OrderListEntry.js
zzacharo: Isn't it better to just put the default currency in parenthesis in case it's different than the order currency? Like `70.91 USD (72 CHF)`?

ntarocco: Now that I see it, please do not display "grand" total to the user: this is more a technical term, I doubt you ever saw it in an e-commerce website...

topless: I think I would put them the other way around, because everybody will have default currency

`grand total 80 CHF (75 EUR)`

topless: @ntarocco I changed it in order to reflect the sort option from the dropdown, I will change the dropdown too, to display `Total` instead of `Grand Total`

ui/src/pages/backoffice/components/OrderList/OrderListEntry.js
topless: total139.41 CHF(126.74 €)

@zzacharo it will look like this then

zzacharo: @topless then you can use something like : \<label>\`total ${formatPrice(...)}\`\</label>. It should do the trick no?

topless:
Screenshot 2020-01-07 13 49 38

PR #686, opened 2020-01-07
documents module with simple minter/fetcher

ntarocco:
Move Document code in the document module, no other changes.

Refactoring before implementing https://github.com/inveniosoftware/invenio-app-ils/issues/687


invenio_app_ils/documents/api.py
```diff
@@ -0,0 +1,149 @@
...
+from invenio_app_ils.fetchers import pid_fetcher
+from invenio_app_ils.minters import pid_minter
+from invenio_app_ils.proxies import current_app_ils
+from invenio_app_ils.records.api import IlsRecordWithRelations
+
+DOCUMENT_PID_TYPE = "docid"
+DOCUMENT_PID_MINTER = "docid"
+DOCUMENT_PID_FETCHER = "docid"
+
+DocumentIdProvider = type(
```
ntarocco, line 27: For reviewers, this is the only real change: inline minters/fetcher instead of having multiple files (minter, fetcher, provider, pid). Do you think this is better/more readable?

kpsherva (approved):
I like the approach to group everything by record type.  I have only few questions:
- Do you plan also to handle schemas in the same way, to put doc schema into document module?
- Do we care that 
```
DOCUMENT_PID_TYPE = "docid"
DOCUMENT_PID_MINTER = "docid"
DOCUMENT_PID_FETCHER = "docid"
```
will not be in exposed the config anymore?

ntarocco:
> I like the approach to group everything by record type. I have only few questions:
> 
>     * Do you plan also to handle schemas in the same way, to put doc schema into document module?
> 
>     * Do we care that
> 
> 
> ```
> DOCUMENT_PID_TYPE = "docid"
> DOCUMENT_PID_MINTER = "docid"
> DOCUMENT_PID_FETCHER = "docid"
> ```
> 
> will not be in exposed the config anymore?

1. yes, it is done in this PR, see `invenio_app_ils/schemas/documents/document-v1.0.0.json → invenio_app_ils/documents/schemas/documents/document-v1.0.0.json` in the list of diff
2. It is actually exposed and used, see [here](https://github.com/inveniosoftware/invenio-app-ils/pull/686/files#diff-56c674e5fed0533f8941caf8761816a4R42), no change of behaviour

topless:
@ntarocco I like where we are going with this. Since we started grouping functionality per model, and as the list of directories grows it starts to feel like we are mixing different stuff. Would it make sense to organise them a level more, we could do something like, 

```python
modules/
  ill
  acquisition
  circulation

model/
  document_requests
  documents
  eitems
  internal_locations
  items
  locations
  patrons
  records_relations
  relations
  series

# ... and organize somehow the rest that makes sense, I am guessing once we split 
# them all there will be no need for pidstore, jsonresolver etc since each one is going 
# to be included in its relevant model.

records
files
jsonresolver
mail
mappings
pidstore
schemas
search
stats
templates
translations
vocabularies
ui
```

invenio_app_ils/documents/api.py
zzacharo: Personally I like this approach so I don't have to jump to different files to see these values :) My only concern is if in the future there is the need of adding some custom code on these providers then the `type` approach would not be in handy compared to a class definition. But this can easily be refactored whenever and if is needed :)

zzacharo (approved):

invenio_app_ils/documents/api.py
topless: It is more organised with less noise for sure 👍

topless (approved):

equadon (approved):
Looks really good to me! A lot cleaner and more organized. :+1:

ntarocco:
> @ntarocco I like where we are going with this. Since we started grouping functionality per model, and as the list of directories grows it starts to feel like we are mixing different stuff. Would it make sense to organise them a level more, we could do something like,
> 
> ```python
> modules/
>   ill
>   acquisition
>   circulation
> 
> model/
>   document_requests
>   documents
>   eitems
>   internal_locations
>   items
>   locations
>   patrons
>   records_relations
>   relations
>   series
> 
> # ... and organize somehow the rest that makes sense, I am guessing once we split 
> # them all there will be no need for pidstore, jsonresolver etc since each one is going 
> # to be included in its relevant model.
> 
> records
> files
> jsonresolver
> mail
> mappings
> pidstore
> schemas
> search
> stats
> templates
> translations
> vocabularies
> ui
> ```

As discussed I will take into account this in the next PRs.

PR #689, opened 2020-01-07
documents: add open_access field

### LINKED ISSUE #681: Items search: search suggestion by `Shelf`
The items search page should have a search suggestion: `Search by Shelf`

### LINKED ISSUE #687: Document: add field `open_access`
Librarian can change visibility of a Document and set open/close access.
1. Add a field to the Document schema for open/close access, similar to e-items.
2. Make sure that CRUD operations are restricted to librarians only if close access

ntarocco:
* documents can be restricted by `open_access` field (implicit permissions)
* `_access` (explicit permissions), if and when defined`, takes always precedence
* refactor records item/list permissions to filter out closed access
  documents for non-admins/librarians
* changed records restrictions to use _access only when configured. In
  this way, the field is not evaluated on each search query when
  disabled.
* closes #687
* closes #681

### DIFF
--- a/invenio_app_ils/search/permissions.py
+++ b/invenio_app_ils/search/permissions.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""ILS search permissions."""
+
+import re
+
+from elasticsearch_dsl import Q
+from flask import current_app, g, has_request_context, request
+from flask_login import current_user
+
+from invenio_app_ils.errors import SearchQueryError, UnauthorizedSearchError
+from invenio_app_ils.permissions import backoffice_permission
+
+
+def _get_user_provides():
+    """Extract the user's provides from g."""
+    provides = []
+    for need in g.identity.provides:
+        try:
+            provides.append(need.value.lower())
+        except AttributeError:
+            # Add the user ID (integer) to the list
+            provides.append(need.value)
+    return provides
+
+
+def search_filter_record_permissions():
+    """Filter list of results by `_access` and `open_access` fields."""
+    if not has_request_context() or backoffice_permission().allows(g.identity):
+        return Q()
+
+    # A record is public if `open_access` field True or missing
+    open_access_field_missing = ~Q("exists", field="open_access")
+    is_open_access = open_access_field_missing | Q("term", open_access=True)
+
+    combined_filter = is_open_access
+
+    if current_app.config.get("ILS_RECORDS_EXPLICIT_PERMISSIONS_ENABLED"):
+        # if `_access`, check `_access.read` against the user. It takes
+        # precedence over `open_access`.
+        # if not `_access`, check if open access as before.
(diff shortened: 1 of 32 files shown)


ntarocco:
Quick explanation: for `e-items`, when `open_access: False` then only logged in users can get the  file.
Here for documents, when `open_access: False`, it means that only librarians/admins can see the record.
IMHO, even if the meaning is different, it is not an issue: this implicit permissions means that, given the value of the filed, some restriction is applied, depending on some conditions.
If you believe we should change the field, I can do that.

ntarocco:
@kprzerwa I have added the form field and the facet.
To be added in your PR (after you rebase):
1. search result, display open/closed on each as for e-items
2. document detail: clearly show if open/close access

invenio_app_ils/search/permissions.py
```diff
@@ -0,0 +1,117 @@
...
+        if has_request_context() and not backoffice_permission().allows(
+            g.identity
+        ):
+            # patron can find only his records
+            if not query_string:
+                # force query to be patron_pid:<logged in user>
+                query_string = "patron_pid:{}".format(g.identity.id)
+            else:
+                # check for patron_pid query value
+                match = re.match(r"patron_pid:(?P<pid>\d)", query_string)
```
equadon:
```suggestion
                match = re.match(r"patron_pid:(?P<pid>\d+)", query_string)
```
Or else it would only match with the first digit of the patron ID.

zzacharo: Now I am thinking, is patron_pid alphanumeric? @ntarocco

kpsherva: it is not for now, it does not pass through invenio-pidstore, no?

tests/api/test_record_permissions.py
```diff
@@ -73,45 +74,6 @@ def login_and_test(user_id):
     login_and_test(users["admin"].id)
 
 
-@pytest.mark.parametrize(
```
kpsherva, line 76: I think we should leave those to check the explicit permissions anyway. it helps if we change something in invenio-access

kpsherva (approved):

tests/api/documents/test_document_permissions.py
```diff
@@ -0,0 +1,102 @@
...
+    indexer = RecordIndexer()
+    doc1 = Document.get_record_by_pid("docid-open-access")
+    doc2 = Document.get_record_by_pid("docid-closed-access")
+    for doc in [doc1, doc2]:
+        doc.update(dict(_access=dict(read=[users["patron2"].id])))
+        doc.commit()
+        db.session.commit()
+        indexer.index(doc)
+
+    test_data = [
```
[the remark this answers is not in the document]
ntarocco: I thought about that, but then tests are super slow because they will re-create db/es each time

invenio_app_ils/search/permissions.py
ntarocco: Well spotted! Actually I just moved the code from one place to another. I have applied the fix.
Yes user ids (int) at the moment.

tests/api/test_record_permissions.py
ntarocco: I removed them because to me they look exactly the same as the previous one, so a duplication. Can you confirm?

equadon (approved):

Issue #669, opened 2019-12-17
backoffice: attach/change document to document request


kpsherva:
* `Attach document`:
  * enabled only when 
    * state `PENDING`
    * a document **is not attached** to the request
  * when disabled, it should say that it is disabled because document already attached or request already fulfilled.
* `Change attached document`:
  * enabled only when 
    * state `PENDING`
    * a document **is attached** to the request
  * when disabled, it should say that it is disabled because document not yet attached or request  not already fulfilled
More info: https://github.com/inveniosoftware/invenio-app-ils/issues/634

PR #696, opened 2020-01-10
document for document-request

### LINKED ISSUE #668: Backoffice: create a document from doc request
* `Create a document from this request`: it opens the new document request already pre-filled
  * enabled only when 
    * state `PENDING`
    * a document **is not attached** to the request
  * when disabled, it should say that it is disabled because document already attached or request
  * it would be nice, after the creation of the document, that the DocumentRequest has the new document attached.
More info: https://github.com/inveniosoftware/invenio-app-ils/issues/634

topless:
- refactor: moved document requests outside of document
- create a document from document request, prefilled form
- backlink newly created document to the document-request
- fixed issue with values supplied for Document Form
- introduced DocumentControls component for creating/searching/replacing document in document request
- added endpoint to remove document and set status to PENDING
- resolve $ref when we update the record
- attaching document from search
- closes #668, #669

dr-final-low


ui/src/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
```diff
@@ -10,6 +11,31 @@ export class DocumentEditor extends Component {
...
   };
 
+  get documentRequest() {
+    const request = _get(this.props, 'location.state', null);
+    if (request) {
+      const data = {
+        documentRequestPid: request.pid,
+        metadata: {
+          // NOTE: Cheating so the authors don't break our document form
+          authors: [{ full_name: request.metadata.authors, type: 'PERSON' }],
```
equadon: Should we set the default type here? Also, since the author type is a vocabulary can we assume there is a vocabulary called `PERSON`?

ui/src/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
```diff
@@ -10,6 +11,31 @@ export class DocumentEditor extends Component {
...
+          title: request.metadata.title,
+        },
+      };
+      const journal_title = _get(request, 'metadata.journal_title', null);
```
equadon:
```suggestion
      const journalTitle = _get(request, 'metadata.journal_title', null);
```

ui/src/pages/backoffice/DocumentRequest/DocumentRequestDetails/components/DocumentRequestSteps/DocumentRequestSteps.js
```diff
@@ -0,0 +1,131 @@
...
+  render() {
+    const { document_pid: docPid, state } = this.props.data.metadata;
+    const hasDocument = docPid ? true : false;
+    // TODO: when we integrated ill_pid or acq_pid
+    const hasProvider = false;
+    return state !== 'REJECTED' ? (
+      <>
+        {this.documentInfo()}
+        {this.renderSteps(hasDocument, hasProvider)}
+        <Segment>{this.renderStepsContent(hasDocument, hasProvider)}</Segment>
```
equadon: Right now the calculation to figure out the current "step state" happens independently in both `renderSteps` and `renderStepsContent`. This might be a bit messy to keep these synchronized in case there are changes so maybe we could have that calculation happen once and store the "step state" and pass to these functions. What do you think?

ntarocco (commented):
In terms of UI, but check also with @kprzerwa, I would have put the Document maybe smaller in the Metadata box, but it is just IMHO.
I like the new UI much more! 👍 👏

invenio_app_ils/records/views.py
```diff
@@ -154,7 +164,31 @@ class AcceptRequestResource(ContentNegotiatedMethodView):
     @pass_record
     def post(self, pid, record, **kwargs):
         """Accept request post method."""
-        raise DocumentRequestError("Accept request is not implemented")
+        data = request.get_json() or dict()
```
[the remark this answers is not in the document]
topless: @ntarocco Thanks a lot I didn't manage to make it work yesterday with the loaders, this is probably what I was missing.

ui/src/pages/backoffice/DocumentRequest/DocumentRequestDetails/components/DocumentRequestSteps/DocumentRequestSteps.js
topless: I totally agree, there should be one source of truth instead of applying fragile logic everywhere. I had some issues rendering while also applying updates to the state.

ui/src/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
topless: This one just follows JavaScript naming convention

edit: lol which makes you right, I am blind

ui/src/pages/backoffice/DocumentRequest/DocumentRequestDetails/components/DocumentRequestSteps/DocumentRequestSteps.js
```diff
@@ -0,0 +1,160 @@
...
+
+  renderStepContent(step) {
+    const { data } = this.props;
+    switch (step) {
+      case STEPS.document:
+        return (
+          <DocumentAttachStep
+            data={data}
+            fetchDocumentRequestDetails={this.props.fetchDocumentRequestDetails}
+          />
```
kpsherva: I would suggest to have two(three split components instead of having conditional props. they are different enough I think with separate goals and I think that a component should do one thing at a time (like one below, DocumentAttachStep)

ui/src/pages/backoffice/DocumentRequest/DocumentRequestDetails/components/DocumentRequestSteps/components/DocumentProvidersStep.js
```diff
@@ -0,0 +1,110 @@
...
+        </Grid>
+        <Divider vertical>Or</Divider>
+      </Segment>
+    );
+  }
+
+  render() {
+    return (
+      <>
+        {this.renderAcqSegment()}
```
kpsherva: do you think wrapping them in methods is better for readability? for me more natural is to follow the lines instead of jumping to the method. @zzacharo @equadon what do you think? (if there is not a lot of code, when there is more it should probably be split for smaller components

ui/src/pages/backoffice/Document/DocumentDetails/DocumentDetails.js
```diff
@@ -27,19 +28,18 @@ import {
 class DocumentHeader extends Component {
   render() {
     const { data } = this.props;
-
     const recordInfo = (
       <>
-        <label>Document</label> #{data.metadata.pid}{' '}
+        <Label className="muted">Document</Label> {data.metadata.pid}{' '}
```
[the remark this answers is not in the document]
topless: We agreed to introduce `@mutedTextColor` for the less important captions. Disabled indicates that something  cannot be used.

ui/src/pages/backoffice/Document/DocumentForms/DocumentEditor/DocumentEditor.js
```diff
@@ -10,6 +11,31 @@ export class DocumentEditor extends Component {
...
+          // NOTE: Cheating so the authors don't break our document form
+          authors: [{ full_name: request.metadata.authors }],
+          title: request.metadata.title,
+        },
+      };
+      const journalTitle = _get(request, 'metadata.journal_title', null);
+      if (journalTitle) data.metadata.journal_title = journalTitle;
+
+      const edition = _get(request, 'metadata.edition', null);
+      if (edition) data.metadata.edition = edition;
```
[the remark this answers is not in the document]
topless: Great catch, apparently we did that because the forms throw errors when we supply `null` values. If we provide `undefined` though they work as expected.

ui/src/pages/backoffice/DocumentRequest/DocumentRequestDetails/components/DocumentRequestSteps/components/DocumentProvidersStep.js
```diff
@@ -0,0 +1,114 @@
...
+} from '@components/ESSelector/serializer';
+import { goTo } from '@history';
+
+class AcqProvider extends Component {
+  onCreateClick() {
+    return goTo(AcquisitionRoutes.orderCreate);
+  }
+
+  onSelectResult = async data => {
+    // TODO: Fire a request to update document request with provider info
```
equadon: Should we open an issue about this `TODO` too?

equadon (commented):
Great job! It looks really good. :)

invenio_app_ils/document_requests/loaders/jsonschemas/document_request_accept.py
```diff
@@ -0,0 +1,22 @@
...
+
+class DocumentRequestAcceptSchemaV1(RecordMetadataSchemaJSONV1):
+    """Document Request Accept schema."""
+
+    class Meta:
+        """Meta attributes for the schema."""
+
+        unknown = EXCLUDE
+
+    document_pid = fields.Str()
```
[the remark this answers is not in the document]
topless: Not every time, I was planning to re-use the same endpoint so we can update the `document_request` with ILL or Acq information when they become available.

invenio_app_ils/document_requests/views.py
```diff
@@ -0,0 +1,173 @@
...
+    blueprint.add_url_rule(
+        "{0}/reject".format(options["item_route"]),
+        view_func=reject_view,
+        methods=["POST"],
+    )
+
+    return blueprint
+
+
+class PendingRequestResource(ContentNegotiatedMethodView):
```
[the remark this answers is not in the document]
topless: I didn't use a loader because at least until now, it doesn't have any data, will add the loader if we need to pass extra information when Ill or Acq come in play.

ui/src/pages/backoffice/DocumentRequest/DocumentRequestDetails/components/DocumentInfo/DocumentInfo.js
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from 'react';
+import { Link } from 'react-router-dom';
+import { Button, Grid, Header, Message, Segment } from 'semantic-ui-react';
+import { BackOfficeRoutes } from '@routes/urls';
+import { documentRequest as documentRequestApi } from '@api';
+import _get from 'lodash/get';
+import { DocumentIcon } from '@pages/backoffice/components';
+
+export default class DocumentInfo extends Component {
```
kpsherva, line 9: I would just change a name of this component, because it is not info only, maybe `DocumentManager` or `DocumentOrganiser`

ui/src/pages/backoffice/DocumentRequest/DocumentRequestDetails/components/DocumentRequestActions/RejectAction.js
```diff
@@ -47,7 +47,7 @@ export default class RejectAction extends React.Component {
     }
   };
 
-  render() {
+  renderDropdown() {
```
kpsherva, line 50: I would put generating options in separate method, and pass them to dropdown as a prop (and the dropdown itself in the render function) - IMO it is better readable, but it might be in the next PR if you find it relevant

PR #700, opened 2020-01-14
Literature search endpoint

equadon:
* Literature search now searches both documents and series but excludes documents with document_type `PERIODICAL_ISSUE`
* Added flag `include_all` to also include periodical issues
* Backoffice series details page uses literature endpoint
* Change `document_type` to be an enum instead of a vocabulary
* Fix relations indexer issues
* Literature endpoint uses the same filtering as documents search
* Series `publishers` changed to `publisher`
* Converted `literature` to follow the same convention as documents
* Revert backoffice series literature change


invenio_app_ils/config.py
```diff
@@ -665,6 +669,37 @@ def _(x):
...
+        search_serializers_aliases={
+            "csv": "text/csv",
+            "json": "application/json",
+        },
+        item_route="/frontsite-search/<pid(fssid):pid_value>",
+        list_route="/frontsite-search/",
+        default_media_type="application/json",
+        max_result_window=_RECORDS_REST_MAX_RESULT_WINDOW,
+        error_handlers=dict(),
+        list_permission_factory_imp=allow_all,
```
ntarocco: You will still have to take into account closed access docs: https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/documents/search.py#L24

invenio_app_ils/config.py
```diff
@@ -665,6 +669,37 @@ def _(x):
...
+            "application/json": (
+                "invenio_app_ils.records.serializers:json_v1_search"
+            ),
+            "text/csv": ("invenio_app_ils.records.serializers:csv_v1_search"),
+        },
+        search_serializers_aliases={
+            "csv": "text/csv",
+            "json": "application/json",
+        },
+        item_route="/frontsite-search/<pid(fssid):pid_value>",
```
ntarocco: I would suggest to change `frontsite-search`: what about if we do `/records`?

invenio_app_ils/config.py
```diff
@@ -95,6 +95,9 @@
     EITEM_PID_FETCHER,
     EITEM_PID_MINTER,
     EITEM_PID_TYPE,
+    FRONT_SITE_SEARCH_PID_FETCHER,
```
ntarocco: What about `RECORDS` or `ALL_RECORDS` or similar? I am not sure we want to have it "tight" to the front site.
It is more for patrons in this case... `PATRONS_DOCS`?

invenio_app_ils/search/api.py
```diff
@@ -7,12 +7,18 @@
...
 from flask import current_app
-from invenio_search.api import RecordsSearch
+from invenio_search.api import DefaultFilter, RecordsSearch
 
 from invenio_app_ils.errors import MissingRequiredParameterError
 
 
+def filter_periodical_issues():
+    """Remove periodical issues."""
+    return ~Q("term", document_type="PERIODICAL_ISSUE")
```
ntarocco: please take into account also the doc `open_access` field

invenio_app_ils/config.py
equadon: I'm not happy with the "front site search" naming here either. It would be better if it was more generic and indicative of what it searches. `/records` is also a bit misleading too since it only searches documents and series and not all records.

How about `/literature`? We discussed using that for the front site routes when we planned the periodicals feature and would kind of make sense in this context. What do you think @kprzerwa @zzacharo @topless?

equadon: We discussed this on the standup and decided to call it literature instead of "front site search".

ui/src/pages/backoffice/Series/SeriesDetails/__tests__/__snapshots__/SeriesDetails.test.js.snap
```diff
@@ -7,8 +7,7 @@ exports[`SeriesDetails tests should load the details component 1`] = `
   >
     <Error>
       <SeriesMetadata />
-      <SeriesDocuments />
-      <SeriesMultipartMonographs />
+      <SeriesLiterature />
```
kpsherva: as I remember from our conversations. the Literature search was meant to be only for the fronstsite, I think for the librarians it should be still separated - it is clearer for managing

ui/src/pages/frontsite/components/SeriesCard/SeriesCard.js
```diff
@@ -0,0 +1,43 @@
...
+import { getCover } from '@pages/frontsite/config';
+
+export class SeriesCard extends Component {
+  render() {
+    const { data } = this.props;
+    const moi = data.metadata.mode_of_issuance;
+    return (
+      <Card link centered className="fs-book-card">
+        <Card.Meta className={'discrete'}>{moi}</Card.Meta>
+        <Image
```
kpsherva: please place here a placeholder image instead of the cover

ui/src/pages/frontsite/components/SeriesCard/SeriesCard.js
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Card, Image, Label } from 'semantic-ui-react';
+import get from 'lodash/get';
+import { getCover } from '@pages/frontsite/config';
+
+export class SeriesCard extends Component {
+  render() {
+    const { data } = this.props;
+    const moi = data.metadata.mode_of_issuance;
```
kpsherva, line 10: I think we should only display "series", for the end user the "deep" domain language might not be understandable

ui/src/pages/backoffice/Document/DocumentDetails/DocumentDetails.js
```diff
@@ -32,14 +32,13 @@ class DocumentHeader extends Component {
       <>
         <label>Document</label> #{data.metadata.pid}{' '}
         <CopyButton text={data.metadata.pid} />
-        {(
+        {data.metadata.created_by && (
           <>
             <br />
-            <label>Created by</label>{' '}
+            <label>Created by</label> <br />
```
topless: Missing the actual content here `{data.metadata.created_by}`

ui/src/pages/backoffice/Series/SeriesDetails/components/SeriesLiterature/SeriesLiterature.js
```diff
@@ -0,0 +1,111 @@
...
+import { BackOfficeRoutes } from '@routes/urls';
+import { SeeAllButton } from '@pages/backoffice/components/buttons';
+import { recordToPidType } from '@api/utils';
+import get from 'lodash/get';
+
+export default class SeriesLiterature extends Component {
+  constructor(props) {
+    super(props);
+    this.seriesPid = props.seriesDetails.metadata.pid;
+    this.seriesType = props.seriesDetails.metadata.mode_of_issuance;
```
topless: If these are required we should mention it to the PropTypes

ui/src/pages/frontsite/Documents/DocumentsSearch/DocumentSearchResultsGrid/DocumentSearchResultsGrid.js
```diff
@@ -3,11 +3,16 @@ import { BookCard } from '@pages/frontsite/components/BookCard';
...
 import { Card, Responsive } from 'semantic-ui-react';
 import { ResultsGrid } from 'react-searchkit';
+import { SeriesCard } from '@pages/frontsite/components/SeriesCard';
 
 export default class DocumentSearchResultsGrid extends Component {
   renderResultsGrid = results => {
-    const cards = results.map(book => {
-      return <BookCard key={book.metadata.pid} data={book} />;
+    const cards = results.map(result => {
+      return result.metadata.$schema.includes('documents/document') ? (
```
kpsherva, line 12: please, mark here with a comment to change the schema checking to pid_type check. As we plan to have pid object, I think it will be safer than to depend on schema url which is less "static"

topless (approved):

ui/src/pages/backoffice/Series/SeriesDetails/components/SeriesLiterature/SeriesLiterature.js
kpsherva: 👍 in general please try to add PropTypes everywhere, it is much more readable

invenio_app_ils/search/permissions.py
```diff
@@ -98,6 +95,9 @@ def search_factory_filter_by_patron(self, search):
     """Prepare query string to filter records by current logged in user."""
     def filter_by_patron(query_string):
         """Filter search results by patron_pid."""
+        if not current_user.is_authenticated:
```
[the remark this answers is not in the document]
equadon: I wanted to reuse `_ils_search_factory` but it always ensures that the user is authenticated which not all endpoints require. It made more sense to keep the `is_authenticated` check in `filter_by_patrons` since that's where it's applicable. @kprzerwa  Do you see any issues moving it there?

tests/api/test_record_relations.py
```diff
@@ -128,15 +128,7 @@ def _test_create_mm_document(create_using_pid1=True):
...
-                "relations": {
-                    "multipart_monograph": [
-                        {
-                            "pid": child_pid,
-                            "pid_type": child_pid_type,
-                            "relation_type": "multipart_monograph",
-                        }
-                    ]
-                },
+                "relations": {},
```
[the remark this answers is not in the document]
equadon: Since we stopped resolving the children for parents the `relations` had to be updated not to expect those relations to be there.

ui/src/api/literature/literature.js
```diff
@@ -0,0 +1,67 @@
...
+    } else {
+      this.withSeriesQuery.push(
+        `relations.multipart_monograph.pid:${prepareSumQuery(seriesPid)}`
+      );
+    }
+    return this;
+  }
+
+  includeAll() {
+    this.extraParamsQuery.push('include_all=yes');
```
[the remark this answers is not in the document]
equadon: I updated it so you can also supply `include_all` without parameter. I also kept the `y/yes/n/no/true/false` options using https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool

ui/src/api/literature/literature.js
```diff
@@ -0,0 +1,67 @@
...
+    return this;
+  }
+
+  sortBy(order = 'bestmatch') {
```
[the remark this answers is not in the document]
equadon: I'll raise an exception if no order is provided since we don't know if it's documents or series.

PR #701, opened 2020-01-14
circulation: item_pid from string to type/value

### LINKED ISSUE #702: Circulation: enable circulation of different types of items
EPIC: #637

Currently, `invenio-circulation` Loan schema has a single field `item_pid`. This is not enough to distinguish different item types.
Add the PID `type` in circulation for items so also Interlibrary Loans can circulate.

ntarocco:
* changes all implemented methods from circulation to provide the item
  pid as an object type/value and not anymore a string
* ILL borrowing request: adds document_pid field
* fixes record classes dict update() method signature
* changes UI loan details to navigate to item or ILL borrowing request
* fixes a UI bug when retrieving the available items for a loan
* fixes a UI bug when creating a loan request for a document
* closes #702


ntarocco:
This PR will fail until circulation is released. Plus I expect other tests to fail.

There are still a few issues to be solved:
1. we need to get rid of [ILS_DEFAULT_LOCATION_PID](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/config.py#L1436) because is should be "calculated" differently. I will create an issue for that.

> In the meanwhile, if you have to try to create or do actions on loans, you need to change your `/var/instance/invenio.cfg` and add:
`ILS_DEFAULT_LOCATION_PID="4zg8k-5qb35"`
the value is the location PID created via `scripts/setup`. Every time you run, you will have to update it in the `invenio.cfg`.

2. if you try to create loans, they will still have a numeric sequential PID: the fix is non trivial at the moment.
3. I want to refactor all the `if item or if borrowing request` to have better code, I will do in another PR.

invenio_app_ils/circulation/api.py
```diff
@@ -40,7 +40,7 @@ def _validate_delivery(delivery):
 
 def _set_item_to_can_circulate(item_pid):
     """Change the item status to CAN_CIRCULATE."""
-    item = Item.get_record_by_pid(item_pid)
+    item = Item.get_record_by_pid(item_pid["value"])
```
kpsherva, line 43: just thoughts for future: Item.get_record_by_pid should probably accept the `{pid: ... pid_type:...}` object and check if the record we try to get has a correct `pid_type`

ui/src/pages/backoffice/Loan/LoanDetails/components/ItemMetadata/ItemMetadata.js
```diff
@@ -51,18 +51,10 @@ export default class ItemMetadata extends Component {
         <Grid.Column>
           <Table basic="very" definition className="metadata-table">
             <Table.Body>
-              <Table.Row>
-                <Table.Cell>Document ID</Table.Cell>
```
kpsherva, line 55: we do not show then anymore?

ui/src/pages/backoffice/Loan/LoanDetails/components/LoanMetadata/LoanMetadata.js
```diff
@@ -76,16 +78,22 @@ export default class LoanMetadata extends Component {
     );
   }
 
+  getLinkToItem = itemPid => {
```
kpsherva, line 81: there is another usage below, could it be a component (next PR)? I'm trying to avoid the methods in favour of the components, it's easier to give them a common look

kpsherva (approved):

invenio_app_ils/circulation/api.py
ntarocco: Sure, we can see how to improve this. But in this case, you know that you want exactly an `Item`, no? The thing is that it is a very specific case this thingy of `item_pid` that is a dict... if we had for all `***_pid` fields, then I agree, it would make a lot of sense!

ui/src/pages/backoffice/Loan/LoanDetails/components/ItemMetadata/ItemMetadata.js
ntarocco: True, I could have left the doc pid, but it anyway requires quite some refactoring the entire view... So we will work on that!

ui/src/pages/backoffice/Loan/LoanDetails/components/LoanMetadata/LoanMetadata.js
ntarocco: Oh yes, sure! I will take care of this in the next PR!

PR #713, opened 2020-01-15
auth: remove localstorage flag

zzacharo:


ui/src/authentication/state/actions.js
```diff
@@ -31,9 +33,15 @@ export const fetchUserProfile = () => {
...
-export const setAnonymous = () => ({
-  type: IS_ANONYMOUS,
-});
+export const setAnonymous = () => {
+  return async dispatch => {
+    sessionManager.setAnonymous();
+    dispatch({
+      type: IS_ANONYMOUS,
+    });
+    goTo(FrontSiteRoutes.home);
```
topless, line 42: I am not sure if the `actions` is the right place for the redirection to happen. The setAnonymous action should do exactly that, and leave the rest of the logic to the controller who is calling it, in `ILSMenu` for the particular case.

zzacharo: I wanted to centralize the logic through this action but if you think the redirection is not worth to be centralized I am fine with that...@equadon @kprzerwa WDYT?

kpsherva: I think it should be centralised there in the action, for any other component who will use that action we will have the consistent behaviour of logout. Otherwise each controller who is using it would have to implement the redirect and this opens the gate for inconsistencies.

equadon: I don't have a strong preference but it makes sense what @topless said and will give us some more flexibility if we dispatch this action elsewhere and we need a different redirect url.

topless: @kprzerwa inconsistencies yes, but customization too, how will you alter the behaviour if you don't want it to redirect to home (if that is even a case)? (Just doing the devil's advocate here)

- I think the name `logout()` is also more descriptive than `setAnonymous()`
- On another note, you can fire a notification here for the user to be come aware that signed out.

PR #714, opened 2020-01-15
auth: remove `accounts` prefix from login/confirm-email pages.

zzacharo:
* Refactor INSTALL.rst docs


kpsherva (approved):

equadon (approved):

ui/src/authentication/pages/Login/Login.js
```diff
@@ -61,3 +61,14 @@ export default class Login extends Component {
     );
   }
 }
+
+export default class Login extends Component {
```
topless, line 65: Is this component needed? Putting the Notifications in the LoginPage wouldn't do the job?

ui/src/authentication/state/actions.js
```diff
@@ -33,8 +33,9 @@ export const fetchUserProfile = () => {
   };
 };
 
-export const setAnonymous = () => {
+export const logout = () => {
   return async dispatch => {
+    await authenticationService.logout();
     sessionManager.setAnonymous();
```
topless, line 39: Should we make this `setAnonymous` also `logout`?

topless (approved):

ui/src/authentication/pages/Login/Login.js
zzacharo: No because in every rerender the Notifications component would be initialized again losing any information :)

ui/src/authentication/state/actions.js
zzacharo: @topless Yes makes sense! Do you want, just for fast merging of this one, to include it in one of the upcoming PRs?

zzacharo: @topless I was rethinking a bit this renaming. The `SessionManger` is used as an internal storage reflecting the redux state about user data and if the user is anonymous. In that context wouldn't be clearer to keep the names as it is (`setUser` -> when you have user data, `setAnonymous` -> when user you don't have user data) ?

PR #720, opened 2020-01-20
frontsite: series details page

equadon:
* New UI route for books: `/literature/abcde-12345`
* New UI route for series: `/series/abcde-12345`

The series literature search will come in the next PR.

fs-series-details
fs-series-details-titles
fs-series-details-notes-empty


ui/src/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesPanel.js
```diff
@@ -0,0 +1,79 @@
...
+                    />
+                  </ILSParagraphPlaceholder>
+                  <ILSParagraphPlaceholder
+                    linesNumber={20}
+                    isLoading={isLoading}
+                  >
+                    <SeriesAbstract lines={20} />
+                  </ILSParagraphPlaceholder>
+                </Grid.Column>
+                {hasAccessUrls && (
```
kpsherva: what is displayed when there is no ULRs?

ui/src/pages/frontsite/components/Series/SeriesAbstract.js
```diff
@@ -0,0 +1,23 @@
+import ShowMore from 'react-show-more';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+export class SeriesAbstract extends Component {
```
kpsherva: Actually this could be general component `Abstract`, both for series and documents, as I remember they are the same. they work also the same way

ui/src/pages/frontsite/components/Series/SeriesIdentifiers.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import capitalize from 'lodash/capitalize';
+import get from 'lodash/get';
+import { Table, Divider } from 'semantic-ui-react';
+import { SeparatedList, InfoPopup } from '@components';
+
+export const SeriesIdentifiers = ({ metadata }) => {
```
kpsherva: this is similar case as with Abstract, we could use already existing DocumentIdentifiers, especially because the formatting in the metadata is a bit different and it should be unified

ui/src/pages/frontsite/components/Series/SeriesLiterature.js
```diff
@@ -0,0 +1,13 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Divider } from 'semantic-ui-react';
+
+export class SeriesLiterature extends Component {
+  render() {
+    return <Divider horizontal>Literature in this series</Divider>;
```
kpsherva, line 7: please leave a comment next time that you will continue on this section, I was confused for a moment why there is only a divider :)

ui/src/pages/frontsite/components/Series/SeriesMetadataAccordion.js
```diff
@@ -0,0 +1,117 @@
...
+        <Accordion.Title
+          active={activeIndex === 'identifiers'}
+          index="identifiers"
+          onClick={this.handleClick}
+        >
+          <Icon name="dropdown" />
+          Identifiers
+        </Accordion.Title>
+        <Accordion.Content active={activeIndex === 'identifiers'}>
+          <EmptyMessage
```
kpsherva: IMO this should be inside the `Identifiers` component below . The component should know what to do when the Identifiers are empty

ui/src/pages/frontsite/components/Series/SeriesMetadataAccordion.js
```diff
@@ -0,0 +1,117 @@
...
+        <Accordion.Title
+          active={activeIndex === 'notes'}
+          index="notes"
+          onClick={this.handleClick}
+        >
+          <Icon name="dropdown" />
+          Notes
+        </Accordion.Title>
+        <Accordion.Content active={activeIndex === 'notes'}>
+          <EmptyMessage
```
kpsherva: same here

ui/src/pages/frontsite/components/Series/SeriesMetadataAccordion.js
```diff
@@ -0,0 +1,117 @@
...
+          active={activeIndex === 'links'}
+          index="links"
+          onClick={this.handleClick}
+        >
+          <Icon name="dropdown" />
+          Links
+        </Accordion.Title>
+        <Accordion.Content active={activeIndex === 'links'}>
+          <EmptyMessage
+            show={!(isEmpty(metadata.access_urls) && isEmpty(metadata.urls))}
```
kpsherva: and here

ui/src/pages/frontsite/components/Series/SeriesMetadataTabs.js
```diff
@@ -0,0 +1,96 @@
...
+          <Tab.Pane attached={false}>
+            <SeriesInfo />
+          </Tab.Pane>
+        ),
+      },
+      {
+        menuItem: 'Identifiers',
+        render: () => (
+          <Tab.Pane attached={false}>
+            <EmptyMessage
```
kpsherva: and here - then you wouldn't have to also use it here

ui/src/semantic-ui/site/globals/site.overrides
```diff
@@ -111,6 +111,12 @@ form {
     color: @fsSecondaryColor;
   }
 
+  .info-popup {
```
kpsherva: this should be in `icon.overrides` because it modifies the style of icon

ui/src/pages/frontsite/Series/SeriesDetails/SeriesPanel/SeriesPanel.js
equadon: If there are no access urls there will be no action to take so it'll give the space to the abstract. But maybe it'd be more consistent if we kept it there saying there are no access urls?

ui/src/pages/frontsite/components/Series/SeriesAbstract.js
equadon: Yes, they do. It would make sense indeed to use the same component for those two.

ui/src/pages/frontsite/components/Series/SeriesIdentifiers.js
equadon: There is only `DocumentIdentifiers` for the backoffice and that one uses `MetadataTable`. There is an issue open for creating document identifiers for the frontsite but I could rename the series one and reuse on the document details page.

ui/src/pages/frontsite/components/Series/SeriesLiterature.js
equadon: Ahh, sure. I left a comment in the PR message but it's very easy to miss so it's better to have it in the code.

ui/src/pages/frontsite/components/Series/SeriesUrls.js
```diff
@@ -0,0 +1,37 @@
...
+import PropTypes from 'prop-types';
+import { List } from 'semantic-ui-react';
+import get from 'lodash/get';
+import isEmpty from 'lodash/isEmpty';
+import _truncate from 'lodash/truncate';
+
+const Url = ({ truncate, url }) => {
+  const description = url.description || url.value;
+  return (
+    <a href={url.value}>
```
topless, line 11: If they are used for in-app navigation, maybe we should consider using a `Link`

topless (approved):

ui/src/pages/frontsite/Series/SeriesDetails/__tests__/SeriesDetails.test.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { shallow, mount } from 'enzyme';
+import * as testData from '@testData/series.json';
```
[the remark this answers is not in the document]
equadon: Thanks! Didn't know JSON was exported as default. :)

ui/src/pages/frontsite/components/Series/SeriesUrls.js
equadon: I double-checked now and these URLs will be external and not in-app navigation. But perhaps an `ExternalLink` component could be useful that also could open the link in a new tab.

PR #733, opened 2020-02-10
relations: previous and next related documents

### LINKED ISSUE #717: Periodicals: Serial to serial relation
- [x] Create a new relation prev/next

relation name: chain
features
`A->B->C`

`A` is continued by `B`, `B` continued by `C`

`B` is a continuation of `A`
`B` is continued by `C`

It should support multiple prev and multiple next, e.g.

```
A1                 C1
     ->  B  ->
A2                 C2
```
It should block any "looping" possibilities

Examples: 
* https://cds.cern.ch/record/229951
* https://cds.cern.ch/record/229705

- [x] on the backoffice we need to be able to manage the relations of type chain in the same fashion as the other relations.

- [x] it needs to be possible to relate the series by the language and edition too.

topless:
closes #717

### DIFF
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -1294,12 +1294,24 @@ SERIAL_RELATION = ILS_RELATION_TYPE(
     "invenio_pidrelations.serializers.schemas.RelationSchema",
     "invenio_app_ils.records.relations.api.ParentChildRelation",
 )
+SEQUENCE_RELATION = ILS_RELATION_TYPE(
+    5,
+    "sequence",
+    "Sequence",
+    "invenio_app_ils.records.relations.nodes:PIDNodeRelated",
+    "invenio_pidrelations.serializers.schemas.RelationSchema",
+    "invenio_app_ils.records.relations.api.SequenceRelation",
+)
+
 
 PARENT_CHILD_RELATION_TYPES = [MULTIPART_MONOGRAPH_RELATION, SERIAL_RELATION]
 
 SIBLINGS_RELATION_TYPES = [LANGUAGE_RELATION, EDITION_RELATION, OTHER_RELATION]
 
-ILS_PIDRELATIONS_TYPES = PARENT_CHILD_RELATION_TYPES + SIBLINGS_RELATION_TYPES
+SEQUENCE_RELATION_TYPES = [SEQUENCE_RELATION]
+
+ILS_PIDRELATIONS_TYPES = PARENT_CHILD_RELATION_TYPES \
+    + SIBLINGS_RELATION_TYPES + SEQUENCE_RELATION_TYPES
 
 # The HTML tags allowed with invenio_records_rest.schemas.fields.sanitizedhtml
 ALLOWED_HTML_TAGS = []
--- a/invenio_app_ils/records_relations/api.py
+++ b/invenio_app_ils/records_relations/api.py
@@ -11,7 +11,8 @@ from copy import deepcopy
 from flask import current_app
 
 from invenio_app_ils.errors import RecordRelationsError
-from invenio_app_ils.relations.api import ParentChildRelation, SiblingsRelation
+from invenio_app_ils.relations.api import ParentChildRelation, \
+    SequenceRelation, SiblingsRelation
 
 
 class RecordRelationsMetadata(object):
@@ -109,24 +110,6 @@ class RecordRelationsRetriever(object):
         r.update(kwargs)
         return r
 
-    def _build_parent(self, child_pid, relation_type):
-        """Return the relation object with metadata."""
-        # this is a parent, relations_metadata are stored here
-        pid = child_pid.pid_value
-        pid_type = child_pid.pid_type
-
-        r = RecordRelationsMetadata.build_metadata_object(pid, pid_type)
-
-        # get `relations_metadata` field data
-        metadata = RecordRelationsMetadata.get_metadata_for(
-            self.record, relation_type, pid, pid_type
-        )
-        # merge with any `relations_metadata`
-        r.update(metadata or {})
-        r["relation_type"] = relation_type
-
-        return r
-
     def _build_child(self, parent_pid, child_pid, relation_type):
         """Return the relation object with metadata of the parent."""
         # this is a child, relations_metadata are stored in the parent
@@ -153,10 +136,16 @@ class RecordRelationsRetriever(object):
         # merge with any `relations_metadata`
         r.update(metadata or {})
 
-        # add also the title of the parent
         r["title"] = parent.get("title", "")
-        r["relation_type"] = relation_type
+        languages = parent.get("languages")
+        if languages:
+            r["languages"] = languages
 
+        edition = parent.get("edition", "")
+        if edition:
+            r["edition"] = edition
+
+        r["relation_type"] = relation_type
         return r
 
     def _build_sibling(self, related_pid, relation_type):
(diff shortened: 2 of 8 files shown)


invenio_app_ils/relations/api.py
```diff
@@ -294,3 +293,64 @@ def remove(self, pid):
...
+
+    def children_of(self, pid):
+        """Get all children PID for relations where the given PID is parent."""
+        return [r.child for r in self.get_relations_by_parent(pid)]
+
+    def parents_of(self, pid):
+        """Get all parents PID for relations where the given PID is child."""
+        return [r.parent for r in self.get_relations_by_child(pid)]
+
+    def relation_exists(self, parent_pid, child_pid):
```
ntarocco: Here you override the parent `relation_exists` method: is this on purpose? Because I think it is more correct to check for both A -> B and B -> A.

invenio_app_ils/config.py
```diff
@@ -1294,12 +1294,24 @@ def _(x):
...
+    "invenio_app_ils.records.relations.api.ChainRelation",
+)
+
 
 PARENT_CHILD_RELATION_TYPES = [MULTIPART_MONOGRAPH_RELATION, SERIAL_RELATION]
 
 SIBLINGS_RELATION_TYPES = [LANGUAGE_RELATION, EDITION_RELATION, OTHER_RELATION]
 
-ILS_PIDRELATIONS_TYPES = PARENT_CHILD_RELATION_TYPES + SIBLINGS_RELATION_TYPES
+LINKED_RELATION_TYPES = [CHAIN_RELATION]
```
ntarocco: Nothing to do with your code, but I think we should not mix Chain and Link/Linked terms. I think we should find one and stick to it, WDYT? @kprzerwa @zzacharo any input too?
Checking the definition of [Linked list](https://en.wikipedia.org/wiki/Linked_list) for example, it is mentioned of order and sequence.

What about something like:
`SEQUENCES_RELATION_TYPES` = [SEQUENCE_RELATION]`

And then we stick everywhere to sequence or whatever term we chose.

ntarocco:
It would be nice to have `loaders` instead of checking manually payload, but we can them in a second moment.

invenio_app_ils/records_relations/api.py
```diff
@@ -109,11 +110,11 @@ def _build_relations_object(self, pid_value, pid_type, **kwargs):
         r.update(kwargs)
         return r
 
-    def _build_parent(self, child_pid, relation_type):
+    def _build_parent(self, parent_pid, relation_type):
```
ntarocco: actually it looks to me that this `_build_parent` is never called. If true, you can remove it.

invenio_app_ils/records_relations/api.py
```diff
@@ -192,6 +193,26 @@ def _build_sibling(self, related_pid, relation_type):
...
+        from invenio_app_ils.records.api import IlsRecord
+        seq_record = IlsRecord.get_record_by_pid(pid, pid_type=pid_type)
+
+        metadata = RecordRelationsMetadata.get_metadata_for(
+            self.record, relation_type, pid, pid_type
+        )
+        r.update(metadata or {})
+        r["title"] = seq_record.get("title", "")
+        r["relation_type"] = relation_type
+
```
ntarocco, line 202: @kprzerwa might appreciate extra metadata here. Language or edition? Actually, you should fix all of them and add the edition and the date. Library wants to see always the title, edition and date when showing a document, if set. This is because 2 docs could have the same title...

invenio_app_ils/records_relations/api.py
```diff
@@ -424,6 +455,96 @@ def remove(self, first, second, relation_type):
...
         )
-
-        # return the allegedly modified record
         return first, second
+
+
+class RecordRelationsSequence(RecordRelations):
+    """Add/Remove operations for Sequence relations."""
+
+    allowed_metadata = ["title"]
```
ntarocco: you should remove this. This is a check for "extra" metadata of the relation, e.g. the `note` for the sibling relation of type `other`. It is not for the metadata of the document/series.

invenio_app_ils/records_relations/views.py
```diff
@@ -214,6 +213,96 @@ def _delete_sibling_relation(self, record, relation_type, payload):
         )
         return modified_record, record, second
 
+    def _validate_sequence_creation_payload(self, record, payload):
+        """Validate the payload when creating a new sequence relation."""
+        # NOTE: check next, and previous for two way references.
```
ntarocco: what do you mean with this note?

invenio_app_ils/records_relations/views.py
```diff
@@ -244,11 +336,18 @@ def create(payload):
...
-            if (
-                modified.pid == record.pid
-                and modified._pid_type == record._pid_type
-            ):
+            def is_modified(x, r):
+                return (x.pid == r.pid and x._pid_type == r._pid_type)
+
+            # NOTE: modified can be a record or a list of records, if one
+            # matches our record return the modified one.
+            if isinstance(modified, list):
```
ntarocco: to keep it maybe more readable, why not:
```
_modified = modified if isinstance(modified, list) or [modified]
```

invenio_app_ils/relations/api.py
```diff
@@ -294,3 +293,53 @@ def remove(self, pid):
...
+        """Get all children PID for relations where the given PID is parent."""
+        return [r.child for r in self.get_relations_by_parent(pid)]
+
+    def previous_relations(self, pid):
+        """Get all parents PID for relations where the given PID is child."""
+        return [r.parent for r in self.get_relations_by_child(pid)]
+
+    def add(self, previous_pid, next_pid):
+        """Add a new relation between previous and next pid."""
+        if self.relation_exists(previous_pid, next_pid):
```
ntarocco, line 311: so here in the end we are not checking recursively. Do we want to allow that?

tests/api/test_record_relations.py
```diff
@@ -1085,3 +1084,245 @@ def test_siblings_relations(client, json_headers, testdata, users):
...
+        res = client.post(url, headers=json_headers, data=json.dumps(payload))
+        assert res.status_code == status_code
+        if status_code == 400:
+            error = json.loads(res.data.decode("utf-8"))
+            assert "message" in error
+            assert next_pid in error["message"]
+            assert previous_pid in error["message"]
+
+
+def _test_sequence_document(client, json_headers):
```
ntarocco: just about naming, is it more `_test_sequence_series`? And below the same.

invenio_app_ils/records_relations/api.py
topless: Thanks for the clarification.

invenio_app_ils/records_relations/views.py
topless: I meant, in broken English, that `previous` and `next` pids are both required.

invenio_app_ils/records_relations/views.py
topless: Perfect, but all these inconsistencies will go away once we split the endpoints.

tests/api/test_record_relations.py
topless: I have already changed that in the upcoming request, and ported the change in this branch too, I went for `_test_sequence` since its only allowed for series, is redundant to mention it.

invenio_app_ils/records_relations/api.py
topless: I have added `title` and `edition`, `publication_year` is not part of the series schema but I created a good first issue #736

ntarocco (approved):

invenio_app_ils/records_relations/api.py
kpsherva: thanks

invenio_app_ils/records_relations/views.py
```diff
@@ -214,6 +213,94 @@ def _delete_sibling_relation(self, record, relation_type, payload):
...
+
+    def _create_sequence_relation(self, record, relation_type, payload):
+        """Create a Sequence relation.
+
+        Expected payload:
+
+            {
+                next_pid: <pid_value>,
+                next_pid_type: <pid_type>,
+                previous_pid: <pid_value>,
```
kpsherva, line 251: from the conversation I remember we create only the `-> next` relation, no? can we also create the previous?

kpsherva (approved):

invenio_app_ils/records_relations/views.py
topless: We had a couple of talks with @ntarocco about how we want to create the relations. My initial approach was to be explicitly set `next` in one request and `previous` on another, but we agreed to keep it simple and create the two way connection at once, which is the most common (if not every) case.

PR #738, opened 2020-02-14
[FIXING] fix: acq order

topless:
- created_by_pid should be string
- checks for missing props when we serialize

### DIFF
--- a/invenio_app_ils/document_requests/views.py
+++ b/invenio_app_ils/document_requests/views.py
@@ -97,7 +97,7 @@ class PendingRequestResource(DocumentRequestActionResource):
     @need_permissions("document-request-pending")
     def post(self, pid, record, **kwargs):
         """Pending request post method."""
-        # expecting remove_fields list of fields to remove
+        # NOTE: remove_fields is a list of props to remove from the record
         data = self.loader()
         remove_fields = data.get("remove_fields", [])
 
--- a/ui/src/pages/backoffice/Acquisition/Order/OrderEditor/components/OrderForm/OrderForm.js
+++ b/ui/src/pages/backoffice/Acquisition/Order/OrderEditor/components/OrderForm/OrderForm.js
@@ -18,20 +18,24 @@ const orderSubmitSerializer = (values, newRecord) => {
   if (newRecord) {
     submitValues.created_by_pid = sessionManager.user.id;
   }
-  submitValues.vendor_pid = values.vendor.pid;
-  submitValues.order_lines = values.resolved_order_lines.map(line => {
-    if (line.document) {
-      line.document_pid = _has(line.document, 'id')
-        ? line.document.id
-        : line.document.pid;
-    }
-    if (line.patron) {
-      line.patron_pid = _has(line.patron, 'id')
-        ? line.patron.id
-        : line.patron.pid;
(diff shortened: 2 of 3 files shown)


ui/src/pages/backoffice/Acquisition/Order/OrderEditor/components/OrderForm/OrderForm.js
```diff
@@ -16,22 +16,26 @@ import _has from 'lodash/has';
 const orderSubmitSerializer = (values, newRecord) => {
   const submitValues = { ...values };
   if (newRecord) {
-    submitValues.created_by_pid = sessionManager.user.id;
+    submitValues.created_by_pid = String(sessionManager.user.id);
```
kpsherva: did you check if it is possible to cast the `user.id` to string when it's fetched so we don't have to do it in every place, like here? because we need it on other places too

invenio_app_ils/document_requests/views.py
```diff
@@ -97,7 +97,7 @@ class PendingRequestResource(DocumentRequestActionResource):
     @need_permissions("document-request-pending")
     def post(self, pid, record, **kwargs):
         """Pending request post method."""
-        # expecting remove_fields list of fields to remove
+        # NOTE: remove_fields is a list of props to remove from the record
```
ntarocco, line 100: We need to discuss this IRL! :) I am not sure what exactly is this `remove_fields`.

ui/src/pages/backoffice/Acquisition/Order/OrderEditor/components/OrderForm/OrderForm.js
topless: @kprzerwa I will revert this, in favour of the upcoming change where we cast them all in the `api`.

ntarocco: This is what we discussed yesterday. Now re-thinking about that, while I think it would be more correct to keep it as an `int` and cast to String when using REST endpoints that require this field as String, it is probably just easier to always get and send String, otherwise every time we read a loan/purchase/ill, etc. in JS, we will have to cast it to `int`.

ui/src/pages/backoffice/Acquisition/Order/OrderEditor/components/OrderForm/OrderForm.js
```diff
@@ -16,22 +16,26 @@ import _has from 'lodash/has';
...
+  }
+  if (_has(values, 'vendor.pid')) {
+    submitValues.vendor_pid = values.vendor.pid;
+  }
+  if (_has(values, 'resolved_order_lines')) {
+    submitValues.order_lines = values.resolved_order_lines.map(line => {
+      if (line.document) {
+        line.document_pid = _has(line.document, 'id')
+          ? line.document.id
```
ntarocco, line 28: Can't remember why this `doc.id` and not always `doc.pid`, same for the patron, maybe we can discuss it IRL.

topless:
The changes for the serializer will come with #746

topless:
The changes for the serializer will come with #746

PR #746, opened 2020-02-17
frontsite: patron can cancel its own literature request

### LINKED ISSUE #743: patron cancel literature request
- [x] patron can cancel its own literature request
- [x] rename the `PendingRequestResource` to be more explicit
- [x] fix the loader of `PendingRequestResource` to explicitly accept only document_pid or physical...provider and check for `state == PENDING`
- [x] rename the front-end from onChangeDoc to onRemovedDoc
- [x] cast patron id to string in `fetchProfile` method
- [x] #742 add the `circulation` field in the loan resolver in `loan.py:document_resolver(loan_pid)`
- [x] fix OrderSubmitForm serializer from #738 
- [x] restructure document-request endpoints per property and not action
- [x] confirmation modal for canceling literature request
- [x] tests for permissions
- [x] fix `list_permission_factory_imp`: patrons can search only their own doc reqs
- [ ] ~~patron can cancel its own loan when it is in `PENDING` state~~ #598

topless:
- closes #743


ntarocco (commented):
Currently, I see that `dreqid` had `update_permission_factory_imp=backoffice_permission,` so the patron will not be able to update it.
Did you try to be logged in as the Patron that created the request?

invenio_app_ils/document_requests/loaders/jsonschemas/document_request_accept.py
```diff
@@ -21,6 +19,4 @@ class Meta:
 
         unknown = EXCLUDE
 
-    document_pid = fields.Str()
-    physical_item_provider = fields.Nested(PhysicalItemProviderSchema)
-    state = fields.Str()
+    state = fields.Str(required=True)
```
ntarocco: I would remove it and have no payload of the `accept`, no loader. WDYT?

invenio_app_ils/document_requests/views.py
```diff
@@ -88,70 +95,163 @@ def __init__(self, serializers, ctx, *args, **kwargs):
...
-    """Pending document request action resource."""
+class DocumentRequestDocumentResource(DocumentRequestActionResource):
+    """Endpoint to manage the Document of the Document Request."""
 
-    view_name = "{}_pending_request"
+    view_name = "{}_document"
 
     @pass_record
-    @need_permissions("document-request-pending")
+    @need_permissions("document-request-accept")
```
ntarocco: Since all the actions have same permissions, maybe I would go for `document-request-actions` for all of them, WDYT?

invenio_app_ils/document_requests/loaders/jsonschemas/document_request_document.py
```diff
@@ -5,18 +5,18 @@
...
+class DocumentRequestDocumentSchemaV1(RecordMetadataSchemaJSONV1):
+    """Document Request Document schema."""
 
     class Meta:
         """Meta attributes for the schema."""
 
         unknown = EXCLUDE
 
-    remove_fields = fields.List(fields.Str())
+    document_pid = fields.Str()
```
ntarocco: This is not mandatory because on delete there is no `document_pid` payload, right?

invenio_app_ils/document_requests/views.py
```diff
@@ -88,70 +95,163 @@ def __init__(self, serializers, ctx, *args, **kwargs):
...
+
+        document_pid = data.get("document_pid")
+        if not document_pid:
+            raise DocumentRequestError("document_pid is required")
+
+        record["document_pid"] = document_pid
+        record.commit()
+        db.session.commit()
+        current_app_ils.document_request_indexer.index(record)
+        return self.make_response(pid, record, 202)
```
ntarocco, line 117: I was wondering why we return `202`  as status code for all these ones and not `200`. Is there any particular reason?

invenio_app_ils/document_requests/views.py
```diff
@@ -88,70 +95,163 @@ def __init__(self, serializers, ctx, *args, **kwargs):
...
 
         state = data.get("state")
-        if state:
-            record["state"] = state
+        if state != 'ACCEPTED':
+            raise DocumentRequestError(
+                "{} is not a valid state".format(record["state"])
+            )
 
+        record["state"] = state
```
ntarocco: if you take in consideration my previous comment, here: `record["state"] = "ACCEPTED"`

invenio_app_ils/document_requests/views.py
```diff
@@ -88,70 +95,163 @@ def __init__(self, serializers, ctx, *args, **kwargs):
...
+    @need_record_permission("reject_permission_factory")
     def post(self, pid, record, **kwargs):
         """Reject request post method."""
         data = self.loader()
         reject_reason = data.get("reject_reason")
         document_pid = data.get("document_pid")
 
+        if record["state"] != "PENDING":
+            raise DocumentRequestError(
+                "You cannot reject a Document Request that "
```
ntarocco: Minor: I would rather show to the user an error msg saying " You cannot cancel..." rather than "You cannot reject", because he clicked on Cancel button. It is the librarian that can reject.
Honestly, I would just change all `reject` to `cancel`, since now we have a proper cancel reason.

tests/api/document_requests/test_document_requests_permissions.py
```diff
@@ -0,0 +1,206 @@
+# -*- coding: utf-8 -*-
```
ntarocco, line 1: Nice! Clear and understandable!

tests/api/document_requests/test_document_requests_permissions.py
```diff
@@ -0,0 +1,206 @@
...
+#
+# Copyright (C) 2018 CERN.
```
ntarocco: Just change the date here

tests/data/document_requests.json
```diff
@@ -3,7 +3,12 @@
     "pid": "dreq-1",
     "state": "PENDING",
     "patron_pid": "1",
-    "title": "Test title 1"
+    "title": "Test title 1",
+    "document_pid": "docid-1",
+    "physical_item_provider": {
+      "pid": "acq-order-pid",
+      "pid_type": "acuisition"
```
ntarocco: Very minor: typo

ui/src/authentication/services/AuthenticationService.js
```diff
@@ -32,7 +33,12 @@ class AuthenticationService {
   };
 
   fetchProfile = () => {
-    return http.get('/me');
+    return http.get('/me').then(res => {
+      if (_has(res, 'data.id')) {
+        res.data.id = res.data.id.toString();
```
ntarocco, line 38: 👍

ui/src/pages/backoffice/DocumentRequest/DocumentRequestDetails/components/DocumentRequestSteps/components/ReviewStep/ReviewStep.js
```diff
@@ -14,14 +13,9 @@ export const ReviewStep = ({ step }) => (
...
-  onAcceptClick = async () => {
+  onAcceptClick = () => {
     const { pid } = this.props.data;
-    const resp = await documentRequestApi.accept(pid, {
-      state: 'ACCEPTED',
-    });
-    if (resp.status === 202) {
-      this.props.fetchDocumentRequestDetails(pid);
-    }
+    this.props.acceptRequest(pid, { state: 'ACCEPTED' });
```
ntarocco: To be removed if don't pass any payload.

ui/src/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
```diff
@@ -50,23 +52,62 @@ export default class PatronCurrentDocumentRequests extends Component {
...
+      </Item.Group>
     );
   };
 
   renderNoResults = () => {
     return (
       <NoResultsMessage
-        messageHeader={'No loan requests'}
-        messageContent={'Currently you do not have any loan requests'}
+        messageHeader={'No literature requests'}
```
ntarocco: I am sorry if I stress on this: if I am the user, and I never did a request, I go to my profile, not sure I understand what is that.

What do you think of:
```
**Requests for new literature**
You did not create any request for literature not available in the library catalog yet.
```
And I would add a button:
```
Create a new request
```

ui/src/pages/frontsite/PatronProfile/PatronPendingLoans/PatronPendingLoans.js
```diff
@@ -44,6 +44,15 @@ class LoanRequestListEntry extends Component {
               <Item.Meta>
                 <DocumentAuthors metadata={loan.metadata.document} />
                 Requested on {toShortDate(loan.metadata.request_start_date)}
+                <br />
+                Valid until {toShortDate(loan.metadata.request_expire_date)}
```
ntarocco, line 49: Do we have a task to implement this? :)
Ping @kprzerwa maybe

ui/src/pages/frontsite/PatronProfile/PatronPendingLoans/PatronPendingLoans.js
```diff
@@ -44,6 +44,15 @@ class LoanRequestListEntry extends Component {
...
+                <Popup
+                  content={
+                    'If the request was not processed ' +
```
ntarocco, line 52: I think it is bad if we show to the user this: he might thinks that the librarian do not want to process his/her request! :)
Maybe better to highlight this in the backoffice, and not in the frontsite!

ntarocco (commented):
It looks neat! Thanks a lot for the refactoring! I find it more clear to understand now. Good job!

invenio_app_ils/document_requests/loaders/jsonschemas/document_request_document.py
topless: Exactly, if we want to be super conservative, in a previous commit, I was actually submitting the document_pid and I was checking if it matches the one we are trying to remove. We can go down either path wdyt?

invenio_app_ils/document_requests/views.py
topless: I didn't touch that part, the 202 is used in general when the request has been accepted for processing, but its not done yet if I am not wrong. I have seen it in other places too since the very beginning. I can always switch them to `200` but it would be better to talk with the team first I guess maybe someone has a better idea, @zzacharo @kprzerwa wdyt?

ui/src/pages/frontsite/PatronProfile/PatronPendingLoans/PatronPendingLoans.js
topless: That is existing code from previous, I just moved it on the left under the `Requested on` information so we can accommodate the cancel button on the right. I am sure @kprzerwa has a better idea what to do with it :D

invenio_app_ils/document_requests/loaders/jsonschemas/document_request_document.py
ntarocco: I think it is good enough! :)
I would maybe add a comment that on POST you have it, and on DELETE no.

invenio_app_ils/document_requests/views.py
ntarocco: I think we should switch to 200 where it makes sense!

ui/src/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
topless: The message is misleading indeed, I am not sure about the button though, maybe @kprzerwa has a stronger opinion 🥇

ui/src/pages/frontsite/PatronProfile/PatronPendingLoans/PatronPendingLoans.js
ntarocco: Sure, not saying that you added it, I was just "thinking out loud" when I saw it! :) :)

invenio_app_ils/document_requests/views.py
topless: For the reject -> cancel renaming, does it also apply for the backoffice?

ntarocco (approved):
LGTM! :)

invenio_app_ils/document_requests/loaders/jsonschemas/document_request_document.py
ntarocco: @topless what about if you use [partial](https://marshmallow.readthedocs.io/en/stable/quickstart.html#partial-loading)? You can set it to required here, so that in the `POST` method it is automatically checked and you can get rid of your own validation, and in the `DELETE`, use `partial=True`.

invenio_app_ils/document_requests/views.py
kpsherva: we could define first if the ongoing indexing means for us that the request was still not fully processed or not, IMO we could use the 202 to have user notifications of type "your request is being processed, the results will be available in x seconds" which would actually be more explanatory in the end why changes did not appear immediately.

ui/src/pages/frontsite/PatronProfile/PatronCurrentDocumentRequests/PatronCurrentDocumentRequests.js
kpsherva: I would add a link to search first, not a button to create a new request, to invite the user to search, instead of creating the requests for something which might exist

ui/src/pages/frontsite/PatronProfile/PatronPendingLoans/PatronPendingLoans.js
kpsherva: as I remember we didn't have a task but it makes sense to have this information here. @topless as you have made some changes in the looks, can we see a screen or demo?

ui/src/pages/frontsite/PatronProfile/PatronPendingLoans/PatronPendingLoans.js
kpsherva: I think that leaving this without the information will be confusing, as the request will just disappear  from the profile page of the user without giving him a chance to act upon the request which is not processed by the librarian.
In that case we should probably show the expired requests too and have a feature of "request again" or similar. We could discuss it in person.

ui/src/pages/frontsite/components/Document/DocumentLinks.js
```diff
@@ -21,7 +21,7 @@ export class DocumentLinks extends Component {
 
   hasPermissions = eitem => {
     if (!eitem.open_access) {
-      return sessionManager.authenticated;
+      return sessionManager.isAuthenticated();
```
kpsherva: thanks for fixing this, I have no idea why it was working for me, I will check it

kpsherva (approved):

invenio_app_ils/document_requests/loaders/jsonschemas/document_request_document.py
topless: @ntarocco I have made the `document_pid` required, and now I am submitting it also in the `DELETE` request. The main reason for that is that I didn't find a simple way to pass `partial=True` to the load method. As far I understood this is the place where the `load()` is called: https://github.com/inveniosoftware/invenio-records-rest/blob/526dcaef46dfe79acf737feee4570fd6e0b25679/invenio_records_rest/loaders/marshmallow.py#L91. According to the documentation for partial it should be something like `.load(request_json, partial=True). In invenio-records-rest we don't allow extra params to be passed down to load function. There are ways to fix/cheat this around, I am looking forward for you input cause I might have missed something in the process.

topless: As it concerns, submitting data with `DELETE` there is a worth reading SO question https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request, and a related comment on axios github https://github.com/axios/axios/issues/897#issuecomment-343715381

PR #748, opened 2020-02-18
ui: document details page fixes

kpsherva:
* adds link visible for librarian (from frontsite
  to backoffice
* adds link to the frontsite of the document from backoffice
* fixes the scrolling menu for the document details page

### DIFF
--- a/ui/src/pages/backoffice/Document/DocumentDetails/components/DocumentActionMenu/DocumentActionMenu.js
+++ b/ui/src/pages/backoffice/Document/DocumentDetails/components/DocumentActionMenu/DocumentActionMenu.js
@@ -8,6 +8,7 @@ import { BackOfficeRoutes } from '@routes/urls';
 import { isEmpty } from 'lodash';
 import React, { Component } from 'react';
 import { Button, Divider, Menu } from 'semantic-ui-react';
+import { Link as ScrollLink } from 'react-scroll';
 
 export default class DocumentActionMenu extends Component {
   constructor(props) {
@@ -43,14 +44,10 @@ export default class DocumentActionMenu extends Component {
     this.props.requestLoanForPatron(documentPid, patronPid, optionalParams);
   };
 
-  scrollTo(ref, menuItemName) {
-    ref.current.scrollIntoView(false, { behaviour: 'smooth', block: 'end' });
-    this.setState({ activeItem: menuItemName });
-  }
-
   render() {
-    const { document, relations, anchors } = this.props;
+    const { document, relations } = this.props;
     const { activeItem } = this.state;
+    const offset = -250;
     return (
       <div className={'bo-action-menu'}>
         <EditButton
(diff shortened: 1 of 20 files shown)


ntarocco (approved):

ui/src/pages/backoffice/Document/DocumentDetails/components/DocumentActionMenu/DocumentActionMenu.js
```diff
@@ -43,14 +44,10 @@ export default class DocumentActionMenu extends Component {
...
-  scrollTo(ref, menuItemName) {
-    ref.current.scrollIntoView(false, { behaviour: 'smooth', block: 'end' });
-    this.setState({ activeItem: menuItemName });
-  }
-
   render() {
-    const { document, relations, anchors } = this.props;
+    const { document, relations } = this.props;
     const { activeItem } = this.state;
+    const offset = -250;
```
topless, line 50: Since it doesn't change the `offset` variable doesn't need to be in the `render`, it can be declared as a class static or even a global const in the `invenioConfig` if it is or will be used elsewhere.

ui/src/pages/backoffice/Document/DocumentDetails/components/DocumentActionMenu/DocumentActionMenu.js
```diff
@@ -93,59 +90,91 @@ export default class DocumentActionMenu extends Component {
...
-            active={activeItem === 'header'}
-            onClick={(e, { name }) => this.scrollTo(anchors.summaryRef, name)}
+            name="metadata"
+            active={activeItem === 'metadata'}
+            activeClass="active"
+            as={ScrollLink}
+            to={'metadata'}
+            spy={true}
+            onSetActive={() => this.setState({ activeItem: 'metadata' })}
+            offset={offset}
```
topless, line 100: These menu items are super similar. They can be created programmatically somehow like this
```js
[
  {name: "metadata", label: "Metadata"},
  {name: "loan-requests", label: "loan-requests"},
].map(item => <Menu.Item name={item.name} spy={true} ... >{item.name}</Menu.Item>)

```

ui/src/pages/backoffice/Series/SeriesDetails/SeriesDetails.js
```diff
@@ -54,7 +47,7 @@ export default class SeriesDetails extends Component {
         title: 'Literature in this series',
         content: (
           <Accordion.Content>
-            <div ref={this.anchors.documents} id="series-documents">
+            <div id="series-documents">
               <SeriesDocuments />
             </div>
```
topless, line 52: Cant we skip the extra div and  make like
`<SeriesDocuments id="series-documents"/>`
or 
```
<Accordeon.Content id="series-documents">
  <SeriesDocuments />
</Accordeon.Content>
```

ui/src/pages/backoffice/Document/DocumentDetails/components/DocumentActionMenu/DocumentActionMenu.js
kpsherva: I think I had that discussion already somewhere. It can't global because they are not always the same. but per record type it makes sense to have it in common

ui/src/pages/backoffice/Document/DocumentDetails/components/DocumentActionMenu/DocumentActionMenu.js
kpsherva: I can make them as a reusable component, I think it is more "react way", I will do it in my next PR.

ui/src/pages/backoffice/Series/SeriesDetails/SeriesDetails.js
kpsherva: Accordeon :) this div was used there because not all props work in the `<Accordion`. However `id=` should appear in the final DOM. if possible I will fix it in next PR

PR #751, opened 2020-02-21
ILL: create a loan from a borrowing request

ntarocco:


invenio_app_ils/ill/api.py
```diff
@@ -102,7 +107,48 @@ def create(cls, data, id_=None, **kwargs):
...
+        """
+        tloc_pid = (
+            transaction_location_pid
+            or current_app.config["ILS_DEFAULT_LOCATION_PID"]
+        )
+        transaction_user_pid = transaction_user_pid or current_user.id
+        if arrow.get(self["loan_end_date"]) < arrow.now():
+            raise ILLError("The loan end date cannot be in the past.")
+
+        item_pid = dict(type=self._pid_type, value=self["pid"])
```
kpsherva, line 143: do you know if we only have this in ILS or at some point in invenio too?

invenio_app_ils/ill/schemas/ill_borrowing_requests/borrowing_request-v1.0.0.json
```diff
@@ -54,13 +54,17 @@
...
     "$schema": {
-      "title": "Interlibrary borrowing request schema",
+      "title": "InterLibrary borrowing request schema",
       "type": "string"
     },
     "cancel_reason": {
       "title": "Reason for cancelling the ILL borrowing request",
       "type": "string"
     },
+    "created_by_pid": {
```
kpsherva, line 64: I know that it is not the point of the PR, but as a suggestion, to unify this.
it is not possible that the borrowing request gets created by a bot, right?
Anyway, we have the `created_by` in the document, but there it might be also a harvest potentially. Why not go with (in every case where we need it):
```
created_by: {
  scheme: "user",
  value: "3"
}
```
it would be more flexible.

kpsherva (approved):

invenio_app_ils/ill/api.py
ntarocco: I don't think it will happen in Invenio for the time being...

invenio_app_ils/ill/schemas/ill_borrowing_requests/borrowing_request-v1.0.0.json
ntarocco: Good point. I honestly copy the `created_by_pid` field, but I will change it to something like this in the next PR! :)

invenio_app_ils/acquisition/api.py
```diff
@@ -85,6 +85,7 @@ def create(cls, data, id_=None, **kwargs):
         cls.build_resolver_fields(data)
         return super().create(data, id_=id_, **kwargs)
 
+    @preserve_fields(fields=["created_by_pid"])
```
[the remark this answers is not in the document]
ntarocco: I will also add it to the document then...

topless (approved):

ui/src/api/acquisition/serializers.js
```diff
@@ -1,64 +1,48 @@
...
-      const {
-        received_date,
-        expected_delivery_date,
-        order_date,
-      } = hit.metadata;
-      result.metadata = hit.metadata;
-      result.metadata.order_date = order_date ? fromISO(order_date) : null;
-      if (_hasIn(hit, 'metadata.received_date')) {
-        result.metadata.received_date = fromISO(received_date);
+const OrderSerializers = {
```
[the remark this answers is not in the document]
ntarocco: Yes, I guess it will be a change that will come slowly! :)

PR #765, opened 2020-03-11
fix: queues declare

topless:


ntarocco:
Not needed because already done by `index queue init`:
* see [ILS here](https://github.com/topless/invenio-app-ils/blob/master/invenio_app_ils/cli.py#L1272)
* see [invenio-indexer here](https://github.com/inveniosoftware/invenio-indexer/blob/master/invenio_indexer/cli.py#L107)
@topless can you please confirm or elaborate why this one was needed?

Issue #261, opened 2019-04-11
[REQ] Loan extension


kpsherva:
**As a patron, I can request extension for my loan**
The user can see his on-going (including expired) loans in his profile page. For each loan, a button labeled e.g. "Request extension" is displayed.
The button should extend the loan automatically, unless (order is important):
1. the action `extend` is not available for this loan
2. the number of extensions for this loan has reached the max (configurable)
3. the document of the loan is overbooked (more requests than available items)

If the patron cannot request for extension because the button is disabled, she/he should contact the library to address the issue. The library will choose what to do.

**As a librarian, I can always extend a loan**
From the backoffice, the librarian can always extend the loan overriding the checks for max number of extensions and overbooked loan.

- [ ] see specific issue #520

[Link to requirements](https://codimd.web.cern.ch/0p-ZhZDnTAycgWGNULBXBQ#UC-User-can-extend-a-loan)

---
**Nice to have**
It would be nice to model the extension request as real request object so that the patron can create a new extension request even if the conditions above are not met and the librarian can take actions (decide to extend or not) without using e-mails.

kpsherva:
it would be nice to develop after we have a user's profile page where they can extend by themselves

topless:
> The loan extension should be automatically approved if there is no waiting list for this document, otherwise Librarian decides:
> https://codimd.web.cern.ch/0p-ZhZDnTAycgWGNULBXBQ#UC-User-can-extend-a-loan

As I am trying to explain in my [comment](https://github.com/inveniosoftware/invenio-app-ils/issues/520#issuecomment-600000246) there is no structure currently to line up requests for extension. The user either is able to request an extension or not. I understand what they describe and what they want to do, but this feature will require further decisions like where do we store the extend requests, interface for the librarian to handle them etc ...

kpsherva:
so as I answered in the comment you've linked. The loan can be automatically approved if there is no waiting list for the book (or if they don't exceed the maximum allowed number. Otherwise we block the extend button for the patron, and it has to be handled via email

ntarocco:
The issue text has been updated accordingly.

PR #773, opened 2020-03-17
loan: patron extend loan

### LINKED ISSUE #520: [TASK] Loan extension
**Backend**
* implement checks and permissions for loan extension feature (taking into account patron and librarian)
* add tests

**Frontend**
* add the `Request extension` button to the patron profile page
* enable/disable based on conditions
* show human readable error messages

Requirement: #261

topless:
**[UPDATED]**
- introduced extend loan permission
- frontend checks for extend button
- refetch current loans after extend
- validation if document is overbooked
- validation max extensions
- validation default duration before request allowed
- the user is the owner of the loan or admin or librarian
- an extend action is available (frontend only)
- closes #520, #261

### DIFF
--- a/invenio_app_ils/circulation/utils.py
+++ b/invenio_app_ils/circulation/utils.py
@@ -70,7 +70,7 @@ def circulation_default_extension_duration(loan):
 
 def circulation_default_extension_max_count(loan):
     """Return a default extensions max count."""
-    return float("inf")
+    return 3
 
 
 def circulation_is_loan_duration_valid(loan):
@@ -94,7 +94,7 @@ def circulation_overdue_loan_days(loan):
 def circulation_upcoming_return_range():
     """Return a default upcoming return range."""
     return arrow.utcnow() + timedelta(
-        days=current_app.config["ILS_UPCOMING_RETURN_RANGE"])
+        days=current_app.config["ILS_LOAN_OVERDUE_DAYS_UPFRONT_NOTIFICATION"])
 
 
 def circulation_transaction_location_validator(transaction_location_pid):
--- a/invenio_app_ils/circulation/utils.py
+++ b/invenio_app_ils/circulation/utils.py
@@ -12,12 +12,13 @@ from __future__ import absolute_import, print_function
 from datetime import timedelta
 
 import arrow
-from flask import current_app
+from flask import current_app, g, has_request_context
 from flask_login import current_user
 
 from invenio_app_ils.errors import UnknownItemPidTypeError
 from invenio_app_ils.ill.api import BORROWING_REQUEST_PID_TYPE
 from invenio_app_ils.ill.proxies import current_ils_ill
+from invenio_app_ils.permissions import backoffice_permission
 from invenio_app_ils.pidstore.pids import ITEM_PID_TYPE
 from invenio_app_ils.proxies import current_app_ils
 
@@ -70,6 +71,9 @@ def circulation_default_extension_duration(loan):
 
 def circulation_default_extension_max_count(loan):
     """Return a default extensions max count."""
+    # NOTE: If user is admin or librarian always have 1 extra extension count.
+    if has_request_context() and backoffice_permission().allows(g.identity):
+        return loan.get("extension_count", 0) + 1
     return 3
 
 
@@ -92,7 +96,7 @@ def circulation_overdue_loan_days(loan):
 
 
 def circulation_loan_will_expire_days():
-    """Return a default upcoming return range."""
+    """Return a number of days before a loan expires."""
     return arrow.utcnow() + timedelta(
         days=current_app.config["ILS_LOAN_WILL_EXPIRE_DAYS"])
 
--- a/invenio_app_ils/circulation/utils.py
+++ b/invenio_app_ils/circulation/utils.py
@@ -91,10 +91,10 @@ def circulation_overdue_loan_days(loan):
     return (arrow.get().utcnow() - end_date).days
 
 
-def circulation_upcoming_return_range():
+def circulation_loan_will_expire_days():
     """Return a default upcoming return range."""
     return arrow.utcnow() + timedelta(
-        days=current_app.config["ILS_LOAN_OVERDUE_DAYS_UPFRONT_NOTIFICATION"])
+        days=current_app.config["ILS_LOAN_WILL_EXPIRE_DAYS"])
 
 
 def circulation_transaction_location_validator(transaction_location_pid):
--- a/invenio_app_ils/permissions.py
+++ b/invenio_app_ils/permissions.py
@@ -11,14 +11,22 @@ from __future__ import absolute_import, print_function
 
 from functools import wraps
 
+import arrow
 from flask import abort, current_app
 from flask_login import current_user
 from flask_principal import UserNeed
 from invenio_access import action_factory
 from invenio_access.permissions import Permission, authenticated_user
+from invenio_circulation.proxies import current_circulation
 from invenio_records_rest.utils import allow_all, deny_all
 
+from invenio_app_ils.errors import InvalidLoanExtendError
 from invenio_app_ils.proxies import current_app_ils
+
+from invenio_app_ils.circulation.utils import (  # isort:skip
+    circulation_default_extension_max_count,
+    circulation_overdue_loan_days
+)
 
 backoffice_access_action = action_factory("ils-backoffice-access")
 
@@ -87,6 +95,73 @@ class LoanOwnerPermission(Permission):
         super(LoanOwnerPermission, self).__init__(
             UserNeed(int(record["patron_pid"])), backoffice_access_action
         )
+
+
+class LoanExtendPermission(Permission):
+    """Return Permission to evaluate if the user can extend the loan."""
+
+    def __init__(self, record):
+        """Constructor."""
+        self.record = record
+
+        # NOTE: extra validation if the user is not librarian or admin
+        if current_user.id == int(record["patron_pid"]):
+            self.validate()
+
+        super().__init__(
+            UserNeed(int(record["patron_pid"])), backoffice_access_action
+        )
+
+    def validate(self):
+        """Validate if the loan can be extended."""
+        if not self.validate_extend_enabled():
+            raise InvalidLoanExtendError(
+                "You can extend this loan {} days before it "
+                "expires!".format(
+                    current_app.config["ILS_LOAN_WILL_EXPIRE_DAYS"])
+            )
+
+        if not self.validate_max_extensions():
+            raise InvalidLoanExtendError(
+                "You have reached the max number of extensions for THIS loan!"
+            )
+
+        if not self.validate_pending_loans():
+            raise InvalidLoanExtendError(
+                "There is a high demand for this literature!"
+            )
+
+        if self.validate_is_overdue():
+            raise InvalidLoanExtendError(
+                "This loan is overdue, its end date has passed!"
+            )
+
+    def validate_extend_enabled(self):
+        """Validate loan is in the period that can be extended."""
+        end_date = arrow.get(self.record["end_date"])
+        return (
+            (arrow.get().utcnow() - end_date).days <=
+            current_app.config["ILS_LOAN_WILL_EXPIRE_DAYS"]
+        )
+
+    def validate_max_extensions(self):
+        """Validate if we have reached the max extension count."""
+        return (
+            self.record["extension_count"] <=
+            circulation_default_extension_max_count(self.record)
+        )
+
+    def validate_pending_loans(self):
+        """Validate the loaned document has no pending loan requests."""
+        loan_search = current_circulation.loan_search_cls()
+        pending_loans_count = loan_search.get_pending_loans_by_doc_pid(
+            self.record["document_pid"]
+        ).count()
+        return pending_loans_count == 0
+
+    def validate_is_overdue(self):
+        """Validate if the loan is overdue."""
+        return circulation_overdue_loan_days(self.record) > 0
 
 
 class DocumentRequestOwnerPermission(Permission):
--- a/invenio_app_ils/permissions.py
+++ b/invenio_app_ils/permissions.py
@@ -11,22 +11,15 @@ from __future__ import absolute_import, print_function
 
 from functools import wraps
 
-import arrow
 from flask import abort, current_app
 from flask_login import current_user
 from flask_principal import UserNeed
 from invenio_access import action_factory
 from invenio_access.permissions import Permission, authenticated_user
-from invenio_circulation.proxies import current_circulation
 from invenio_records_rest.utils import allow_all, deny_all
 
 from invenio_app_ils.errors import InvalidLoanExtendError
 from invenio_app_ils.proxies import current_app_ils
-
-from invenio_app_ils.circulation.utils import (  # isort:skip
-    circulation_default_extension_max_count,
-    circulation_overdue_loan_days
-)
 
 backoffice_access_action = action_factory("ils-backoffice-access")
 
@@ -60,9 +53,19 @@ def check_permission(permission):
         abort(403)
 
 
+def authenticated_user_permission(*args, **kwargs):
+    """Return an object that evaluates if the current user is authenticated."""
+    return Permission(authenticated_user)
+
+
 def backoffice_permission(*args, **kwargs):
     """Return permission to allow only librarians and admins."""
     return Permission(backoffice_access_action)
+
+
+def circulation_permission(patron_pid):
+    """Return circulation status permission for a patron."""
+    return Permission(UserNeed(int(patron_pid)), backoffice_access_action)
 
 
 def file_download_permission(obj):
@@ -87,6 +90,19 @@ def files_permission(obj, action=None):
     return backoffice_permission()
 
 
+def loan_extend_circulation_permission(loan):
+    """Return permission to allow only owner and librarians to extend loan."""
+    if not current_user or not current_user.id:
+        abort(401)
+
+    if current_user.id == int(loan["patron_pid"]):
+        loan = loan.replace_refs()
+        if loan.get("document").get("circulation").get("overbooked"):
+            raise InvalidLoanExtendError("This document is overbooked!")
+
+    return LoanOwnerPermission(loan)
+
+
 class LoanOwnerPermission(Permission):
     """Return Permission to evaluate if the current user owns the loan."""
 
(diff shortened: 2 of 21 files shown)


ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,130 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { DateTime } from 'luxon';
+import { Button, Icon, Popup } from 'semantic-ui-react';
+import { invenioConfig, ES_DELAY } from '@config';
+import _get from 'lodash/get';
+import _has from 'lodash/has';
+
+export const INFO_MESSAGES = {
```
zzacharo: minor: Do you need this export? I see you use it only for tests so how about defining these messages as `static` inside the `ExtendButton` component? Or otherwise in separate file `messages.js` inside the folder?

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,130 @@
...
+    hasExtendAction: false,
+    hasMaxExtensions: false,
+    hasPendingLoans: false,
+    infoMessage: '',
+    isOverdue: false,
+    isOwner: false,
+  };
+
+  componentDidMount = () => {
+    this.checkMaxExtensions();
```
zzacharo: comment: I just want to stress out that these checks will run only when the component will be mounted. This means that if you depend on re running these after a prop change or a rerender they will not run :)

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/__tests__/ExtendButton.test.js
```diff
@@ -0,0 +1,160 @@
...
+        loan={loan}
+        user={otherUser}
+      />
+    );
+    expect(component.state().infoMessage).toEqual(INFO_MESSAGES.isOwner);
+    const btn = component.find('Button');
+    expect(btn.props().disabled).toBe(true);
+  });
+
+  it('should be disabled if extend action is in available actions', () => {
```
zzacharo: I think you meant to say `should be disabled if extend action is not in available actions` right?

zzacharo (commented):
I tried to use the guidelines that we are using in our reviews in RDM as described in the section [Collaboration Guidelines](https://cernbox.cern.ch/index.php/apps/files/?dir=/__myprojects/digital-repositories/Services_Projects/Invenio/Talks/2020&#pdfviewer) of the slides here. Let me know if they are useful or not :)

invenio_app_ils/circulation/utils.py
```diff
@@ -94,7 +94,7 @@ def circulation_overdue_loan_days(loan):
 def circulation_upcoming_return_range():
     """Return a default upcoming return range."""
     return arrow.utcnow() + timedelta(
-        days=current_app.config["ILS_UPCOMING_RETURN_RANGE"])
+        days=current_app.config["ILS_LOAN_OVERDUE_DAYS_UPFRONT_NOTIFICATION"])
```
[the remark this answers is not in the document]
topless: `ILS_LOAN_OVERDUE_DAYS_UPFRONT_NOTIFICATION` is ridiculous, this is a good reason to remain as is `upcoming_return_range` and add a better explanation in the `config.py`, my comments will be **old school** since your link points to a folder of files.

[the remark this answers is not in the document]
zzacharo: Or `ILS_OVERDUE_LOANS_FILTER_RANGE` plus better explanation in the `config.py`? :)

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,130 @@
...
+  };
+
+  render() {
+    const { loan } = this.props;
+    const showExtendButton =
+      DateTime.fromISO(loan.metadata.end_date).diffNow('days').days <=
+      invenioConfig.circulation.loanOverdueDaysUpfrontNotification;
+
+    return (
+      showExtendButton && (
```
ntarocco: I would say "always show" but disabled?

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,130 @@
...
+  };
+
+  checkHasExtendAction = () => {
+    const hasExtendAction = _has(this.props.loan, 'availableActions.extend');
+    this.setState({ hasExtendAction: hasExtendAction });
+    if (!hasExtendAction)
+      this.setState({ infoMessage: INFO_MESSAGES.hasExtendAction });
+  };
+
+  checkIsOwner = () => {
```
ntarocco: this is an extra check, maybe for safety?, but we should only return and display from ES the user's loans...

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,130 @@
...
+
+    await this.props.extendLoan(extendUrl, document.pid, patronPid);
+    setTimeout(() => {
+      onExtendSuccess();
+    }, ES_DELAY);
+  };
+
+  render() {
+    const { loan } = this.props;
+    const showExtendButton =
```
ntarocco: jumping on what Zac said, probably you should call the check of the button state here, e.g. `updateBtnState()` or whatever, and then run all the checks.
If you then find yourself doing that, I guess you don't need any more this `setState` (which re-triggers a rendering?) but just return value/msg (this complicates things though, I don't have a solution out of mind now :)).
I like your approach!

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,130 @@
...
+      DateTime.fromISO(loan.metadata.end_date).diffNow('days').days <=
```
ntarocco: This should be already a date object, otherwise can you please check the serializer and fix it?

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,130 @@
...
+
+export const INFO_MESSAGES = {
+  hasMaxExtensions: 'You have reached the max number of extensions for a loan!',
+  hasExtendAction:
+    'Manually extending this loan, is not currently an available action!',
+  hasPendingLoans:
+    'Other users requested that book, therefore you cannot extend your loan.',
+  isOverdue: 'You loan is overdue, therefore you cannot extend the loan!',
+  isOwner: 'You are not the owner of this loan, so you cannot extend it!',
+};
```
ntarocco: I would propose to change a couple of words here:
```
  hasMaxExtensions: 'You have reached the max number of extensions for <THIS> loan!',
  hasExtendAction:
    'Manually extending this loan, is not currently an available action!', -> '<It is not possible this loan>' the user does not know the technicalities of loan actions
  hasPendingLoans:
    'Other users requested that book, therefore you cannot extend your loan.', -> '<Unfortunately it is not possible to extend this loan because there are waiting requests for this literature' (to be improved probably)
  isOverdue: 'You loan is overdue, therefore you cannot extend the loan!', (do we allow this or not? maybe to be discussed)
```

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/LoanListEntry/LoanListEntry.js
```diff
@@ -39,10 +56,8 @@ export class LoanListEntry extends Component {
                 Loaned on {toShortDate(loan.metadata.start_date)}
               </Item.Meta>
               <Item.Description>
-                {}
-                You have extended this loan {
-                  loan.metadata.extension_count
-                } of {invenioConfig.loans.maxExtensionsCount} times
+                You have extended this loan {loan.metadata.extension_count} of{' '}
+                {invenioConfig.circulation.extensionsMaxCount} times
```
ntarocco: In CDS this will be a very high number and it won't be easy to hide it. May I suggest to remove the second part `of ... times`?

ui/src/pages/frontsite/PatronProfile/PatronPastLoans/PatronPastLoans.js
```diff
@@ -40,10 +40,8 @@ class PastLoanListEntry extends Component {
                 Loaned on {toShortDate(loan.metadata.start_date)}
               </Item.Meta>
               <Item.Description>
-                {}
-                You have extended this loan {
-                  loan.metadata.extension_count
-                } of {invenioConfig.loans.maxExtensionsCount} times
+                You have extended this loan {loan.metadata.extension_count} of{' '}
+                {invenioConfig.circulation.extensionsMaxCount} times
```
ntarocco: same here

ntarocco (commented):
While it is good that we do these checks in the frontend, we still need to implement the same checks in the backend for the user and, at the same time, allow that librarians can always extend.

invenio_app_ils/circulation/utils.py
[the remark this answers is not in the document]
zzacharo: @ntarocco I got it meant to mean but I just checked that in the end we are using it in the overdue facet to just say "Check all loans from now until than date" where that date is the value of the `circulation_upcoming_return_range `. So I just proposed to simplify the name if it will be used only there...But not strong opinion on this!

invenio_app_ils/permissions.py
```diff
@@ -89,6 +97,73 @@ def __init__(self, record):
...
+
+    def validate_max_extensions(self):
+        """Validate if we have reached the max extension count."""
+        return (
+            self.record["extension_count"] <=
+            circulation_default_extension_max_count(self.record)
+        )
+
+    def validate_pending_loans(self):
+        """Validate the loaned document has no pending loan requests."""
```
kpsherva: here we should check not if we have requests, but if the document is  overbooked. Because if we have 4 items but 2 requests I don't see the reason why not to extend.

invenio_app_ils/permissions.py
```diff
@@ -89,6 +97,73 @@ def __init__(self, record):
...
+        if self.validate_is_overdue():
+            raise InvalidLoanExtendError(
+                "This loan is overdue, its end date has passed!"
+            )
+
+    def validate_extend_enabled(self):
+        """Validate loan is in the period that can be extended."""
+        end_date = arrow.get(self.record["end_date"])
+        return (
+            (arrow.get().utcnow() - end_date).days <=
```
zzacharo: If you keep this, you should check when the `arrow.get().utcnow() - end_date` results to a negative result!

invenio_app_ils/permissions.py
```diff
@@ -89,6 +97,73 @@ def __init__(self, record):
...
 
+class LoanExtendPermission(Permission):
+    """Return Permission to evaluate if the user can extend the loan."""
+
+    def __init__(self, record):
+        """Constructor."""
+        self.record = record
+
+        # NOTE: extra validation if the user is not librarian or admin
+        if current_user.id == int(record["patron_pid"]):
```
ntarocco: Can I propose another version more similar to the other permissions? I think it might be more readable, but let me know what you think.

Transform this into a function:
```
def loan_extend_circulation_permission(loan):
    """ explanation """
    if not current_user or not current_user.id:
        abort(401)

    if current_user.id == loan["patron_pid"]:
        validate_automatic_loan_extension(loan)

    return is_loan_owner_permission(loan["patron_pid"])
```
where `is_loan_owner_permission` is the already existing function `circulation_permission` renamed.

This is because I find it a bit weird to have all these checks in an object constructor and also because we can extract and expose the loan automatic extension business logic to easily test it and eventually re-use it.

WDYT?

invenio_app_ils/permissions.py
```diff
@@ -89,6 +97,73 @@ def __init__(self, record):
...
+            self.validate()
+
+        super().__init__(
+            UserNeed(int(record["patron_pid"])), backoffice_access_action
+        )
+
+    def validate(self):
+        """Validate if the loan can be extended."""
+        if not self.validate_extend_enabled():
```
ntarocco: to be removed as discussed

invenio_app_ils/permissions.py
ntarocco: to be removed anyway

invenio_app_ils/permissions.py
```diff
@@ -89,6 +97,73 @@ def __init__(self, record):
...
+            raise InvalidLoanExtendError(
+                "You can extend this loan {} days before it "
+                "expires!".format(
+                    current_app.config["ILS_LOAN_WILL_EXPIRE_DAYS"])
+            )
+
+        if not self.validate_max_extensions():
+            raise InvalidLoanExtendError(
+                "You have reached the max number of extensions for THIS loan!"
```
ntarocco: why `THIS` uppercase? :)

invenio_app_ils/permissions.py
```diff
@@ -89,6 +97,73 @@ def __init__(self, record):
...
+        if not self.validate_pending_loans():
+            raise InvalidLoanExtendError(
+                "There is a high demand for this literature!"
+            )
+
+        if self.validate_is_overdue():
+            raise InvalidLoanExtendError(
+                "This loan is overdue, its end date has passed!"
+            )
+
```
ntarocco: I enumerated all the checks hopefully in a clear way in the issue, see: https://github.com/inveniosoftware/invenio-app-ils/issues/520

For the error messages, well they should not appear in the frontsite because there are checks there, but they appear in the APIs.

More I think about it and more I convince myself that we should have implemented errors in a different way:
* API consumers, e.g. a script or another app, should get a more technical message with a code
* frontsite, basically user interface, should probably ignore the technical message and use the code to re-map the error to something human readable and undestandable. WDYT?

ui/src/config/invenioConfig.js
```diff
@@ -9,6 +9,8 @@ export const invenioConfig = {
       PICKUP: 'Pick it up at the library desk',
       DELIVERY: 'Have it delivered to my office',
     },
+    extensionsMaxCount: 3,
```
ntarocco, line 12: Please add this to the issue above, thanks!

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,110 @@
...
+import _get from 'lodash/get';
+import _has from 'lodash/has';
+import _isEqual from 'lodash/isEqual';
+
+const expireDays = invenioConfig.circulation.loanWillExpireDays;
+
+export default class ExtendButton extends Component {
+  static INFO_MESSAGES = {
+    hasMaxExtensions:
+      'You have reached the max number of extensions for THIS loan!',
```
ntarocco: Why `THIS` uppercase?

ntarocco (commented):
Sorry, again many comments. I think this is probably a lesson learned as a team that we need to align first to then avoid to re-do the PR many times and create frustration. Unfortunately, we are not doing it :(

Do not forget the backend tests! :)

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,110 @@
...
+
+  render() {
+    const isDisabled = this.isDisabled();
+
+    return (
+      <>
+        <Button
+          color="purple"
+          size="mini"
+          content="extend loan"
```
ntarocco: Please change to `Request extension`

invenio_app_ils/permissions.py
topless: I was trying to avoid having everything in the constructor, will give it a try for sure.

invenio_app_ils/permissions.py
topless: I guess you wanted to stand out, on your comment your wrote `<THIS>` will make it right case :)

invenio_app_ils/permissions.py
zzacharo: I agree with that. We should not compromise the technical description for just more user friendly errors in the frontend :)

invenio_app_ils/circulation/utils.py
```diff
@@ -70,6 +71,8 @@ def circulation_default_extension_duration(loan):
 
 def circulation_default_extension_max_count(loan):
     """Return a default extensions max count."""
+    if 'librarian' in current_user.roles or 'admin' in current_user.roles:
```
topless: Not quire sure if this is the right place for this.

tests/api/circulation/test_loan_extend.py
```diff
@@ -0,0 +1,50 @@
...
+def test_loan_extend_stuff(
+    app, client, json_headers, users, testdata, app_config, loan_params
+):
+    """Test loan can be extended."""
+    # Create a Loan with patron_pid: 1
+    user_login(client, "admin", users)
+    loan_data = testdata["loans"][1]
+    loan = Loan.get_record_by_pid(loan_data["pid"])
+
+    # FIXME: AttributeError: '_AppCtxGlobals' object has no attribute 'identity'
```
topless: #HELPME! I might be doing it all wrong.

invenio_app_ils/permissions.py
```diff
@@ -79,6 +97,53 @@ def files_permission(obj, action=None):
...
+        validate_automatic_loan_extension(loan)
+
+    return is_loan_owner_permission(loan["patron_pid"])
+
+
+def validate_automatic_loan_extension(loan):
+    """Validate if the loan can be extended automatically."""
+    # NOTE: Validate max extensions
+    if not (
+        loan["extension_count"] <=
```
ntarocco: This is already checked by the transition itself. Do we keep it to be checked again?
If we keep it: `loan["extension_count"]` -> `loan.get("extension_count", 0)`
WDYT?

invenio_app_ils/permissions.py
```diff
@@ -79,6 +97,53 @@ def files_permission(obj, action=None):
...
+    pending_loans = loan_search.get_pending_loans_by_doc_pid(doc_pid).count()
+    active_loans = loan_search.get_active_loans_by_doc_pid(doc_pid).count()
+
+    item_search = current_app_ils.item_search_cls()
+    items = item_search.search_by_document_pid(doc_pid).count()
+    unavailable_items = item_search.get_unavailable_items_by_document_pid(
+        doc_pid).count()
+
+    remain_items = items - active_loans - unavailable_items
+    return pending_loans > remain_items
```
ntarocco: why not simply re-using the resolver?
```
doc_pid = loan["document_pid"]
Document = current_app_ils.document_record_cls()
doc = Document.get_record_by_pid(doc_pid)
doc.replace_refs()  # this is needed only if refs are not computed
return doc and doc["circulation"]["overbooked"]
```
WDYT?

tests/api/circulation/test_loan_extend.py
ntarocco: try to use `from flask_security import login_user` instead

ui/src/pages/backoffice/Series/SeriesDetails/__tests__/__snapshots__/SeriesDetails.test.js.snap
```diff
@@ -72,7 +72,7 @@ exports[`SeriesDetails tests should load the details component 1`] = `
                           </div>
                         </AccordionContent>,
                         "key": "series-documents",
-                        "title": "Literature in this series",
+                        "title": "Literature in this series (0)",
```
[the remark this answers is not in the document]
KonstantinaStoikou: I didn't run ```npm test```, I run the ```run-tests.sh``` only and it didn't show any error.

invenio_app_ils/circulation/utils.py
topless: If we do that for circulation let me note that we will have to adjust `_ensure_valid_extension` accordingly.

But I agree, circulation should not be bothered with validations about your max extensions, it should come in from the overlay.

topless: I created an issue https://github.com/inveniosoftware/invenio-circulation/issues/128 for future reference

invenio_app_ils/permissions.py
topless: Anything wrong with `loan = loan.replace_refs()`?

ui/src/pages/frontsite/PatronProfile/PatronPastLoans/PatronPastLoans.js
ntarocco: `of{' '} {invenioConfig.circulation.extensionsMaxCount} times` should be removed

ui/src/pages/frontsite/PatronProfile/PatronCurrentLoans/components/ExtendButton/ExtendButton.js
```diff
@@ -0,0 +1,112 @@
...
+        <Button
+          color="purple"
+          size="mini"
+          content="Request extension"
+          disabled={isDisabled}
+          onClick={this.handleExtendRequest}
+        />
+        {isDisabled && (
+          <Popup
+            content={this.infoMessage()}
```
ntarocco: here we evaluate 2 times the conditions. Not really a big issue, I am just wondering if there is a better solution, e.g. (not super elegant)
```
let result = true, msg = '';
!this.validateExtendUrl() ?
    result = false;
    msg = ExtendButton.INFO_MESSAGES['extendUrl'];
: !this.validateExtendAction() ?
    result = false;
    msg = ...
: ...
return {
    result: result,
    msg: msg
}
```

invenio_app_ils/permissions.py
```diff
@@ -79,6 +90,24 @@ def files_permission(obj, action=None):
...
+    """Return permission to allow only owner and librarians to extend loan."""
+    if not current_user or not current_user.id:
+        abort(401)
+
+    if current_user.id == int(loan["patron_pid"]):
+        loan = loan.replace_refs()
+        if loan.get("document").get("circulation").get("overbooked"):
+            raise InvalidLoanExtendError("This document is overbooked!")
+
+    return is_loan_owner_permission(loan["patron_pid"])
```
ntarocco:
```suggestion
    return LoanOwnerPermission(loan)
```

and remove `is_loan_owner_permission`

ntarocco (approved):

ntarocco (approved):

Issue #474, opened 2019-08-29
Series: show number of documents


ntarocco:
In the Series search results and details page, show the number of documents that the series has.

Both on the frontsite and in the backoffice
Consult the task before starting!

Request changes:
semanic-ui <Label> to highlight the number of volumes
revert the order of words: eg. 1 volume, 2 volumes
`online` `<Label>` when we have the `access_urls` field (not empty) on the series

PR #777, opened 2020-03-24
Loans: display number of available items

### LINKED ISSUE #711: UX-backoffice: loans search page, add number of available items to the loan result list
or at least if the document to which the loan refers to is overbooked or not (both)

it would be nice to have a facet there too (we can discuss about the facets before)

goal: easily see which loan requests are actually possible to handle

KonstantinaStoikou:
Closes #711 and very minor fix for #474 (volumes in frontsite).

Selection_001

### DIFF
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -1181,6 +1181,12 @@ RECORDS_REST_FACETS = dict(
             state=dict(terms=dict(field="state")),
             delivery=dict(terms=dict(field="delivery.method")),
             returns=overdue_agg,
+            availability=dict(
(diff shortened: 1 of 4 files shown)


kpsherva:
🚀 
in the series task (https://github.com/inveniosoftware/invenio-app-ils/issues/474) we suggested to put the online and volumes `<Label>` from semantic-ui-react instead of `<div>` to the `SeriesCard`, can you add it to this PR too?
and the screenshot as Nicola mentioned. Thanks!

kpsherva (approved):

ntarocco (approved):

PR #782, opened 2020-03-26
Fix access badge on documents and loans

### LINKED ISSUE #776: fix the access badge on the documents/loans/etc
on the patron profile view all the documents are appearing as restricted.
Problems here:
- [ ] restricted documents should not appear at all at the Patron's view
- [ ] not all the documents should be restricted, there is probably wrongly implemented if/changed field name or similar tiny issue

KonstantinaStoikou:
Fix the restricted access badge that appeared in all documents in loans (in patron's profile).

Before:
Selection_002

After:
Selection_001

Closes #776

### DIFF
--- a/ui/src/components/Document/DocumentItemCover.js
+++ b/ui/src/components/Document/DocumentItemCover.js
@@ -7,8 +7,8 @@ import _get from 'lodash/get';
 
 export default class DocumentItemCover extends Component {
   getLabel = metadata => {
-    const isAccessed = _get(metadata, 'open_access', false);
-    return isAccessed
+    const hasOpenAccess = _get(metadata, 'open_access', false);
+    return hasOpenAccess
       ? null
       : {
           corner: 'left',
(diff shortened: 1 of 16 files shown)


ui/src/components/Document/DocumentItemCover.js
```diff
@@ -6,8 +6,12 @@ import { Item } from 'semantic-ui-react';
 import isEmpty from 'lodash/isEmpty';
 
 export default class DocumentItemCover extends Component {
-  getRestrictions = meta => {
-    if (!isEmpty(meta) && meta.open_access) return null;
+  getRestrictions = document => {
+    if (
```
ntarocco: I am trying to remove this behaviour of accepting both `metadata` and `document.metadata`, I think we should fix it when used.
So you could you please:
1. define the propTypes
```
DocumentItemCover.propTypes = {
  metadata: PropTypes.object.isRequired,
};
```
2. fix any usage of `DocumentItemCover` and make sure we always pass the `metadata`, and never the `document`
3. change this code to just do:
```
getLabel = metadata => {
  const isRestricted = _get(metadata, 'open_access', true);
  return isRestricted ? { ... } : null;
}
```

ntarocco (approved):

PR #785, opened 2020-03-30
ui: backoffice refactor access labels and imports

### LINKED ISSUE #740: Document: change open_access field to public or restricted
* change the name of the field open_access
* pay attention that this was used for access restriction to record details and search results, for both `/api/documents` and `/api/literatures`

kpsherva:
* refactors the document field  from `open_access` -> `restricted`
* refactors imports from the backoffice components (it was causing circular imports in jest)
* closes #740


invenio_app_ils/search/permissions.py
```diff
@@ -30,27 +30,27 @@ def _get_user_provides():
...
+        # precedence over `restricted`.
         # if not `_access`, check if open access as before.
         _access_field_exists = Q("exists", field="_access.read")
         provides = _get_user_provides()
         user_can_read = _access_field_exists & Q(
             "terms", **{"_access.read": provides}
         )
         combined_filter = user_can_read | (
-            ~_access_field_exists & is_open_access
+            ~_access_field_exists & is_restricted
```
ntarocco:
```suggestion
            ~_access_field_exists & ~is_restricted
```
I think it should be that if the `_access` does not exist, then the doc should be **not** restricted. WDYT?

kpsherva: 👀 I fixed it. Thanks for spotting it!

ntarocco (approved):

PR #799, opened 2020-04-06
ui: acq fixes for ILL

ntarocco:
- prefixes orders/vendors to acq for better isolation and avoid
confusion with ILL
- fixes to schemas and mappings
- fixes to ChangedBy component to display createad_by and updated_by
- fixes CopyButton globally


ui/src/api/index.js
```diff
@@ -12,5 +12,8 @@ export { patron } from './patrons/patron';
 export { series } from './series/series';
 export { stats, circulationStats } from './stats';
 export { vocabulary } from './vocabularies/vocabulary';
-export { order, vendor } from './acquisition';
-export { borrowingRequest, library } from './ill';
+export { acqOrder, acqVendor } from './acquisition';
+export {
```
ntarocco, line 16: added this rename just to be able to split in 2 PRs. It is remove in the next PR.

ui/src/components/CopyButton/CopyButton.js
```diff
@@ -18,11 +18,64 @@ export class CopyButton extends React.Component {
...
+
+  onCopy = () => {
+    this.setState(() => ({
+      confirmationPopupIsOpen: true,
+      confirmationPopupMsg: 'Copied!',
+    }));
+    this.delayClosePopup();
+  };
+
+  delayClosePopup = () => {
```
kpsherva, line 40: could you tell what was the problem that is solved here?  just to be aware for the future

kpsherva (approved):

ui/src/components/CopyButton/CopyButton.js
ntarocco: True! So here how it works:
1. when you mouse over, it shows a popup saying "Copy to clipboard"
2. when you click, this popup disappears, and it shows new popup saying "Copied!"
3. this last one appears for a few seconds, then it disappears

PR #801, opened 2020-04-07
global: removed document.circulation from loan

### LINKED ISSUE #795: Remove circular dependency loan-document
From @ntarocco: adding a few notes here in case we have to come back again to this

**Implementation**

This feature was implemented in the following way:
1. In the `Loan` class, the `$ref` for the document was also resolving the field `circulation`
2. Such `circulation` performs, at the time of writing, a few searches on Elasticsearch, in particular a search on the `loans` index, which causes a circular reference and therefor the issues described below.

**Issues with such implementation**

The usage of `$ref` to resolve some related records might be a good solution but hides the risk of circular dependencies.
In this particular case, when adding the `document.circulation` field, what was happening is described in the following example scenario:

1. A specific Loan is changed, e.g. after a `checkout`, so it transitions from `state` with value `PENDING` to `ON_LOAN`
2. The Loan record is re-indexed
3. When re-indexing, the `$ref` are resolved, including the `document.circulation` field, which will perform a search on the `loans` index to retrieve PENDING, ACTIVE, etc... loans
4. Given that the current being indexed Loan is not yet indexed, the search will **not** include the change of this loan. This means that the count of pending/active loans will be wrong.

A search/filter/facets to `/api/circulation/loans` for any value in the field `loan.document.circulation` will therefor be wrong, while the response to the item endpoint `/api/circulation/loans/<pid>` will be correct, because the search performed by `document.circulation` is done in place (the loan was already indexed).

topless:
- removed document.circulation from loan resolver, closes #795 
- for loan extension, document overbook validation happens on backend only
- removed "physical copies available" from LoanListEntry
- removed from FrontSite/PatronDetails/LoanRequests "You can also read it on-site only"
- removed item availability filter for loans 
- split LoanList, LoanListEntry and LoanDates to separate files

### DIFF
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -1133,20 +1133,10 @@ RECORDS_REST_FACETS = dict(
             state=dict(terms=dict(field="state")),
             delivery=dict(terms=dict(field="delivery.method")),
             returns=overdue_agg,
-            availability=dict(
-                range=dict(
-                    field="document.circulation.has_items_for_loan",
-                    ranges=[{"key": "Available for loan", "from": 1}],
-                )
-            ),
         ),
         post_filters={
             "state": terms_filter("state"),
             "delivery": terms_filter("delivery.method"),
-            "availability": keyed_range_filter(
-                "document.circulation.has_items_for_loan",
-                {"Available for loan": {"gt": 0}},
-            ),
             "returns.end_date": overdue_loans_filter("end_date"),
             "loans_from_date": date_range_filter("start_date", "gte"),
             "loans_to_date": date_range_filter("start_date", "lte"),
--- a/invenio_app_ils/permissions.py
+++ b/invenio_app_ils/permissions.py
@@ -96,12 +96,21 @@ def loan_extend_circulation_permission(loan):
         abort(401)
 
     if current_user.id == int(loan["patron_pid"]):
-        loan = loan.replace_refs()
-        if loan.get("document").get("circulation").get("overbooked"):
-            raise InvalidLoanExtendError("This document is overbooked!")
-
+        Document = current_app_ils.document_record_cls
+        document_rec = Document.get_record_by_pid(loan["document_pid"])
+        document = document_rec.replace_refs()
+        is_overbooked = document.get("circulation", {}).get("overbooked")
+        if is_overbooked == None:
+            # NOTE: this should never happen, if it happens it means that the
+            # document does not have `circulation.overbooked` field. Fix it!
+            abort(500)
+        elif is_overbooked == True:
+            raise InvalidLoanExtendError(
+                "The extension cannot be automatically accepted due to high"
+                "demand for this literature. Please contact the library to"
+                "request a loan extension."
+            )
     return LoanOwnerPermission(loan)
-
 
 class LoanOwnerPermission(Permission):
     """Return Permission to evaluate if the current user owns the loan."""
--- a/ui/src/pages/backoffice/Loan/LoanSearch/LoanDates.js
+++ b/ui/src/pages/backoffice/Loan/LoanSearch/LoanDates.js
@@ -1,0 +1,47 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { invenioConfig } from '@config';
+import { Icon, List } from 'semantic-ui-react';
+
+export class LoanDates extends Component {
+  render() {
+    const { loan } = this.props;
+    const isRequest = invenioConfig.circulation.loanRequestStates.includes(
+      loan.metadata.state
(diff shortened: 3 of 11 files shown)


ui/src/pages/backoffice/Loan/LoanSearch/LoanDates.js
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react';
+import { invenioConfig } from '@config';
+import { Icon, List } from 'semantic-ui-react';
+
+export class LoanDates extends Component {
```
ntarocco, line 6: @topless for new components, may I ask that we always include all the PropTypes and defaults?
I would also pass, as we discussed, the minim amount of info necessary, maybe in this case only the `metadata` object.
I found that it was really helpful while refactoring. I will pay attention to this also in my code (need to re-check my PRs)

ui/src/pages/backoffice/Loan/LoanSearch/LoanListEntry.js
```diff
@@ -0,0 +1,116 @@
...
+import { BackOfficeRoutes } from '@routes/urls';
+import _isEmpty from 'lodash/isEmpty';
+import React, { Component } from 'react';
+import { Link } from 'react-router-dom';
+import { Grid, Header, Item, Label, List } from 'semantic-ui-react';
+import { LoanLinkToItem } from '../../components/Loan';
+import { OverdueLoanSendMailModal } from '../../components/OverdueLoanSendMailModal';
+import { LoanDates } from './LoanDates';
+
+export class LoanListEntry extends Component {
```
ntarocco, line 13: Same here

ui/src/pages/backoffice/Loan/LoanSearch/LoanDates.js
topless: @ntarocco we can start filling them up, those who miss. On top of that instead of using the generic and without any value object, maybe we could use something like this https://github.com/dowjones/react-json-schema-proptypes so we start working with proper propTypes.shapes and "stronger" type safety.

ntarocco: interesting idea, but it will require some changes in our base code. I think that what we are missing is just more strict prop types (e.g. if you pass something not defined in the prop types, it should complain), which is discussed but not yet addressed [here](libraryMetadata).

invenio_app_ils/config.py
```diff
@@ -1180,20 +1180,10 @@ def _(x):
             state=dict(terms=dict(field="state")),
             delivery=dict(terms=dict(field="delivery.method")),
             returns=overdue_agg,
-            availability=dict(
```
kpsherva, line 1136: @topless @ntarocco  this was one of the requirements, it made a lot of sense to have this filter, so I think we should discuss about alternative solutions

ntarocco: ok, we can brainstorm this because I don't have ideas at the moment

invenio_app_ils/permissions.py
```diff
@@ -96,12 +96,23 @@ def loan_extend_circulation_permission(loan):
...
+                "The extension cannot be automatically accepted due to high"
+                "demand for this literature. Please contact the library to"
+                "request a loan extension."
+            )
+        elif is_overbooked == False:
+            return LoanOwnerPermission(loan)
+        else:
+            # NOTE: this should never happen, if it happens it means that the
+            # document does not have `circulation.overbooked` field. Fix it!
+            abort(500)
```
ntarocco: missing the final return, I would change it maybe to:
```
is_overbooked = document.get("circulation", {}).get("overbooked")
if is_overbooked == None:
    # NOTE: this should never happen, if it happens it means that the
    # document does not have `circulation.overbooked` field. Fix it!
    abort(500)
elif is_overbooked == True:
    raise InvalidLoanExtendError(
        "The extension cannot be automatically accepted due to high"
        "demand for this literature. Please contact the library to"
        "request a loan extension."
    )
return LoanOwnerPermission(loan)
```

invenio_app_ils/config.py
topless: @kprzerwa I tried to come up with an alternative so we can keep the feature, but I failed. When I was reviewing the changes, this `availability` filter, I found it hard to understand what was the purpose of it. But since its a requirement we have to work around it.

kpsherva: yes I know it is tricky. I created an issue for this to discuss and tackle later. https://github.com/inveniosoftware/invenio-app-ils/issues/804

ntarocco (approved):

kpsherva (approved):

PR #807, opened 2020-04-09
ui: configurable static pages

### LINKED ISSUE #772: Static pages
ILS should come by default with a set of static pages similar to what users can find in a website. The default content can be something fake because it must be customizable.

Pages to create by default in ILS:
- About Us
- Contacts 

Each page should have simply a title and a content. Static pages links should be displayed in the footer.

What should be customizable (already supplied by invenio-pages:
- the title
- the content
- the name of the link (can differ from the title)
- the URL, if possible

The overlay that overrides ILS should be able to map the routing from pages in react to pages in /api
```
const staticPages = {
  { name: 'about', route: '/about', apiURL: '1' }
]}
```
The overriding is possible by changing this variable in the overlay

Tasks
---

1. Add to `react-router` the possibility of defining routes by config (this `link.urlPath`)
2. Implement the generic static page cmp

Decisions
----------
- [ ] decided not to overwrite the individual pages as react components - every time changing the content of such a page would require a deployment of react app. Instead the pages will be fetched from invenio-pages, where static page can be easily edited, with a simple formatting editor included
- [ ] overridable component will be the footer of the frontsite -  the footer is for now injectable by a `renderElement` function, until the strategy for components is established
- [ ] custom pages (from any overlay) will be added as a param to the routing

kpsherva:
* closes #772

* footer is fully overridable, IMO its very custom so it does not make sense to override parts of it - it is overridable in the same way as react-searchkit

Decisions taken
----------
1. decided not to overwrite the individual pages as react components - every time changing the content of such a page would require a deployment of react app. Instead the pages will be fetched from invenio-pages, where static page can be easily edited, with a simple formatting editor included
2. overridable component will be the footer of the frontsite
3. custom pages (from any overlay) will be added as a param to the routing 
4. override was implemented like in the react-searchkit. I think we should discuss in detail the id approach proposed by indico first.

Loading:
Screenshot 2020-04-09 at 14 33 18
Loaded:
Screenshot 2020-04-09 at 14 33 31

### DIFF
--- a/invenio_app_ils/cli.py
+++ b/invenio_app_ils/cli.py
@@ -17,6 +17,7 @@ from random import randint
 import arrow
 import click
 import lorem
+import pkg_resources
 from flask import current_app
 from flask.cli import with_appcontext
 from invenio_accounts.models import User
@@ -24,6 +25,7 @@ from invenio_circulation.api import Loan
 from invenio_circulation.pidstore.pids import CIRCULATION_LOAN_PID_TYPE
 from invenio_db import db
 from invenio_indexer.api import RecordIndexer
+from invenio_pages import Page
 from invenio_pidstore.models import PersistentIdentifier, PIDStatus
 from invenio_pidstore.providers.recordid_v2 import RecordIdProviderV2
 from invenio_search import current_search
@@ -973,7 +975,6 @@ class OrderGenerator(Generator):
         db.session.commit()
         return recs
 
-
 @click.group()
 def demo():
     """Demo data CLI."""
@@ -1242,6 +1243,39 @@ def create_userprofile_for(email, username, full_name):
         db.session.commit()
 
 
+@click.group()
+def fixtures():
+    """Create initial data and demo records."""
+
+
+@fixtures.command()
+@with_appcontext
+def pages():
+    """Register CDS static pages."""
+    def page_data(page):
+        return pkg_resources.resource_stream(
+            'invenio_app_ils', os.path.join('templates/static_pages', page)
+        ).read().decode('utf8')
+
+    pages = [
+        Page(url='/about',
+             title='About',
+             description='About',
+             content=page_data('about.html'),
+             template_name='invenio_pages/default.html'),
+        Page(url='/contact',
+             title='Contact',
+             description='Contact',
+             content=page_data('contact.html'),
+             template_name='invenio_pages/default.html'),
+    ]
+    with db.session.begin_nested():
+        Page.query.delete()
+        db.session.add_all(pages)
+    db.session.commit()
+    click.echo('static pages created :)')
+
+
 @click.command()
 @click.option("--recreate-db", is_flag=True, help="Recreating DB.")
 @click.option(
@@ -1354,5 +1388,7 @@ def setup(recreate_db, skip_demo_data, skip_file_location, skip_patrons,
     # Generate demo data
     if not skip_demo_data:
         run_command("demo data")
+
+    run_command("fixtures pages")
 
     click.secho("ils setup finished successfully", fg="blue")
--- a/invenio_app_ils/templates/static_pages/contact.html
+++ b/invenio_app_ils/templates/static_pages/contact.html
@@ -1,0 +1,2 @@
+You can contact InvenioILS developers on
+<a href="https://gitter.im/inveniosoftware/invenio">our chatroom</a>
--- a/setup.py
+++ b/setup.py
@@ -113,6 +113,7 @@ install_requires = [
     # --- extra deps of ILS -----------------------------------------------
     "invenio-circulation>=1.0.0a21,<1.1.0",
     "invenio-stats>=1.0.0a17",
+    "invenio-pages>=1.0.0a5",
     "invenio-pidrelations>=1.0.0a6,<1.1.0",
     "invenio-opendefinition>=1.0.0a9,<1.1.0",
     "sentry-sdk>=0.10.2",
@@ -154,6 +155,7 @@ setup(
             "setup = invenio_app_ils.cli:setup",
             'stats = invenio_stats.cli:stats',
             "vocabulary = invenio_app_ils.vocabularies.cli:vocabulary",
+            "fixtures = invenio_app_ils.cli:fixtures",
         ],
         "invenio_base.apps": [
             "ils_ui = invenio_app_ils.ext:InvenioAppIlsUI",
--- a/ui/src/api/staticPages/page.js
+++ b/ui/src/api/staticPages/page.js
@@ -1,0 +1,20 @@
+import { http } from '../base';
+
+const pageURL = '/pages/';
+
+/**
+ * GET the static page by providing ID
+ *
+ * page is not a record type (read more in https://github.com/inveniosoftware/invenio-pages)
+ *
+ * @param pageID id of the static page
+ * @returns {Promise<AxiosResponse<T>>} page object
+ */
+const get = async pageID => {
+  return http.get(`${pageURL}${pageID}`);
+};
+
+export const page = {
+  get: get,
+  url: pageURL,
+};
--- a/ui/src/components/ILSPlaceholder/ILSPlaceholder.js
+++ b/ui/src/components/ILSPlaceholder/ILSPlaceholder.js
@@ -1,14 +1,16 @@
 import React, { Component } from 'react';
 import { Item, Placeholder } from 'semantic-ui-react';
+import {random} from 'lodash/number';
 
 export class ILSParagraphPlaceholder extends Component {
   renderLines = () => {
     const lines = [];
+    const lineLengths = ['full', 'very long', 'long', 'medium', 'short', 'very short'];
     for (let i = 0; i < this.props.linesNumber; i++) {
       lines.push(
         <Placeholder.Line
           key={i}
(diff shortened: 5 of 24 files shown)


ui/src/components/ILSPlaceholder/ILSPlaceholder.js
```diff
@@ -1,14 +1,16 @@
 import React, { Component } from 'react';
 import { Item, Placeholder } from 'semantic-ui-react';
+import {random} from 'lodash/number';
 
 export class ILSParagraphPlaceholder extends Component {
   renderLines = () => {
     const lines = [];
+    const lineLengths = ['full', 'very long', 'long', 'medium', 'short', 'very short'];
```
kpsherva, line 8: This is because semantic-ui is not very good with producing random length lines

ui/src/pages/frontsite/StaticPage/StaticPage.js
```diff
@@ -0,0 +1,50 @@
...
+    const staticPage = getStaticPageByRoute(this.props.match.path);
+    const staticPageID = staticPage['apiURL'];
+    this.props.fetchStaticPageDetails(staticPageID);
+  }
+
+  parseStaticPageContent = () => {
+    const { data } = this.props;
+    if (!_isEmpty(data)) {
+      // unescapes HTML and strips unsafe tags
+      const parsed = new DOMParser().parseFromString(data.content, 'text/html');
```
kpsherva: please verify this, I found it as a method which does not allow for XSS

ui/src/routes/components/BackOffice/BackOfficeRoutesSwitch.js
```diff
@@ -248,6 +249,9 @@ export default class extends Component {
           component={BorrowingRequestDetails}
         />
         <Route exact path={BackOfficeRoutes.stats.home} component={Stats} />
+        <Route>
```
kpsherva, line 252: 404 error was not visible before because of how we structured the routes

ui/src/api/staticPages/page.js
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * GET the static page by providing ID
+ *
+ * page is not a record type (read more in https://github.com/inveniosoftware/invenio-pages)
+ *
+ * @param pageID id of the static page
+ * @returns {Promise<AxiosResponse<T>>} page object
+ */
+const get = async pageID => {
+  return await http.get(`${pageURL}${pageID}`);
```
zzacharo: No need for `await` here :)

ui/src/pages/frontsite/StaticPage/StaticPage.js
```diff
@@ -0,0 +1,50 @@
...
+      return ReactHtmlParser(parsed.body.innerText);
```
zzacharo: Isn't this enough https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml instead of having another dependency?

ntarocco:
General comment: Zac was asking me the reasons behind adding `invenio-pages`, maybe we should document it (copy/paste from the discussion):
1. normally you would define the static pages before going live, and most probably you won't change much (adding/removing). It can happen, but not frequently.
2. the content of a page can instead change more frequently: think to a contact pages where there are opening hours/days off that might change over the year, or a FAQ that you are gonna enrich from time to time

kpsherva:
> General comment: Zac was asking me the reasons behind adding `invenio-pages`, maybe we should document it (copy/paste from the discussion):
> 
> 1. normally you would define the static pages before going live, and most probably you won't change much (adding/removing). It can happen, but not frequently.
> 2. the content of a page can instead change more frequently: think to a contact pages where there are opening hours/days off that might change over the year, or a FAQ that you are gonna enrich from time to time

I have updated the description of the PR so the motivation for choosing the invenio-pages would be now written somewhere

invenio_app_ils/templates/static_pages/contact.html
```diff
@@ -0,0 +1,2 @@
+You can contact Invenio-ILS developers on
```
ntarocco:
```suggestion
You can contact InvenioILS developers on
```

invenio_app_ils/cli.py
```diff
@@ -1242,6 +1243,39 @@ def create_userprofile_for(email, username, full_name):
...
+             title='Contact',
+             description='Contact',
+             content=page_data('contact.html'),
+             template_name='invenio_pages/default.html'),
+    ]
+    with db.session.begin_nested():
+        Page.query.delete()
+        db.session.add_all(pages)
+    db.session.commit()
+    click.echo('static pages created:)')
```
ntarocco:
```suggestion
    click.echo('static pages created :)')
```
and if you use Black will probably change all your quotes `'` -> `"`

ui/src/pages/backoffice/components/Sidebar/AdminMenu.js
```diff
@@ -0,0 +1,22 @@
...
+import { Link } from 'react-router-dom';
+import { Divider, Menu } from 'semantic-ui-react';
+
+export default class AdminMenu extends Component {
+  render() {
+    return (
+      <>
+        <Divider horizontal>Admin menu</Divider>
+        <Menu text vertical className="bo-menu">
+          <Menu.Item as={'a'} href="https://localhost:5000/admin">
```
ntarocco: We cannot hardcode `localhost` here, I would add a new route in `urls.js`

ui/src/pages/backoffice/components/Sidebar/AdminMenu.js
```diff
@@ -0,0 +1,22 @@
+import { FrontSiteRoutes } from '@routes/urls';
```
ntarocco: nice thanks!

ui/src/pages/frontsite/StaticPage/StaticPage.js
ntarocco: are you sure we need all this? if the `data.content` contains some HTML, it should just render when you add it to the DOM. What it is happening here?

ui/src/pages/backoffice/components/Sidebar/AdminMenu.js
zzacharo: I would suggest to create a `baseUrl` similar to [this](https://github.com/inveniosoftware/invenio-app-ils/blob/master/ui/src/api/base.js#L5)

ntarocco: > I would suggest to create a `baseUrl` similar to [this](https://github.com/inveniosoftware/invenio-app-ils/blob/master/ui/src/api/base.js#L5)

Right, better, because of the 5000 port, otherwise it won't be needed...
Probably refactor the existing one to expose the base url and the API url.

ui/src/pages/frontsite/StaticPage/state/actions.js
```diff
@@ -0,0 +1,25 @@
...
+import { page as pageApi } from '@api';
+
+export const fetchStaticPageDetails = pageID => {
+  return async dispatch => {
+    dispatch({
+      type: IS_LOADING,
+    });
+
+    await pageApi
+      .get(pageID)
```
ntarocco: can you use `await` everywhere instead of Promise `then`?

ui/src/api/base.js
```diff
@@ -10,6 +10,13 @@ const apiConfig = {
   withCredentials: true,
 };
 
+const adminConfig = {
```
ntarocco: I would have gone with something like:
```
const invenioBaseURL = process.env.NODE_ENV === 'production' ? '/' : `${process.env.REACT_APP_BACKEND_DEV_BASE_URL}/
const apiConfig = {
  baseURL: invenioBaseURL,
  withCredentials: true,
}
```
to avoid duplication, and then append the `/admin` in the route file, but I am also fine with this :)

setup.py
```diff
@@ -105,6 +105,7 @@
     # --- `metadata` bundle without records UI -----------------------------
     "invenio-indexer>=1.1.0,<1.2.0",
     "invenio-jsonschemas>=1.0.0,<1.1.0",
+    "invenio-pages>=1.0.0a5",
```
ntarocco: very very very minor, for another time: can you maybe move this package a bit below under
`--- extra deps of ILS -----------------------------------------------`?
thanks!

ntarocco (approved):

ui/src/pages/frontsite/StaticPage/StaticPage.js
```diff
@@ -0,0 +1,49 @@
...
+import { getStaticPageByRoute } from '@config/uiConfig';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Container, Header } from 'semantic-ui-react';
+import _isEmpty from 'lodash/isEmpty';
+
+export default class StaticPage extends Component {
+  componentDidMount() {
+    const staticPage = getStaticPageByRoute(this.props.match.path);
+    const staticPageID = staticPage['apiURL'];
```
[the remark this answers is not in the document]
kpsherva: if it is not configured it will go to 404, because of how I wrote the routing based on the config. If there will be a route in `match.path` that is not coming from the static pages configuration, it will not even arrive to this component. Checking all the paths against the config to throw the error seems to me a bit complicated in comparison to what we gain, unless you have an idea ?

PR #808, opened 2020-04-09
Fix ui forms

### LINKED ISSUE #752: fix the submit button behaviour for forms
Problem: Now it is possible to submit twice, there is no feedback about the submit which is ongoing.

Expected:
When clicking on submit, the button should be in the loading state, the whole form should be blocked against making any changes (f.e. also loading state) before the response comes back, and the users get's redirected.

ntarocco:
- removes `ES_DELAY` when saving forms
- replaces the various `setTimeout` with common `delay` function that returns a Promise
- replaces `window.open` with `goTo` where it makes sense
- refactors a few Promises to use async/await instead
- closes #752

There a still a couple of setTimeout left and weird behaviour on submit, in the components in Patron Profile pages: I couldn't change it easily because they require some refactoring.
We will do them later on.

### DIFF
--- a/ui/src/forms/core/BaseForm/BaseForm.js
+++ b/ui/src/forms/core/BaseForm/BaseForm.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import { Formik, getIn } from 'formik';
 import isEmpty from 'lodash/isEmpty';
 import { Form, Button, Header, Container } from 'semantic-ui-react';
-import { ES_DELAY } from '@config';
 import { ErrorMessage } from '../ErrorMessage';
 
 export class BaseForm extends Component {
@@ -24,15 +23,13 @@ export class BaseForm extends Component {
         ? await this.props.editApiMethod(this.props.pid, submitValues)
         : await this.props.createApiMethod(submitValues);
 
-      setTimeout(() => {
-        this.props.sendSuccessNotification(
-          'Success!',
-          this.props.successSubmitMessage
-        );
-        if (this.props.successCallback) {
-          this.props.successCallback(response, submitButton);
-        }
-      }, ES_DELAY);
+      this.props.sendSuccessNotification(
+        'Success!',
+        this.props.successSubmitMessage
+      );
+      if (this.props.successCallback) {
+        this.props.successCallback(response, submitButton);
+      }
     } catch (error) {
       const errors = getIn(error, 'response.data.errors', []);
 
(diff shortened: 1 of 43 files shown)


kpsherva (approved):

ui/src/pages/backoffice/Location/LocationForms/InternalLocationEditor/components/InternalLocationForm/InternalLocationForm.js
```diff
@@ -28,12 +28,16 @@ export class InternalLocationForm extends Component {
     ]);
   };
 
-  updateInternalLocation = (pid, data) => {
-    return internalLocationApi.update(pid, data);
+  updateInternalLocation = async (pid, data) => {
+    const response = await internalLocationApi.update(pid, data);
+    await delay();
```
topless, line 33: While here we inject the delay in the component, in the `internalLocation.delete` the delay is part of the actions which I think is the place we should put all the delays in order to avoid the pitfall of having them twice (not sure if this is the case anywhere).

topless (approved):

ui/src/forms/core/BaseForm/BaseForm.js
```diff
@@ -24,15 +23,13 @@ export class BaseForm extends Component {
...
-        this.props.sendSuccessNotification(
-          'Success!',
-          this.props.successSubmitMessage
-        );
-        if (this.props.successCallback) {
-          this.props.successCallback(response, submitButton);
-        }
-      }, ES_DELAY);
+      this.props.sendSuccessNotification(
+        'Success!',
```
zzacharo, line 27: nit: I like this refactoring here and I would say that even the "Success" argument should be handled from the function `sendSuccessNotification`. It was like that before as it was required from the action for the notification part. But I think is clearer if the function just accepts a function. Also, in that case I would say the `successSubmitMessage` is redundant.

zzacharo (approved):

ui/src/pages/backoffice/Location/LocationForms/InternalLocationEditor/components/InternalLocationForm/InternalLocationForm.js
topless: I checked the code and there are reducers only for `locationList` and `internalLocationList` so, the components is the only place.

ntarocco: Discussed by voice.

ui/src/forms/core/BaseForm/BaseForm.js
ntarocco: I agree. Honestly I started to do it, but then I saw it was touching all forms, so better to do it in another PR if it is ok with you.

PR #813, opened 2020-04-15
560 book covers

### LINKED ISSUE #560: Book cover provider
## Problem

Covers for our literature are provided by a configurable third party service. ILS should provide an interface for customisation. 

## User Stories 

There are two user profiles for our stories, the developer `dev`, `librarian` and `user`

- [x] As `dev`, I need to configure my book cover provider or repository of covers.
- [x] As `dev`, I should be able to get different size of covers.
- [x] As `dev` & `lib`, I need to be able to override the `No Cover`.
- [x] As a `user` I should be able to see the medium size image or `No Cover` in front-site suggestions `Popular topics`, `Most Loaned` etc ..
- [x] As `user` I should be able to see the large size image or `No Cover` for document details.
- [x] As `user` I should be able to see the medium size image for or `No Cover` for document search grid view.
- [x] As `user` I should be able to see the small size image for or `No Cover` for document search list view.
- [x] As `user` I should experience the same cases in mobile too.
- ~[ ] As a `librarian` if there are multiple covers available, I want to choose which one to use.~

## Other
- [x] `cover_metadata` for literature
- Order of the ISBN's matters. When an item gets associated/dissociated with document a cron jobs checks/updates the order of ISBNs. There are ISBNs in `document.identifiers` and in `document.items`. We have to check them in an order from newer to older and the first one which returns a valid cover should be placed first in the list. Added to CERNDocumentServer/cds-books#92
- [x] Journals have also covers, replace the `isbn` with `issn` url parameter.

related CernDocumentServer/cds-books#92

topless:
- cli: added ISBN to items
- config: variable to override book cover creation method
- document: serializer to use the configured covers variable
- serializer for literature
- seriliazer for loan resolved document
- series: if ISSN identifier available is used for cover
- fix: broken frontsite patron details covers
- fix: DocumentItemCover missing param
- ui: replaced all DocumentCover occurencies except Series
  - DocumentDetails (backoffice)
  - DocumentRelations/DocumentCard (backoffice)
  - BookCard (frontsite)
  - DocumentPanel (frontsite)
  - DocumentPanelMobile (frontsite)
- related CERNDocumentServer/cds-books#92
- closes #560

Screenshot 2020-04-24 03 24 47


invenio_app_ils/documents/serializers/document.py
```diff
@@ -0,0 +1,44 @@
...
+            record_hit,
+            links_factory=links_factory,
+            **kwargs
+        )
+        hit = self.build_cover_urls(hit)
+        return hit
+
+    def build_cover_urls(self, document):
+        """Build cover urls for document."""
+        url_builder = current_app.config.get("ILS_BUILD_LITERATURE_COVER_URLS")
```
ntarocco: minor: IMHO, I like the `builder` name that you used!
```suggestion
        url_builder = current_app.config.get("ILS_LITERATURE_COVER_URLS_BUILDER")
```

invenio_app_ils/documents/utils.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
```
ntarocco: why not naming this files `covers_builder` or something instead of utils?
And I think it should not be inside `Document`, because it should be used for Series and Literature too.

ui/src/components/Document/utils.js
```diff
@@ -1,8 +1,6 @@
-// NOTE: random function to get different covers for books.
 
+// NOTE: random function to get different covers for books.
 export function getCover(edition = '42') {
   const imageId = 46276847 + parseInt(edition);
   return `https://assets.thalia.media/img/${imageId}-00-00.jpg`;
```
ntarocco: why not moving this to backend instead? Just double check if `thalia.media` can be used freely.

invenio_app_ils/config.py
```diff
@@ -374,12 +373,12 @@ def _(x):
...
         record_serializers={
             "application/json": (
-                "invenio_app_ils.records.serializers:json_v1_response"
+                "invenio_app_ils.documents.serializers:document_v1_response"
             )
         },
         search_serializers={
             "application/json": (
-                "invenio_app_ils.records.serializers:json_v1_search"
+                "invenio_app_ils.documents.serializers:document_v1_search"
```
ntarocco: I think the same should be applied to `series` and `literatures`, WDYT?

invenio_app_ils/documents/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    default_url = "/images/placeholder.png"
+    document["metadata"]["cover_urls"] = {
+        "small": "{0}".format(default_url),
+        "medium": "{0}".format(default_url),
+        "large": "{0}".format(default_url),
+    }
+    return document
+
+
+def build_literature_cover_urls(document):
```
ntarocco: I think this method should be deleted, or partially moved to the CDS overlay.

ntarocco (commented):
LGTM! The main comment is that you have 3 records with covers:
- Documents (ISBN and Item)
- Series (ISSN or maybe other fields)
- Literatures (not a record, but it searches in Documents and Series, so you have to pay attention in the serializer)
So the cover builder should be generic enough for all of them, or maybe better having different builders?

ui/src/components/Document/utils.js
topless: This will be removed completely once I address couple of issues with BookCard

invenio_app_ils/documents/utils.py
topless: Yes, it will, this is my testing for syndetic and for the config override.

invenio_app_ils/records/serializers/__init__.py
```diff
@@ -55,3 +57,12 @@
 
 loan_v1_search = search_responsify(loan_v1, 'application/json')
 """JSON response builder decorates loan."""
+
+literature_v1 = LiteratureJSONSerializer(ILSRecordSchemaJSONV1, replace_refs=True)
```
ntarocco: Please move this in the `literature` module, check how it has been done in [my PR](https://github.com/inveniosoftware/invenio-app-ils/pull/820) or rebase from there since it will be merged soon.

invenio_app_ils/circulation/jsonresolvers/loan.py
```diff
@@ -73,6 +73,7 @@ def document_resolver(loan_pid):
         obj = pick(
             document,
             "authors",
+            "cover_urls",
```
ntarocco: Given the implementation below, you are missing the `identifiers` field here.

invenio_app_ils/records/serializers/loan.py
```diff
@@ -38,8 +39,29 @@ def transform_search_hit(
...
 
+    def document_cover_urls(self, metadata):
+        """Build cover ulrs for the related document."""
+        url_builder = current_app.config.get(
+            "ILS_LITERATURE_COVER_URLS_BUILDER")
+
+        # NOTE: we need the actual document here, bacause covers depend on,
+        # items or indentifiers.
+
+        Document = current_app_ils.document_record_cls
```
ntarocco: This cannot be done like this because you are performing a db query for each search result: this will drastically slow down searches. You can use only the available metadata.
You can add the `identifiers` in the json resolver, as commented above, and use that.

ui/src/pages/backoffice/components/Document/DocumentList/DocumentListEntry.js
```diff
@@ -100,8 +100,8 @@ export default class DocumentListEntry extends Component {
         <div className={'item-image-wrapper'}>
           <DocumentItemCover
             linkTo={BackOfficeRoutes.documentDetailsFor(document.metadata.pid)}
-            metadata={document.metadata}
-            coverUrl={document.metadata.edition}
+            isRestricted={_get(document, 'metadata.restricted', false)}
+            coverUrl={_get(document, 'metadata.cover_urls.large')}
```
ntarocco: in the list of results, shouldn't we use a smaller version and not the `large`?

ui/src/pages/frontsite/Documents/DocumentsDetails/DocumentPanel/DocumentPanel.js
```diff
@@ -24,7 +22,12 @@ export default class DocumentPanel extends Component {
             <Grid columns={3}>
               <Grid.Row>
                 <Grid.Column>
-                  <DocumentCover document={doc} />
+                  <DocumentCover
+                    coverUrl={
+                      isLoading ? '' : _get(doc, 'metadata.cover_urls.large')
```
ntarocco: I would rather implement this if inside the `DocumentCover`: the component should not display anything or a default image when it does not have the cover url.

ui/src/pages/frontsite/Documents/DocumentsSearch/DocumentListEntry/DocumentListEntry.js
```diff
@@ -69,8 +70,8 @@ export default class DocumentListEntry extends Component {
     const image = (
       <DocumentItemCover
         size="mini"
-        coverUrl={this.metadata.edition}
-        metadata={this.metadata}
+        coverUrl={_get(this, 'metadata.cover_urls.large')}
```
ntarocco: same comment here for the size of the image in the list of results

ntarocco (commented):
Frontend looks good to me! For the backend part, the main issue is that we cannot have db calls in the serializer for the list of results, otherwise it will dramatically slow down searches. We experienced this issue in CDS Videos.

invenio_app_ils/circulation/jsonresolvers/loan.py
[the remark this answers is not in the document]
topless: This is not needed, its a leftover.

invenio_app_ils/records/serializers/__init__.py
topless: I left it here on purpose because was expecting your pr to be merged first and then rebase it on top. Will move it in literature as you suggested.

invenio_app_ils/config.py
```diff
@@ -373,14 +372,14 @@ def _(x):
         },
         record_serializers={
             "application/json": (
-                "invenio_app_ils.records.serializers:json_v1_response"
+                "invenio_app_ils.literature.serializers:json_v1_response"
```
ntarocco, line 375: This works for sure, but I am wondering if we should split in different serializers at some point, even if for now the code is the same.

invenio_app_ils/documents/covers_builder.py
```diff
@@ -0,0 +1,30 @@
...
+#
+# Copyright (C) 2018-2020 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Document configuration callbacks."""
+
+
+def build_default_cover_urls(metadata):
```
ntarocco: Dead method? to be removed?

invenio_app_ils/documents/covers_builder.py
```diff
@@ -0,0 +1,30 @@
...
+    """Build default ulrs for literature, documents & series."""
+    default_url = "/images/placeholder-portrait.png"
+    return {
+        "small": "{0}".format(default_url),
+        "medium": "{0}".format(default_url),
+        "large": "{0}".format(default_url),
+    }
+
+
+def build_ils_demo_cover_urls(metadata):
```
ntarocco, line 11: 👍 
Only one small remark: since you are re-using this for both `Documents` and `Series`, you might want to use the `ISBN` for `Documents` but the `ISSN` for `Series`. WDYT?
This has to be applied then in CDS Books with Syndetics.

invenio_app_ils/documents/loaders/jsonschemas/document.py
```diff
@@ -197,6 +197,7 @@ class Meta:
     authors = fields.List(fields.Nested(AuthorSchema), required=True)
     conference_info = fields.Nested(ConferenceInfoSchema)
     copyrights = fields.List(fields.Nested(CopyrightsSchema))
+    cover_metadata = fields.Dict()
```
ntarocco, line 202: I am thinking now that if you go to the editor, and update a field, e.g. the title, this will be empty and override any previous value... to be tested!
I think you should "preserve" the field, like if it was a system field, as done for the `created_by` and `updated_by`.
Here a small refactoring of the existing `@pre_load set_changed_by`:
```
    @pre_load
    def preload_fields(self, data, **kwargs):
        """Automatically inject system fields."""
        prev_record = self.context.get("record")
        data.update(set_changed_by(data, prev_record)))
        data.update(preserve_cover_metadata(data, prev_record))
        return data
```

Be aware that you have to do the same in the `series` loader.

invenio_app_ils/cli.py
```diff
@@ -296,6 +296,35 @@ class DocumentGenerator(Generator):
         {"date": "2019-08-02", "place": "Geneva", "publisher": "CERN"},
         {"date": "2017-08-02", "place": "Hamburg", "publisher": "Springer"},
     ]
+    ISBNS = [
```
ntarocco, line 299: Fantastic! Can you copy paste the same for the ISSN for the series to show a cover?

setup.py
```diff
@@ -122,6 +122,7 @@
     # needed to have namedtuple json serialized as dict
     "simplejson>=3",
     "webargs>=5.5.2,<6.0",
+    "wtforms==2.2.1",
```
ntarocco: Just be aware that you will have a conflict with Karolina's PR

ntarocco:
I have noticed another small thing that maybe you can fix, while changing book, for example clicking on a related language, the previous book cover is still displayed and then the new one appears.

animated

I think that on component mount you should remove the img and put the place holder.

invenio_app_ils/documents/covers_builder.py
kpsherva: ISBN is for some multipart monographs, ISSN mostly for journals

kpsherva (approved):

invenio_app_ils/records/loaders/schemas/preserve_cover_metadata.py
```diff
@@ -0,0 +1,19 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Preserve cover metadata."""
+
+from flask import has_request_context
+
+
+def preserve_cover_metadata(data, prev_record):
+    """Preserver cover metadata if they exist."""
+    if not has_request_context() or prev_record is None:
```
ntarocco, line 15:
```suggestion
    if prev_record is None:
```
You don't need the `has_request_context` because you are not using anything from current request

ntarocco (approved):

PR #816, opened 2020-04-17
ui: removed redundant refs

kpsherva:


ntarocco (approved):

ui/src/pages/backoffice/Acquisition/Order/OrderDetails/OrderDetails.js
```diff
@@ -129,25 +122,19 @@ class OrderPanels extends React.Component {
           Order information
         </Header>
         <Segment attached className="bo-metadata-segment">
-          <div ref={this.orderTopRef} id="order-info">
-            <OrderInformation order={data.metadata} />
-          </div>
+          <OrderInformation order={data.metadata} />
```
topless, line 125: Do we need the complete metadata in these components? Can we pass a smaller payload?

ui/src/pages/backoffice/Acquisition/Order/OrderDetails/OrderInformation.js
```diff
@@ -31,7 +31,7 @@ export class OrderInformation extends React.Component {
     order.status === 'CANCELLED' &&
       rightTable.push({ name: 'Cancel reason', value: order.cancel_reason });
     return (
-      <Grid columns={2}>
+      <Grid columns={2} id="order-info">
```
topless, line 34: For all the following occurrences, is there a particular reason with went with `id` instead of `class`?

topless (approved):

ui/src/pages/backoffice/Acquisition/Order/OrderDetails/OrderInformation.js
kpsherva: yes because it is for the scrolling menu and we would like to have there unique id, so if you click on payment info it does not jump to vendor randomly

ui/src/pages/backoffice/Acquisition/Order/OrderDetails/OrderDetails.js
kpsherva: It is a bit not a scope of this PR
but I would rather pass the whole metadata and leave choosing the specific fields of it to the component itself, not the parent. IMO it is cleaner

PR #828, opened 2020-04-22
ui: scroll to top on history change

### LINKED ISSUE #825: Scroll to top on router transition
When you "change page" and you have scrolled down, the scroll does not go to the top, quite annoying.
Here the discussion: https://stackoverflow.com/questions/36904185/react-router-scroll-to-top-on-every-transition
Probably this is the fix: https://stackoverflow.com/a/52784960/675178

To be tested with anchors and transition to the same page.

ntarocco:
- fixes the behaviour of nothing scrolling to top when the history
  changes
- closes #825

### DIFF
--- a/ui/src/pages/backoffice/Document/DocumentDetails/DocumentDetails.js
+++ b/ui/src/pages/backoffice/Document/DocumentDetails/DocumentDetails.js
@@ -1,20 +1,18 @@
-import { DocumentContent } from './DocumentContent';
-import { DocumentHeader } from './DocumentHeader';
-import React, { Component } from 'react';
+import { Error, Loader } from '@components';
 import PropTypes from 'prop-types';
+import React, { Component } from 'react';
 import { Container, Divider, Grid, Ref, Sticky } from 'semantic-ui-react';
-import { Loader, Error } from '@components';
 import {
-  DocumentMetadata,
   DocumentActionMenu,
+  DocumentMetadata,
   DocumentSummary,
 } from './components';
+import { DocumentContent } from './DocumentContent';
+import { DocumentHeader } from './DocumentHeader';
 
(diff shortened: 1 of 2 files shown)


ntarocco:
@topless I tested the behaviour you were mentioning and the browser is smart enough to remember the scroll of the previous page.
So, when you change page, it will scroll to top. When you click back, it will be at the same scroll as before (not on top), probably is a browser feature to remember it.

KonstantinaStoikou (approved):

ui/src/pages/backoffice/Document/DocumentDetails/DocumentDetails.js
```diff
@@ -38,7 +38,7 @@ export default class DocumentDetails extends Component {
     const samePidFromRouter =
       prevProps.match.params.documentPid === documentPid;
     if (!samePidFromRouter) {
-      this.props.fetchDocumentsDetails(documentPid);
+      this.props.fetchDocumentDetails(documentPid);
```
KonstantinaStoikou, line 39: Since in the constructor there is `this.fetchDocumentDetails = this.props.fetchDocumentDetails` why not use `this.fetchDocumentDetails` instead of `this.props.fetchDocumentDetails`?

ntarocco: Nicely spotted thanks!
I have removed it from constructor, I think that @topless was proposing to just use `props` and avoid to set it in the constructor.

topless: I suggested that in order to avoid extra constructors for assigning props, and to prevent people (myself included) from renaming stuff, `fetchDocumentsDetails` would become `getDocument` and so on.

PR #835, opened 2020-04-27
PatronPastLoans: fix state in backoffice details

KonstantinaStoikou:
In patron details in backoffice, the state for past loans is always `ITEM_ON_LOAN` instead of `ITEM_RETURNED` and `CANCELLED`.


ui/src/state/PatronPastLoans/actions.js
```diff
@@ -7,7 +7,11 @@ const selectQuery = (patronPid, page) => {
   return loanApi
     .query()
     .withPatronPid(patronPid)
-    .withState(invenioConfig.circulation.loanCompletedStates)
+    .withState(
+      invenioConfig.circulation.loanCompletedStates.concat(
```
ntarocco: if you check the usage of `loanCancelledStates`, it is always use with the completed ones. More over, if you check for ILL, we put them together, see [here](https://github.com/inveniosoftware/invenio-app-ils/blob/master/ui/src/config/invenioConfig.js#L107).
I would suggest to eliminate the `loanCancelledStates` and change the `loanCompletedStates` to `loanCompletedStates: ['ITEM_RETURNED', 'CANCELLED'],`

ntarocco (approved):

PR #841, opened 2020-04-30
ui: complementary for 560 literature covers

topless:
- default cover placeholder when no cover
- moved all placeholders to backend static/images
- removed existing placeholders from front end
- related to https://github.com/CERNDocumentServer/cds-books/pull/93


ui/src/pages/frontsite/Documents/DocumentsSearch/SeriesListEntry/SeriesListEntry.js
```diff
@@ -21,7 +22,9 @@ export default class SeriesListEntry extends Component {
         as={Link}
         to={FrontSiteRoutes.seriesDetailsFor(this.metadata.pid)}
       >
-        <SeriesImage metadata={this.metadata} />
+        <SeriesImage
```
kpsherva: I know that you didn't introduce the, but if you could rename `SeriesImage` to `SeriesCover`, to keep it consistent with the Document. I think it would be better.

invenio_app_ils/circulation/mail/messages.py
```diff
@@ -16,7 +16,7 @@
 class LoanMessage(BlockTemplatedMessage):
     """Loan message."""
 
-    templates_base_dir = "invenio_app_ils_mail/circulation"
+    templates_base_dir = "invenio_app_ils_assets/circulation"
```
kpsherva: any reason to change the name? because if you wanted to generalise it, because we plan to put other assets than just e-mails in there, maybe they should be namespaced here by category f.e. /emails /static etc

invenio_app_ils/documents/covers_builder.py
```diff
@@ -7,14 +7,37 @@
...
+    return {
+        "small": "{url}/{isbn}-S.jpg".format(url=url, isbn=isbn),
+        "medium": "{url}/{isbn}-M.jpg".format(url=url, isbn=isbn),
+        "large": "{url}/{isbn}-L.jpg".format(url=url, isbn=isbn),
+    }
+
+
+def build_placeholder_urls():
+    """Build urls for default cover placeholders."""
+    image_path = url_for(
```
kpsherva: I am hesitant on placing this to be served by the backend. Because we mix now frontend and backend, especially if we will overwrite, everything on the frontend will be overwritten in react or theme and just this one on the backend. Plus we will operate on split repos... @ntarocco WDYT?

invenio_app_ils/circulation/mail/messages.py
topless: Thanks @kprzerwa I reverted it because it broke tests, @ntarocco as I see it we have a few options here.

- should we just put everything under `invenio_app_ils_mail` (although we shouldn't)
- should we introduce a new blueprint only for static
- if I am not wrong, all theses configurations templates, static, is part of the invenio-theme. Should we introduce `theme` module?

What do you think?

invenio_app_ils/documents/covers_builder.py
topless: This is actually the reason I re-worked on it. Initially I used the existing placeholder that was in ui public, but as @ntarocco foresaw the end client might be anything from webapp, mobile app or just a rest consumer, and then there will be missing a cover.

ntarocco:
@topless please remember to add a small part in the ILS doc on how you can override the default place holder. No need for a PR, you can just push the text. We will review all text at some point.

invenio_app_ils/documents/covers_builder.py
ntarocco: It is a long discussion, and we had many with Chris, here the TL;DR:
* ILS should not know what is your book provider: it could be Invenio itself (e.g. a folder with all the covers), or an external book provider. By default, it is the first, and so the backend should generate the URLs, based on the record fields
* The frontend should only put a `<img` with the URL, nothing else. If you start to have logic in the frontend, you might end up with this mixing part of the logic in the backend, to prepare some data, and part in the frontend
* The default place holder could be provided by the book cover provider or by your backend, and that's why we decided to go for this solution.

We can discuss IRL if you want.

invenio_app_ils/ext.py
```diff
@@ -189,7 +189,8 @@ def init_app(self, app):
             Blueprint(
                 "invenio_app_ils_mail",
                 __name__,
-                template_folder="templates"
+                static_folder="static",
```
ntarocco: I think it might be better not to touch mail for such thing. Can't we add it to the `invenio-theme` blueprint instead? Otherwise I would create one specific to ILS...

invenio_app_ils/literature/covers_builder.py
```diff
@@ -0,0 +1,43 @@
...
+        "medium": "{url}/{isbn}-M.jpg".format(url=url, isbn=isbn),
+        "large": "{url}/{isbn}-L.jpg".format(url=url, isbn=isbn),
+    }
+
+
+def build_placeholder_urls():
+    """Build urls for default cover placeholders."""
+    image_path = url_for(
+        "invenio_app_ils_mail.static",
+        filename="images/placeholder.png",
```
ntarocco, line 37: great, is it easy to override in an overlay?

invenio_app_ils/literature/covers_builder.py
```diff
@@ -0,0 +1,43 @@
...
+        _external=True
+    )
+    return {
+        "small": image_path,
```
ntarocco, line 42: I don't know if @kprzerwa wants to comment here, but we were chatting that maybe we want to know in the UI if we are displaying a real cover or a placeholder, to allow even more customization? A solution would be just to add an extra field `is_placeholder: true/false`...

invenio_app_ils/ext.py
topless: I have created a dummy blue print only for static.

```python
app.register_blueprint(
    Blueprint(
        "invenio_app_ils_static",
        __name__,
        static_folder="static",
    )
)
```

This is the global static folder configuration, and the image can be accessed like `/static/images/placeholder.png`. We can always extend the `invenio-theme`, this will move the file from the global static to the theme module specific static, so it will be available from `theme/static/images/placeholder.png`. The later approach is the clean-separated approach, but the current one will suffice too.

invenio_app_ils/documents/covers_builder.py
kpsherva: I discussed with @ntarocco, to solve possible UI problems I was thinking about,  if you attach in the response payload for the cover a flag that it indicates we are sending the placeholder url, I think it would be ok like `no_cover: True` so the UI can know that it is getting a placeholder.

topless: Done, very good idea, it is `is_placeholder` the property that comes in `cover_metadata`.

ntarocco (approved):

kpsherva:
@topless you have the same problem as @KonstantinaStoikou had (with your UI tests). Deleting and installing fresh node_modules should help

Issue #840, opened 2020-04-29
Change SortBy/SortOrder by Sort components


ntarocco:
## Describe the bug

The current sorting does not allow to handle correctly the default sorting order on empty query.
What happens is that the list of results is sorted by default `Newest` `Asc`, so basically you see the oldest.

## Steps to Reproduce

1. Go to home page
2. Click on click on search
3. See the list of results and default sort options

## The current behavior

Show oldest first by default.

## The expected behavior

Show newes first by default.

PR #842, opened 2020-04-30
various bug fixes while testing UI

ntarocco:


kpsherva (approved):

ntarocco:
@kprzerwa @topless @KonstantinaStoikou 
I am not 100% of my changes about the sorting honestly, in particular if you see this:
```
sortByNewest() {
    this.sortBy = `&sort=-mostrecent`;
    return this;
  }
```
shouldn't it be more `sort=mostrecent`, logically speaking? I expect to see on the top, as first, the most recent, so why the `-` (desc) to achieve that?

I have created an issue [here](https://github.com/inveniosoftware/invenio-app-ils/issues/840), we should review all this and change component to use only one sort by (not sort order) and have:
```
- Newest
- Oldest
...
```

PR #844, opened 2020-04-30
ReactSearchkit: hide components when zero results

### LINKED ISSUE #745: ReactSearchKit components are displayed even if 0 results
In all search pages with `<ReactSearchKit`, frontsite and backoffice, when the search query gives zero results, you can still see some components displayed, see the screenshot:

Banners_and_Alerts_and_Mozilla_Firefox

On top of that, when you do not input anything in the query string but you select some filters and you have zero results, you will see that component with `Current search ""`.

ReactSearchKit `0.19.0` supports `label` prop in some components.

1. Change the text to use the `label` prop instead.
2. Show the `Current search ""` text only when the query string is not `''`.

KonstantinaStoikou:
Closes #745


ui/src/components/SearchControls/SearchControls.js
```diff
@@ -9,28 +9,19 @@ import { SearchControlsMobile } from './SearchControlsMobile';
...
-    );
-  };
-
   render() {
     return (
       <>
         <Responsive {...Responsive.onlyComputer}>
           <Grid columns={3} className={'search-controls'}>
             <Grid.Column largeScreen={5} computer={6}>
+              <div style={{ marginBottom: '10px' }}>
```
kpsherva: please don't use inline styles (they will be hard to find and impossible to overwrite in theming) and try to limit the usage of absolute CSS units to when it is last resort (it causes usually problems with responsive layouts)
instead of that use one of existing CSS classes or if not available, create a new one in `ui/semantic-ui/.../site.overwrites`

kpsherva (commented):
thanks for fixing it, good job with using the react-searchkit component in the end.
there is just one comment to fix

ntarocco (commented):
@KonstantinaStoikou does this fixes basically all search pages?

KonstantinaStoikou:
@ntarocco I checked only for the frontsite search

ntarocco:
> @ntarocco I checked only for the frontsite search

and the search page in the Serials details page in the frontsite?
can you have a look to the backoffice too? You can fix also there or please do not close the issue, just update it mentioning that it has been fixed for frontsite only.
Thanks!

KonstantinaStoikou:
@ntarocco I fixed the series details search and the backoffice and I also removed the `SeriesLiteratureResultsList.test.js` because it failed and there were no tests for the respective `DocumentSearchResultsList`,`DocumentSearchResultsGrid` and `SeriesLiteratureSearchResultsGrid`

ntarocco (approved):

PR #866, opened 2020-05-19
ILL: implement extension workflow backend

ntarocco:
blocked by: https://github.com/inveniosoftware/invenio-circulation/pull/131


invenio_app_ils/circulation/api.py
```diff
@@ -77,6 +79,8 @@ def request_loan(
         raise PatronHasRequestOnDocumentError(patron_pid, document_pid)
     _validate_delivery(kwargs.get("delivery"))
 
+    transaction_user_pid = transaction_user_pid or current_user.id
```
topless: Instead of redefining `transaction_user_pid` we can put the condition where it is actually used.

invenio_app_ils/circulation/loaders/schemas/json/base.py
```diff
@@ -9,11 +9,23 @@
 
 from flask import current_app
 from flask_babelex import lazy_gettext as _
-from invenio_records_rest.schemas import RecordMetadataSchemaJSONV1
+from invenio_rest.serializer import BaseSchema as InvenioBaseSchema
```
topless, line 12: Why not `BaseSchema`?

invenio_app_ils/circulation/utils.py
```diff
@@ -59,19 +59,10 @@ def circulation_build_document_ref(loan_pid, loan):
...
-def circulation_default_extension_duration(loan):
-    """Return a default extension duration in timedelta."""
-    return timedelta(days=30)
-
-
 def circulation_default_extension_max_count(loan):
     """Return a default extensions max count."""
-    # NOTE: If user is admin or librarian always have 1 extra extension count.
+    # NOTE: If user is admin or librarian it should be unlimited.
+    # Set it has 1 more than the current.
```
topless: `Set it has 1 more than the current.` not very clear

invenio_app_ils/ill/api.py
```diff
@@ -159,40 +186,152 @@ def create_loan(
...
             raise ILLError("The loan end date cannot be in the past.")
 
-        item_pid = dict(type=self._pid_type, value=self["pid"])
+        item_pid = dict(type=rec._pid_type, value=rec["pid"])
         pid, loan = checkout_loan(
             start_date=start_date,
-            end_date=self["loan_end_date"],
-            document_pid=self["document_pid"],
+            end_date=end_date,
+            document_pid=rec["document_pid"],
```
topless, line 202: ~~Is it guaranteed that it will have `document_pid`?~~ as I understand the aforementioned record is an `item` so `document_pid` is guaranteed since it is required in the item schema.

invenio_app_ils/ill/api.py
```diff
@@ -159,40 +186,152 @@ def create_loan(
...
+    item = Item.get_record_by_pid(item_pid["value"])
+    return circulation_default_loan_duration_for_item(item)
+
+
+def circulation_default_extension_duration(loan, initial_loan):
+    """Return a default extension duration in timedelta."""
+    item_pid = loan["item_pid"]
+    validator = validate_item_pid(item_pid)
+    if validator.is_brw_req:
+        new_end_date_already_set = loan["end_date"] == initial_loan["end_date"]
```
topless: nit: since its a boolean maybe something like  `is_end_date_set`?

invenio_app_ils/ill/loaders/jsonschemas/patron_loan_extension_actions.py
```diff
@@ -0,0 +1,120 @@
...
+    if record["status"] != "ON_LOAN":
+        raise ValidationError(
+            "This interlibrary loan is currently not on loan."
+        )
+
+    ext_status = (
+        record.get("patron_loan", {}).get("extension", {}).get("status")
+    )
+    is_valid = ext_status in BorrowingRequest.EXTENSION_STATUSES
+    if ext_status and not is_valid:
```
topless: Why not directly `if ext_status in BorrowingRequest.EXTENSION_STATUSES` and skip the `is_valid`

invenio_app_ils/ill/loaders/jsonschemas/patron_loan_extension_actions.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @pre_load
+    def validate_action(self, data, **kwargs):
+        """Validate action."""
+        record = self.context["record"]
+        ext_status = validate_statuses(record)
+        if ext_status == "PENDING":
+            raise ValidationError(
+                "The extension of this interlibrary loan has been already "
+                "requested."
```
topless: nit: "An extension for this interlibrary loan has already been requested"

invenio_app_ils/ill/loaders/jsonschemas/patron_loan_extension_actions.py
```diff
@@ -0,0 +1,120 @@
...
+            )
+        elif ext_status == "DECLINED":
+            raise ValidationError(
+                "Cannot request an extension for this interlibrary loan "
+                "because it has been previously declined."
```
topless:
```suggestion
                "because the most recent extension request has been declined."
```

invenio_app_ils/permissions.py
```diff
@@ -100,34 +100,25 @@ def loan_extend_circulation_permission(loan):
         document_rec = Document.get_record_by_pid(loan["document_pid"])
         document = document_rec.replace_refs()
         is_overbooked = document.get("circulation", {}).get("overbooked")
-        if is_overbooked == None:
+        if is_overbooked is None:
```
topless, line 103: Do we have to explicitly check for `None`?

ui/src/pages/backoffice/ILL/BorrowingRequest/BorrowingRequestDetails/components/BorrowingRequestPatronLoan/BorrowingRequestPatronLoan.js
```diff
@@ -32,7 +32,7 @@ class CreateLoanButton extends React.Component {
     }
 
     const hasRequestedStatus = brwReq.status === 'REQUESTED';
-    const hasAlreadyLoan = brwReq.loan_pid;
+    const hasAlreadyLoan = _get(brwReq, 'patron_loan.pid', false);
```
topless:
```suggestion
    const hasLoan = _get(brwReq, 'patron_loan.pid', false);
```

ui/src/pages/backoffice/ILL/BorrowingRequest/BorrowingRequestDetails/BorrowingRequestStatistics.js
```diff
@@ -43,13 +43,13 @@ export class BorrowingRequestStatistics extends React.Component {
     }
   }
 
-  renderLoanEnds() {
-    const { loan_end_date: loanEndDate } = this.props.brwReq;
+  renderDueDate() {
+    const { due_date: dueDate } = this.props.brwReq;
```
topless, line 47: Maybe we can put the props that are used in the component like  `due_date` and the `status` in the propTypes so we start becoming more explicit, lets not collide though with @kprzerwa

topless (approved):

invenio_app_ils/circulation/api.py
ntarocco: I guess this is more developer style: I personally prefer, in general, to have all inputs validation/sanitization on the top. If a variable is used multiple times (not sure this is the case), I would set this once and not multiple times, each time it is used.

invenio_app_ils/circulation/loaders/schemas/json/base.py
ntarocco: because `Schema` is coming from `marshmallow`, and for me `BaseSchema` might mean that it is also coming from there. I want to make sure that we understand that it is coming from Invenio, for some reasons.

invenio_app_ils/circulation/utils.py
ntarocco: thanks, changed code and removed comment

invenio_app_ils/ill/api.py
ntarocco: The record is a `BorrowingRequest` and yes you are right the `document_pid` must be there, enforced by the schema.

invenio_app_ils/ill/api.py
ntarocco: Thanks, changed

invenio_app_ils/ill/loaders/jsonschemas/patron_loan_extension_actions.py
ntarocco: Because in normal cases `ext_status` is non-existing: the check make sure that, if defined, that must be valid.
Extension statuses are: [`PENDING`, `DECLINED`]. The absence of it means that the extension is not requested.

The initial design of this feature had an extra state `NOT_REQUESTED`, but this will imply that any new `BorrowingRequest` record will have, by default, newly created a:
```
...
patron_loan: {
  extension: {
    status: 'NOT_REQUESTED'
  }
  pid: None
}
...
```
IMHO, this is polluting very much records, so I removed.
I changed the code to be hopefully more clear.

invenio_app_ils/permissions.py
ntarocco: yes, read the comment just below, if you remember you developed it. It is an extra check just in case is not `True` or `False` for some reasons.

ui/src/pages/backoffice/ILL/BorrowingRequest/BorrowingRequestDetails/BorrowingRequestStatistics.js
ntarocco: This requires quite a huge change. I am not sure we want to have the object shape of an entire record honestly. If yes, then IMHO we will have to create strict backend serializers. To be discussed.

ntarocco:
I had to temporary skip tests because we hit the limit on Travis. To be restored when we fix it.

PR #873, opened 2020-05-25
ui: display more details about author

### LINKED ISSUE #812: display more information about the author
- [x] Find a good way to extend the display of the author additional information if user interested (popup over the author? - review the existing semantic-ui components and propose one)

- [x] Limit the number of displayed authors to 10 (DocumentAuthors component).
If more than 10, the authors should be folded but accessible by click ("more" button). In extreme cases (300 authors - here are such use cases), the area of this component should be limited to a box and scrollable within.

KonstantinaStoikou:
Closes #812 
(all icons have changed to `i` instead of `?`)
Selection_005
Selection_003
Selection_004
Screenshot from 2020-05-26 11-32-14
Screenshot from 2020-05-26 09-43-53

### DIFF
--- a/ui/src/components/Document/DocumentAuthors.js
+++ b/ui/src/components/Document/DocumentAuthors.js
@@ -3,6 +3,95 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import isEmpty from 'lodash/isEmpty';
 
+class Roles extends Component {
+  render() {
+    return (
+      <>
+        <span>{this.props.roles.join(', ')}</span>
+        <br />
+      </>
+    );
+  }
+}
+
+Roles.propTypes = {
+  roles: PropTypes.array.isRequired,
+};
+
+class Affiliations extends Component {
+  render() {
+    return (
+      <>
+        <span>
+          Aff. :{' '}
+          {this.props.affiliations
+            .map(item => {
+              return item.name;
+            })
+            .join(', ')}
+        </span>
+      </>
+    );
+  }
+}
+
+Affiliations.propTypes = {
+  affiliations: PropTypes.array.isRequired,
+};
+
+class AlternativeNames extends Component {
+  render() {
+    return (
+      <>
+        <br />
+        <span>Alter. names: {this.props.alternative_names.join(', ')}</span>
+        <br />
+      </>
+    );
+  }
+}
+
+AlternativeNames.propTypes = {
+  alternative_names: PropTypes.array.isRequired,
+};
+
+class Identifiers extends Component {
+  render() {
+    return (
+      <>
+        <span>
+          Ids:{' '}
+          {this.props.identifiers
+            .map(item => {
+              return `${item.value} (${item.scheme})`;
+            })
+            .join(', ')}
+        </span>
+        <br />
+      </>
+    );
+  }
+}
+
+Identifiers.propTypes = {
+  identifiers: PropTypes.array.isRequired,
+};
+
+class Type extends Component {
+  render() {
+    return (
+      <>
+        <span>Type: {this.props.type}</span>;
+        <br />
+      </>
+    );
+  }
+}
+
+Type.propTypes = {
+  type: PropTypes.string.isRequired,
+};
+
 export class DocumentAuthors extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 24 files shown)


ui/src/components/Document/DocumentAuthors.js
```diff
@@ -1,28 +1,123 @@
...
+      prefix,
+      otherAuthorsDisplay,
+      popupDisplay,
+      delimiter,
+      authorsLimit,
+    } = this.props;
     const otherAuthors = otherAuthorsDisplay ? otherAuthorsDisplay : 'et al.';
+
+    let authors = metadata.authors;
+    let classes = 'document-authors-list-wrapper';
```
kpsherva: to improve readability you could just add "expanded" to the class, f.e (pseudocode)
```
const expandedClass = this.state.expanded && "expanded";
```

and then below, in the div:

```
`document-authors-list-wrapper ${expandedClass}`
```

ui/src/components/Document/DocumentAuthors.js
```diff
@@ -1,28 +1,123 @@
...
+    if (metadata && metadata.authors) {
+      authors = [...metadata.authors];
+      classes =
+        metadata.authors.length > 300 && this.state.expanded
```
kpsherva: `300` should be a prop probably, not hardcoded

ui/src/semantic-ui/site/elements/list.overrides
```diff
@@ -7,6 +7,11 @@
     width: 100%;
   }
 
+  .document-authors-list-height {
```
kpsherva: please do not name classes containing css attribute name in the name of the class, if it is not only defining the attribute
in this case you name it `...-height` when it is also defining overflow and it might be confusing.
for this case, as it is very specific you could just use less parent selector:
```
.ui.list.document-authors-list{
... already defined stuff
&.expanded{
    max-height: 3em; <- also better use em unit in this case to keep consistency with the other text 
    overflow-y: scroll;
}
}

File: ui/src/semantic-ui/site/elements/button.overrides [hunk 4]
```diff
@@ -23,6 +23,7 @@
   .button-show-more {
     color: @fsPrimaryColor;
     font-weight: bold;
+    cursor:pointer;
```
kpsherva: nicely spotted! 
please remember to use code formatting

kpsherva (commented):
Looks nice!
I just have a little comment about the styling. The field with the authors jumps to the next line when it is expanded, below the label `by`, could you fix it to be in the same line ?

Or even better, we could have the scroll box in the metadata section, and simply `et al.` in the top, when it's more than 2 lines, WDYT?

Also a comment (I don't know if you already mentioned @ntarocco ) could you test if the `?` icon  is "clickable" on mobile view?

ui/src/components/Document/DocumentAuthors.js
```diff
@@ -1,28 +1,123 @@
...
+      expanded: false,
+    };
+  }
+
+  renderPopupContent = author => {
+    return (
+      <>
+        {author.full_name}
+        <br />
+        {!isEmpty(author.roles) && (
```
ntarocco: It would be nice to limit the checks for empty. Moreover, the list should be joined rather than using `toString`.
To achieve this, prefer components to functions (pseudocode):
```
class Author {
  render() {
    ...
    {author.roles.join(" ")}
    ...

  }
}

Author.propTypes = {
  author = Obj.shape(
    roles: array
  )
}
Author.default = {
  roles: []
}


class DocumentAuthors {
  render() {
    return authors.map((author, index) => <Author author={author} />
  }
}

DocumentAuthors.propTypes = {
    authors: array
}
DocumentAuthors.default = {
  authors: []
}

```
Like this you don't need less checks and you take advantage of default props. This is applicable to each field.

This is waht I was mentioning to @kprzerwa yesterday too.

ui/src/components/Document/DocumentAuthors.js
```diff
@@ -1,28 +1,123 @@
...
+    );
+  };
+
   render() {
-    const { metadata, prefix, otherAuthorsDisplay, delimiter } = this.props;
+    const {
+      metadata,
+      prefix,
+      otherAuthorsDisplay,
+      popupDisplay,
```
ntarocco, line 161: Where does this come from?

ntarocco:
I would personally go for an `i` icon (for info), rather than a question mark... but this is just IMHO

ntarocco (approved):

PR #880, opened 2020-05-29
global: compatibility agg/sort fields with es mappings text

topless:
@kprzerwa the following configuration will be required to enable them correctly in the moved `searchConfig.js`
```
const searchConfig = {
  ...
  acqVendors: {
    filters: [],
    sortBy: {
      onEmptyQuery: "bestmatch",
      values: [
        {
          default_order: "asc",
          field: "name.keyword",
          order: 1,
          title: "Name",
        },
        {
          default_order: "asc",
          field: "bestmatch",
          order: 2,
          title: "Best match",
        },
      ],
    },
    sortOrder: ["asc", "desc"],
  },
  ...
};
```

### DIFF
--- a/invenio_app_ils/acquisition/config.py
+++ b/invenio_app_ils/acquisition/config.py
@@ -136,7 +136,10 @@ RECORDS_REST_SORT_OPTIONS = dict(
     ),
     acq_vendors=dict(  # VendorSearch.Meta.index
         name=dict(
-            fields=["name"], title="Name", default_order="desc", order=1
+            fields=["name.keyword"],
+            title="Name",
+            default_order="desc",
+            order=1
         ),
         bestmatch=dict(
             fields=["-_score"],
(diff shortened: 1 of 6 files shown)


invenio_app_ils/acquisition/config.py
```diff
@@ -136,7 +136,10 @@
     ),
     acq_vendors=dict(  # VendorSearch.Meta.index
         name=dict(
-            fields=["name"], title="Name", default_order="desc", order=1
+            fields=["name.keyword"],
```
ntarocco, line 139: I am not sure I understand the fix: if I a vendor called `Amazon`, and I search for `Amaz`, I won't find it I believe... Shouldn't we search in the text instead?

topless: @ntarocco yes we should do the search in text, but then we encounter the following issue `Fielddata is disabled on text fields by default.`, it is described in detail in https://www.elastic.co/guide/en/elasticsearch/reference/current/fielddata.html

The other thing we can possibly do, is to introduce them both, text for the searches and an extra keyword field for the aggregations.

ntarocco (approved):

PR #883, opened 2020-06-02
loan: auto cancel after expiration day

### LINKED ISSUE #758: loan: auto expire for valid until date
When we are creating a loan request it comes with a predefined `valid_untii` date. When this date comes the loan should be automatically cancelled with the reject reason (or cancel reason if you prefer), `EXPIRED` .

KonstantinaStoikou:
Closes #758

### DIFF
--- a/invenio_app_ils/circulation/search.py
+++ b/invenio_app_ils/circulation/search.py
@@ -17,9 +17,7 @@ from invenio_circulation.search.api import search_by_pid
 
 def get_pending_loans_by_doc_pid(document_pid):
     """Return any pending loans for the given document."""
-    return search_by_pid(
-        document_pid=document_pid, filter_states=["PENDING"]
-    )
+    return search_by_pid(document_pid=document_pid, filter_states=["PENDING"])
 
 
 def get_active_loans_by_doc_pid(document_pid):
@@ -77,6 +75,17 @@ def get_all_expiring_loans(expiring_in_days):
                 gte="{}||/d".format(future), lte="{}||/d".format(future)
             ),
         )
+    )
+
+
+def get_all_expired_loans():
+    """Return all loans that have expired."""
+    states = current_app.config["CIRCULATION_STATES_LOAN_REQUEST"]
+    search_cls = current_circulation.loan_search_cls
+    return (
+        search_cls()
+        .filter("terms", state=states)
+        .filter("range", request_expire_date=dict(lt="now/d"))
     )
 
 
--- a/invenio_app_ils/circulation/tasks.py
+++ b/invenio_app_ils/circulation/tasks.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Circulation tasks."""
+
+from copy import deepcopy
+
+from celery import shared_task
+from celery.utils.log import get_task_logger
+from flask import current_app
+from invenio_circulation.api import Loan
+from invenio_circulation.proxies import current_circulation
+from invenio_db import db
+
+from invenio_app_ils.circulation.search import get_all_expired_loans
+from invenio_app_ils.records.api import SystemAgent
+
+celery_logger = get_task_logger(__name__)
+
+
+@shared_task
+def cancel_expired_loan_requests():
+    """Cancel loan requests after expiration date has passed."""
+    expired_loans = get_all_expired_loans().execute()
+    for hit in expired_loans.hits:
+        loan = Loan.get_record_by_pid(hit.pid)
+        duration_days = current_app.config[
+            "ILS_CIRCULATION_LOAN_REQUEST_DURATION_DAYS"
+        ]
+        params = deepcopy(loan)
+        system_agent_id = str(SystemAgent.id)
+        params.update(
+            dict(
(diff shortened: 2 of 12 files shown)


invenio_app_ils/circulation/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@shared_task
+def cancel_expired_loan_request():
+    """Cancel loan request after expiration date has passed."""
+    expired_loans = get_all_expired_loans().execute()
+    for hit in expired_loans.hits:
+        loan = Loan.get_record_by_pid(hit.pid)
+        if loan["state"] != "CANCELLED":
+            loan["state"] = "CANCELLED"
```
ntarocco: To cancel a loan, you need to trigger a transition and not modify the loan directly. For example:
```
params = dict(
    cancel_reason=...
)
current_circulation.circulation.trigger(
            loan, **dict(params, trigger="cancel")
        )
```

invenio_app_ils/circulation/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+celery_logger = get_task_logger(__name__)
+
+
+@shared_task
+def cancel_expired_loan_request():
+    """Cancel loan request after expiration date has passed."""
+    expired_loans = get_all_expired_loans().execute()
+    for hit in expired_loans.hits:
+        loan = Loan.get_record_by_pid(hit.pid)
+        if loan["state"] != "CANCELLED":
```
ntarocco: In the search, this loan should be `active`, so I would say that this extra check is not needed.

invenio_app_ils/circulation/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+            loan["state"] = "CANCELLED"
+            loan["cancel_reason"] = "EXPIRED"
```
ntarocco: I would add a configurable text here instead, because we might want to display it or it might be sent by email.

invenio_app_ils/circulation/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+            loan.commit()
+            db.session.commit()
+            current_app_ils.item_indexer.index(loan)
```
ntarocco: why `item_indexer`? Here you are indexing a loan, you will have to use the `loan_indexer`

invenio_app_ils/circulation/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+from invenio_db import db
+
+from invenio_app_ils.circulation.search import get_all_expired_loans
+from invenio_app_ils.proxies import current_app_ils
+
+celery_logger = get_task_logger(__name__)
+
+
+@shared_task
+def cancel_expired_loan_request():
```
ntarocco: minor
```suggestion
def cancel_expired_loan_requests():
```

invenio_app_ils/circulation/search.py
```diff
@@ -80,6 +80,17 @@ def get_all_expiring_loans(expiring_in_days):
...
 
 
+def get_all_expired_loans():
+    """Return all loans that have expired."""
+    states = current_app.config["CIRCULATION_STATES_LOAN_REQUEST"]
+    search_cls = current_circulation.loan_search_cls
+    return (
+        search_cls()
+        .filter("terms", state=states)
+        .filter("range", request_expire_date=dict(lt="now/d"))
```
ntarocco, line 88: does it makes sense to check for past ones too? In case the cron runs late...

tests/api/circulation/test_expired_loans.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # expired loans
+        date = now - timedelta(days=1)
+        new_expiration_date(loans[0], date)
+        new_expiration_date(loans[1], date)
+        date = now - timedelta(days=2)
+        new_expiration_date(loans[2], date)
+        expired_pids = [loans[0]["pid"], loans[1]["pid"], loans[2]["pid"]]
+
+        # not expired loans
```
kpsherva, line 49: nice  that you took into account to check the not expired too!

invenio_app_ils/records/api.py
```diff
@@ -408,9 +416,23 @@ def get_patron(patron_pid):
         if not patron_pid:
             raise PatronNotFoundError(patron_pid)
 
+        if str(patron_pid) == "-1":
+            return SystemPatron()
+
         return Patron(patron_pid)
 
 
+class SystemPatron(Patron):
```
kpsherva: nit: `SystemAgent` maybe as a name, maybe others have ideas, @ntarocco, @topless ? We can also leave it like this.

kpsherva (approved):

invenio_app_ils/records/api.py
ntarocco: Agree, @KonstantinaStoikou I would also add:
```
SystemAgent(Patron):
  id = -1
  ...
```
so that you can replace all occurrences of:
```
SystemPatron().dumps()["id"]
```
with
```
str(SystemAgent.id)
```

invenio_app_ils/records/api.py
```diff
@@ -408,9 +416,23 @@ def get_patron(patron_pid):
         if not patron_pid:
             raise PatronNotFoundError(patron_pid)
 
+        if str(patron_pid) == "-1":
```
ntarocco: After the change below:
```suggestion
        if str(patron_pid) == str(SystemAgent.id):
```

ntarocco (approved):

PR #889, opened 2020-06-08
loans: display error when requesting document if patron already has active loan

### LINKED ISSUE #860: Loan request possible on a book which is currently loaned by the same patron
## Description
A user requests a loan on a book he already has an active loan (ongoing)

## Steps to Reproduce

1. Find a document with an ongoing loan
2. Log in as the patron
3. Try to request the same document
4. Request is accepted

## The current behavior

A patron can create requests on a book he currently has on loan.

## The expected behavior

It should display an error on the attempt of such a request. It helps to overpass the extension

KonstantinaStoikou:
Closes #860

### DIFF
--- a/invenio_app_ils/circulation/api.py
+++ b/invenio_app_ils/circulation/api.py
@@ -21,7 +21,8 @@ from invenio_circulation.search.api import search_by_patron_item_or_document
 from invenio_db import db
 
 from invenio_app_ils.errors import MissingRequiredParameterError, \
-    PatronHasLoanOnItemError, PatronHasRequestOnDocumentError
+    PatronHasLoanOnDocumentError, PatronHasLoanOnItemError, \
+    PatronHasRequestOnDocumentError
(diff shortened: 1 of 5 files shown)


invenio_app_ils/circulation/api.py
```diff
@@ -77,6 +92,12 @@ def request_loan(
         patron_pid=patron_pid, document_pid=document_pid
     ):
         raise PatronHasRequestOnDocumentError(patron_pid, document_pid)
+
+    if patron_has_active_loan_on_document(
```
ntarocco: what about if we avoid 2 searches and we use the check above by checking also for this extra state?
You will be able to throw a different error based on the result eventually.

Can you also please add a test? It is important that we test all these cases.

ntarocco (approved):
Nice! LGTM! And the test looks clean! 👏

kpsherva (approved):

Issue #658, opened 2019-12-13
add identifiers field to e-items


kpsherva:
The E-Items usually have the DOI identifier on them (they belong more to the ebook than the document), So they should be on the level of the electronic copy. Especially because it enables the possibility of keeping separate concerns (and keeping the data clean)

Add the same structure of identifiers field as on Document schema

- [x] in the schema of e-item
- [x] in the create/update forms of e-item
- [x] check if displayed properly on the backoffice and frontsite
- [x] create demo data with some identifiers and with an empty field (to check in each case)

mainly for DOIs use case

PR #914, opened 2020-07-22
schemas: add identifiers to e-items

FlorianCassayre:
* identifiers schema is borrowed from documents (scheme, value, material)
* see #658

### DIFF
--- a/invenio_app_ils/cli.py
+++ b/invenio_app_ils/cli.py
@@ -264,6 +264,20 @@ class EItemGenerator(Generator):
             for pid in range(1, size + 1)
         ]
 
+        for obj in objs:
+            if bool(random.getrandbits(1)):
+                obj["identifiers"] = random.choice([[
+                    {
+                        "scheme": "ISBN",
+                        "value": "0-395-36341-1",
+                        "material": "EBook"
+                    },
+                    {
+                        "scheme": "DOI",
+                        "value": "doi:10.1340/309registries"
+                    }
+                ], []])
+
         self.holder.eitems["objs"] = objs
 
     def persist(self):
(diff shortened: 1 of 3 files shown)


invenio_app_ils/cli.py
```diff
@@ -259,6 +259,21 @@ def generate(self):
...
+                        "value": "URN:ISBN:0-395-36341-1"
+                    },
+                    {
+                        "scheme": "ISBN",
+                        "value": "0-395-36341-1",
+                        "material": "EBook"
+                    },
+                    {
+                        "scheme": "DOI",
+                        "value": ""
```
kpsherva: in general empty value should not happen and if it does, it should raise a ValidationError
checking the empty identifiers case should rather be as:
sometimes the records are created with `identifiers` field filled like you did here 
OR with 
```
`identifiers` = [] # or identifiers not in the record at all
```

to check how the application behaves with empty data field

invenio_app_ils/eitems/loaders/jsonschemas/eitems.py
```diff
@@ -11,6 +11,19 @@
 from marshmallow import EXCLUDE, Schema, fields
 
 
+class IdentifierSchema(Schema):
```
kpsherva: I think we have the same on the document can we reuse ? ping @ntarocco

FlorianCassayre: That's right, it's essentially `invenio_app_ils.documents.loaders.jsonschemas.document.IdentifierSchema`.

invenio_app_ils/cli.py
FlorianCassayre: You're absolutely right, I misunderstood what you meant.

Now that you point it out, it seems that the non-emptiness requirement of `value` is not checked by the server (and the problem is the same for document identifiers).

kpsherva: I think in the schema it basically thinks that empty string is allowed as a value, when it should not be

kpsherva (approved):

PR #921, opened 2020-07-27
global: CSRF support

### LINKED ISSUE #867: Enable REST Api CSRF validation
## Description
With `invenio-rest@1.2.1`, CSRF validation is baked in. This task describes what changes need to be done in ILS:

- [x] Enable CSRF validation by adding into config.py:
```python
REST_CSRF_ENABLED = True
```
- [x] Edit `axios` instance in `ui/src/api/base.js `:
```javascript
const apiConfig = {
 ...
  xsrfCookieName: 'csrftoken',
  xsrfHeaderName: 'HTTP_X_CSRFTOKEN',
```
- [x] CSRF cookie is set by default only in https sites. If you want to run ILS under https you could edit the `package.json` and add a new command/update the `start` command:
```javascript
...
"scripts": {
    "start:secure": "HTTPS=true craco start",
}
...
```
The alternative solution in the development mode would be to add in the `config.py`:
```python
    CSRF_FORCE_SECURE_REFERRER = False
    SESSION_COOKIE_SECURE = False
```
- [x] Update `INSTALL.rst` for using CSRF validation in development. For example:
```
CSRF ENABLE DEVELOPMENT SETUP (Without docker)
----------------------------------------------
If you have enabled the csrf check during development, you need to follow one of the below steps in order to be able to make it work:

1) Run the ui app with `npm run start:secure`. This will run the react application over `https` and the csrf cookie will be corectly set.
2) If you run the ui app over `http` then you need to add in the `config.py` file the following lines:

.. code-block:: console
    CSRF_FORCE_SECURE_REFERRER = False
    SESSION_COOKIE_SECURE = False
```

topless:
- updgrad invenio-rest to version 1.2.1 for CSRF support
- added config variable to enable CSRF
- closes #867

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,8 @@ install_requires = [
     "invenio-pidstore>=1.2.0,<1.3.0",
     "invenio-records-rest>=1.7.1,<1.8.0",
     "invenio-records>=1.3.1,<1.4.0",
+    # needed for CSRF token support
+    "invenio-rest>=1.2.1",
     # --- `files` bundle with only invenio-files-rest ----------------------
     "invenio-files-rest>=1.2.0,<1.3.0",
     # --- extra deps of ILS ------------------------------------------------
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -206,6 +206,8 @@ JSONSCHEMAS_HOST = "127.0.0.1:5000"
 # CORS
 ###############################################################################
 REST_ENABLE_CORS = True
+REST_CSRF_ENABLED = True
+
 # change this only while developing
 CORS_SEND_WILDCARD = True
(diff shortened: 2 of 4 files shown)


ntarocco (approved):

setup.py
```diff
@@ -68,6 +68,8 @@
     "invenio-pidstore>=1.2.0,<1.3.0",
     "invenio-records-rest>=1.7.1,<1.8.0",
     "invenio-records>=1.3.1,<1.4.0",
+    # needed for CSRF token support
```
ntarocco, line 71:
```suggestion
    # --- `base` bundle upgrading invenio-rest -----------------------------
    # needed for CSRF token support
```

PR #927, opened 2020-08-03
Cli: add user anonymization command

### LINKED ISSUE #925: GDPR: right to be forgotten
In case of patron request, the system should allow to anoymise any activity.

### Tasks

- [x] create a callable function that can return all patron activity
- [x] create a callable function that will replace any patron info with anonymous info: this involves both DB and ES indexing
- [x] create 2 CLI commands that can call these 2 functions: the listing activity should display a nice report, the anonymization should ask for confirmation (irreversible operation)
- [x] add a section in the ILS documentation to explain all this and how to use it

### Extra details

**Patron data**

In ILS, records normally store only the patron PID, and then it is the `$ref` that resolves the patron info. Patron info are stored in:
- Loans
- Acquisitions
- ILLs
- Document requests

It should be enough to anonymise info in the DB (see below) and re-index all records referencing this patron PID.

**User info in the DB**

Code to be tested:

```
    with db.session.begin_nested():
        user_identity = UserIdentity.query.filter(UserIdentity.id_user == user_id).one()
        // check table structure if to be anonymized. Data only in case of OAuth (GitHub, CERN, etc..)

        remote_account = RemoteAccount.query.filter(RemoteAccount.user_id == user_id).one()
        // check table structure if to be anonymized. Data only in case of OAuth (GitHub, CERN, etc..)

        user_profile = UserProfile.query.filter(UserProfile.user_id == user_id).one()
        // check table structure if to be anonymized

        user = User.query.filter(User.id == user_id).one()
        // check table structure if to be anonymized
```

KonstantinaStoikou:
Closes #925

- Command to list patron's activity 
   - Display patron related data and activity with a pretty printed dict
- Command to anonymize patron with force option in case of interrupted call
   - Delete user profile and all relations to the user 
   - Re-index all records and replace patron info with dict with `Unknown` values
   - Display number of rows deleted from db and records re-indexed

### DIFF
--- a/invenio_app_ils/anonymization.py
+++ b/invenio_app_ils/anonymization.py
@@ -1,0 +1,201 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2018-2020 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Functions to anonymize user data and activity."""
+
+
+from copy import deepcopy
+
+from elasticsearch.exceptions import NotFoundError
+from flask import current_app
+from invenio_accounts.models import SessionActivity, User, userrole
+from invenio_circulation.api import Loan
+from invenio_circulation.proxies import current_circulation
+from invenio_db import db
+from invenio_oauthclient.models import RemoteAccount, UserIdentity
+from invenio_userprofiles.models import UserProfile
+
+from invenio_app_ils.acquisition.api import Order
+from invenio_app_ils.acquisition.proxies import current_ils_acq
+from invenio_app_ils.document_requests.api import DocumentRequest
+from invenio_app_ils.errors import PatronNotFoundError
+from invenio_app_ils.ill.api import BorrowingRequest
+from invenio_app_ils.ill.proxies import current_ils_ill
+from invenio_app_ils.patrons.api import (Patron, SystemAgent,
+                                         get_anonymous_patron_dict,
+                                         get_patron_or_unknown)
+
+from .acquisition.search import OrderSearch
+from .circulation.search import get_loans_by_patron_pid
+from .document_requests.search import DocumentRequestSearch
+from .ill.search import BorrowingRequestsSearch
+from .patrons.indexer import PatronIndexer
+from .proxies import current_app_ils
+
+
+def get_patron_activity(patron_pid):
+    """Get activity related to the given patron pid."""
+    if patron_pid is None:
+        raise ValueError("No patron pid was provided.")
+
+    patron = get_patron_or_unknown(patron_pid)
+    if not patron:
+        return None
+
+    def dump(search):
+        return [hit.to_dict() for hit in search.scan()]
+
+    patron_document_requests = dump(
+        DocumentRequestSearch().search_by_patron_pid(patron_pid)
+    )
+
+    patron_borrowing_requests = dump(
+        BorrowingRequestsSearch().search_by_patron_pid(patron_pid)
+    )
+
+    patron_acquisitions = dump(OrderSearch().search_by_patron_pid(patron_pid))
+
+    patron_loans = dump(get_loans_by_patron_pid(patron_pid))
+
+    patron_profile = UserProfile.get_by_userid(patron_pid).__dict__
+    del patron_profile["_sa_instance_state"]
+
+    patron_data = {
+        "patron": patron,
+        "profile": patron_profile,
+        "document_requests": patron_document_requests,
+        "borrowing_requests": patron_borrowing_requests,
+        "acquisitions": patron_acquisitions,
+        "loans": patron_loans,
+    }
+
+    return patron_data
+
+
+def anonymize_patron_data(patron_pid, force=False):
+    """Anonymize all the data/activity related to the given patron pid."""
+    if patron_pid is None:
+        raise ValueError("No patron pid was provided.")
+
+    patron = get_patron_or_unknown(patron_pid)
+    if not patron and not force:
+        return None
+
+    patron_object = None
+    try:
+        patron_object = current_app_ils.patron_cls.get_patron(patron_pid)
+    except PatronNotFoundError:
+        if not force:
+            raise PatronNotFoundError(patron_pid)
+
+    if (
+        get_loans_by_patron_pid(patron_pid)
+        .filter("term", state="ITEM_ON_LOAN")
+        .count()
+    ):
+        raise AssertionError(
+            "Cannot delete user %s: they have ongoing loans." % (patron_pid)
+        )
+
+    # Serialize empty patron values
+    anonymous_patron_fields = get_anonymous_patron_dict(patron_pid)
+
+    patron_loans = get_loans_by_patron_pid(patron_pid).execute()
+
+    indices = 0
+
+    for hit in patron_loans:
+        loan = Loan.get_record_by_pid(hit.pid)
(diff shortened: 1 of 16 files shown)


FlorianCassayre:
Left to do: cancel all loan requests for that user.

zzacharo (commented):
From a first look, it looks good!!! I think it would be nice to have some tests for this functionality :) This will consolidate the process of anonymizing the patron!

invenio_app_ils/anonymization.py
```diff
@@ -0,0 +1,177 @@
...
+    dropped = 0
+
+    with db.session.begin_nested():
+        ongoing_loans = 0
+        for loan in (
+            get_loans_by_patron_pid(patron_pid)
+            .filter("term", state="ITEM_ON_LOAN")
+            .scan()
+        ):
+            ongoing_loans += 1
```
zzacharo: Just out of curiosity, there wasn't any other way to retrieve the total amount of results?

jrcastro2: Yes! 
It can be retrieved using ES aggregations and it is faster.
It would look something like this:
```
loan_states = (
            current_app.config["CIRCULATION_STATES_LOAN_ACTIVE"]
        )
loans_result = get_loans_by_patron_pid(patron_pid, loan_states)[:0].execute()
for bucket in loans_result.aggregations.states.buckets:
    if bucket["key"] in current_app.config["CIRCULATION_STATES_LOAN_ACTIVE"]:
        ongoing_loans += bucket["doc_count"]
```
For the states it is better to retrieve it from the config.
And would also need to create a new function or change the `get_loans_by_patron_pid` to something like this:
```
def get_loans_by_patron_pid(patron_pid, states):
    """Returns all the loans (past and current) for a given patron."""
    search = current_circulation.loan_search_cls()\
        .filter("terms", state=states)\
        .filter("term", patron_pid=patron_pid)
    # Aggregation
    aggs = A("terms", field="state")
    search.aggs.bucket("states", aggs)
    return search
```

zzacharo: @jrcastro2 nice! Just a last question, are we sure with the aggregations all the results are being returned to that count? I know that scan is used when you want to retrieve at once more than 10000 results (I think that is the maximum result window). Also, another solution I was thinking is: Normal search and then just retrieve the total count, not sure if that has limitations in the number of results though...

FlorianCassayre: Thanks to you!
Regarding your first comment: we assumed that any given user could only have at most a few active loans at any given time, therefore our decision to scan the results instead of aggregating them.
We will still go with Javier's good suggestion. It does not matter much if the number of retrieved results is limited because it's probably not going to happen in practice, and even if it did the function would not break (we are only interested in knowing if the user has at least one active loan, the exact count is not really important).

invenio_app_ils/anonymization.py
```diff
@@ -0,0 +1,193 @@
...
+    # Serialize empty patron values
+    anonymous_patron_fields = get_anonymous_patron_dict(patron_pid)
+
+    patron_loans = get_loans_by_patron_pid(patron_pid).execute()
+
+    indices = 0
+
+    for hit in patron_loans:
+        loan = Loan.get_record_by_pid(hit.pid)
+        if loan["state"] == "PENDING":
```
kpsherva: it would be better to use (coming from invenio-circulation)
```
current_app.config["CIRCULATION_STATES_LOAN_REQUEST"] 
```
instead of hardcoding the state

invenio_app_ils/anonymization.py
```diff
@@ -0,0 +1,193 @@
...
+            params = deepcopy(loan)
+            params.update(
+                dict(
+                    cancel_reason="Loan cancelled to anonymize user.",
```
kpsherva:
```suggestion
                    cancel_reason="Loan request cancelled to anonymize user.",
```

invenio_app_ils/anonymization.py
```diff
@@ -0,0 +1,193 @@
...
+    for hit in patron_acquisitions:
+        acquisition = Order.get_record_by_pid(hit.pid)
+        acquisition.commit()
+        current_ils_acq.order_indexer.index(acquisition)
+        indices += 1
+
+    # Delete rows from db
+    dropped = 0
+
+    with db.session.begin_nested():
```
kpsherva, line 181: IMO this part should be implemented in separate function, f.e. `delete_user_account`, because it is different than `anonymize_patron_data`, so possibly they could be used separately if for some reason it is needed. In general it would be a separation of concerns

invenio_app_ils/permissions.py
```diff
@@ -117,6 +118,14 @@ def loan_extend_circulation_permission(loan):
     return PatronOwnerPermission(loan)
 
 
+def patron_owner_permission(record):
+    """Return permission to allow owner and librarian to access the record."""
+    if not has_request_context():
+        # allows performing the actions out of the request context f.e. CLI
```
kpsherva, line 124: for other reviewers: please share your thoughts on this solution

currently all of the permissions are blocking us of performing any actions (especially on circulation) programatically, because of `g.identity` (not existing when it is not a request)

invenio_app_ils/anonymization.py
FlorianCassayre: You're right, thanks; I would also like to point out that most of the existing code "hardcodes" the states (thus refactoring might be needed).

invenio_app_ils/anonymization.py
FlorianCassayre: Yes, however it is unlikely that we would need to delete the user rows while keeping the index out-of-date. The only use case I can identify is if we would like to anonymize their visible personal data while preserving their profile.
@KonstantinaStoikou what do you think?

invenio_app_ils/permissions.py
FlorianCassayre: An alternative artificial workaround would be to use a context when permission is required and manually set the `identity` property - but it's probably anti-pattern.

Edit: as of now, the function has an unpredictable behavior if called outside of the CLI (mainly because it depends on other function that rely on permissions). The above workaround would make it runnable regardless of the context. Thus my question: shall we add a `@need_permissions` to this function? (even though those seem to be reserved exclusively to views)

invenio_app_ils/anonymization.py
KonstantinaStoikou: I think it's better to have separate functions

invenio_app_ils/permissions.py
kpsherva: this function is meant to decide to return the permission, so there is no point of adding `@need_permissions`

the point was to check how safe is `has_request_context` as a deciding factor for which permission to return

ping @zzacharo

kpsherva: agreed with @zzacharo that it is a solution good enough for now

kpsherva (approved):

invenio_app_ils/patrons/api.py
```diff
@@ -123,12 +123,23 @@ def patron_exists(patron_pid):
     return User.query.filter_by(id=patron_pid).first() is not None
 
 
-def get_patron_or_empty_dict(patron_pid):
+def get_anonymous_patron_dict(patron_pid):
+    """Return dict with Unknown values for patron."""
+    return {
+        "id": "Unknown",
```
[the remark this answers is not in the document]
FlorianCassayre: You mean only for the `id`, right?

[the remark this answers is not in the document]
FlorianCassayre: → Change all values from `Unknown` to `Anonymous`.

PR #933, opened 2020-09-03
Loan editable dates

FlorianCassayre:
* new endpoint: `/api/circulation/loans/<pid>/update-dates`
* addresses inveniosoftware/react-invenio-app-ils#84

### DIFF
--- a/invenio_app_ils/circulation/api.py
+++ b/invenio_app_ils/circulation/api.py
@@ -8,15 +8,16 @@
 """Invenio App ILS Circulation APIs."""
 
 import uuid
-from copy import deepcopy
-from datetime import timedelta
+from copy import copy, deepcopy
+from datetime import date, timedelta
 from functools import partial
 
 from elasticsearch import VERSION as ES_VERSION
 from flask import current_app
 from flask_login import current_user
 from invenio_circulation.api import Loan
-from invenio_circulation.pidstore.minters import loan_pid_minter
+from invenio_circulation.config import (CIRCULATION_STATES_LOAN_ACTIVE,
+                                        CIRCULATION_STATES_LOAN_COMPLETED)
 from invenio_circulation.pidstore.pids import CIRCULATION_LOAN_PID_TYPE
 from invenio_circulation.proxies import current_circulation
 from invenio_circulation.search.api import search_by_patron_item_or_document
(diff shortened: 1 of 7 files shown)


invenio_app_ils/circulation/api.py
```diff
@@ -211,6 +213,44 @@ def checkout_loan(
...
+        today = date.today().strftime("%Y-%m-%d")
+        if request_start_date or request_expire_date:
+            raise IlsException(description="Cannot modify request dates of an active or completed loan.")
+        if start_date:
+            if start_date > today:
+                raise InvalidParameterError(description="Start date cannot be in the future for active loans.")
+            data["start_date"] = start_date
+        if end_date:
+            data["end_date"] = end_date
+        if data["end_date"] < data["start_date"]:
```
kpsherva, line 234: I would check this earlier before line 228, using the input params

invenio_app_ils/circulation/api.py
```diff
@@ -211,6 +213,44 @@ def checkout_loan(
...
+            raise InvalidParameterError(description="Negative date range.")
+    else:  # Pending or cancelled
+        if start_date or end_date:
+            raise IlsException(description="Cannot modify dates of a pending or cancelled loan.")
+        if request_start_date:
+            data["request_start_date"] = request_start_date
+        if request_expire_date:
+            data["request_expire_date"] = request_expire_date
+        if data["request_expire_date"] < data["request_start_date"]:
```
kpsherva, line 243: same here

invenio_app_ils/permissions.py
```diff
@@ -143,30 +143,7 @@ def views_permissions_factory(action):
...
-    elif action in is_backoffice_permission:
-        return backoffice_permission()
-    elif action in is_patron_owner_permission:
-        return PatronOwnerPermission
-    return deny_all()
-    is_backoffice_permission = [
-        "circulation-loan-checkout",
-        "circulation-loan-force-checkout",
-        "circulation-overdue-loan-email",
+        "circulation-loan-update-dates",
```
kpsherva, line 146: why the permissions below are deleted?

FlorianCassayre: Rollback from this [PR](https://github.com/inveniosoftware/invenio-app-ils/commit/b2303ec292d51935f7f83bb6d7be5298a320bb3f#diff-cdcb95e33636b177167fb105fe784e83R142); I don't know why it was written in the first place (failed c/p?).

invenio_app_ils/circulation/api.py
FlorianCassayre: Both fields are optional, we have to make sure that the final state is correct. I can hardly see how this could be simplified without modifying the function's behavior.

kpsherva (approved):

Issue #843, opened 2020-04-30
Closures module


ntarocco:
# TODO

* create user stories, similarly to what done [here](https://github.com/inveniosoftware/invenio-app-ils/issues/637)
* create tasks

## User stories

https://codimd.web.cern.ch/tXelz_qxQCyDs3OoRJTpNA

## Use cases from brainstorming

* Find the right name for this feature in the English dictionary and use it consistently. 
* Implement CRUD for:
  * opening/closing days and hours (hours just for display, not affecting circulation)
  * exceptions (allow single days or a period of days of opening or closing)
  * update the Location schema and edit form to store such information
* On the librarian changes opening/exceptions, a task should update all existing loans:
  * for active loans, change `end_date` or trigger an extension with a specific `end date`. The extension has the advantage of already sending an e-mail to the patron, but extensions have constraints to be checked (max n. of extension, book requested?)
  * for pending loans, change the `request_expire_date`
  * for ILLs (`item.pid_type: illbid`), it should be the same as normal loans
* Patrons should receive an email notification when the end date has been changed (already done if extension, please verify)
* Run a cron job to update the opening/closing days once a day when opening/exceptions are changed, so that eventually in case of mistake, the librarian can correct the opening/exceptions dates multiple times, without trigger immediate changes to loans.
* Create a dedicated front site page to render the opening/closing days and exceptions for each Location, fetched from the db. Nice to have: if logged in, show patron's Location as first.
* Opening/closing days are expsoed in the `/api/locations` endpoint, just change the permissions to be any user (currently only admins probably).
* The calendar React widget should be aware of the opening/closing days in the following places:
  * In the front site document page (Do you require it before a certain date?) Display based on patron location
  * backoffice document request a loan for document, display based on patron location
  * backoffice borrowing request in requested state when clicking new loan button, display based on patron location
* Endpoint cache: the REST endpoint is actually never cached on purpose, so limit the number of request from the React app if possible (fetch only once, maybe with a timeout?)

**Nice to have:**
* When changing opening/exceptions, the librarian should see a text mentioning how many loans are affected by the change before confirming. The link should clickable to open a new tab with a search on loans. e.g. `Changing these dates will affect _3 requests_ and _10 active loans_`

## Implementation (sketch)

https://codimd.web.cern.ch/CS7l855FTli6ilx-q-uhkA

## Extra

See: https://ils.test.rero.ch/

RERO-ILS implemented such a module, here their explanation:
```
[...]

PR #940, opened 2020-09-14
Closures module

FlorianCassayre:
* addresses #843


invenio_app_ils/locations/loaders/jsonschemas/location.py
```diff
@@ -21,3 +56,44 @@ class LocationSchemaV1(RecordMetadataSchemaJSONV1):
...
+    opening_weekdays = fields.List(fields.Nested(OpeningWeekdaySchema))
+    opening_exceptions = fields.List(fields.Nested(OpeningExceptionSchema))
+
+    @post_load
+    def postload_checks(self, data, **kwargs):
+        """Sort exceptions and validate record."""
+        record = self.context["record"]
+
+        # Bijection between weekdays and indices
+        weekday_names = ["monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday"]
```
ntarocco: I would have created a (fake) "enum" somewhere, to avoid the iteration each time:
```
WEEKDAYS = {
  "monday": 0,
  ...
}
```

invenio_app_ils/holidays/api.py
```diff
@@ -0,0 +1,84 @@
...
+        return exceptionally_open
+    return _is_normally_open(location, date)
+
+
+def next_open_after(location_pid, date):
+    """Finds the next day where this location is open."""
+    location = current_app_ils.location_record_cls.get_record_by_pid(location_pid)
+    n_weekdays = len(location["opening_weekdays"])
+    weekdays_closed = False
+    for exception in location["opening_exceptions"]:
```
kpsherva: lets say that `date="2020-09-10"`
our next opened day is `2020-09-11`
as I understood it, it will still go into this loop and iterate through all the exceptions even though the next day does not belong to an exception, correct me if I am wrong. in this case if we collect many exceptions during few years the loop will get more iterations and be slower, even though in normal cases it should not be an exception day

another thing is that it would not have to iterate through all the exceptions if it is not a weekday and not exceptionally opened

I still feel like there is too many `if`s in this approach

invenio_app_ils/holidays/api.py
```diff
@@ -0,0 +1,84 @@
...
+def is_open_on(location_pid, date):
+    """Checks if the location is open or closed on a given date."""
+    location = current_app_ils.location_record_cls.get_record_by_pid(location_pid)
+    exceptionally_open = _is_exceptionally_open(location, date)
+    if exceptionally_open is not None:
+        return exceptionally_open
+    return _is_normally_open(location, date)
+
+
+def next_open_after(location_pid, date):
```
ntarocco: I am sure this is the first attempt and needs refinement, if it can help, here my pseudocode:

```
def _date_in_range(date, location_exceptions):
    for ex in location_exceptions:
        if ex.start <= TODAY_CHECK <= ex.end:
            return True
    return False


def is_working_day(date, location_weekdays, location_exceptions_is_open, location_exceptions_is_closed):
  day_name = date.get_day_name
  weekday = location_weekdays[day_name]
  if weekday.is_open and not _date_in_range(date, location_exceptions_is_closed):
    return True
  if weekday.is_closed and _date_in_range(date, location_exceptions_is_open):
    return True
  return False
  

def _get_next_working_day(date, weekdays, exceptions_is_open, exceptions_is_closed):
    if is_working_day(date, weekdays, exceptions_is_open, exceptions_is_closed):
      return date
    else:
      return _get_next_working_day(date + timedelta(days=1), weekdays, exceptions_is_open, exceptions_is_closed)
  

def get_next_working_day(location, date):
  weekdays, exceptions_is_open, exceptions_is_closed = get_...(location)
  return _get_next_working_day(date, weekdays, exceptions_is_open, exceptions_is_closed)
```

The idea is:
```
while True (because at some point there must be a date available, otherwise the library is closed!)
  1. if date is normally open, then check if it is exceptionally closed
  2. if date is normally closed, then check if it is exceptionally open 
 date = date+1 day
```

invenio_app_ils/holidays/api.py
FlorianCassayre: You're totally right, that's why we decided to remove the exceptions that expired with a cron that is run everyday (currently what Javier is working on). Therefore in practice there will only be one (or more rarely a few more) iterations in the loop.
Otherwise we can decide to move to a dedicated schema to store the exceptions and then create a scanner that will iterate over all the exceptions that end _after today_ (one ES query). Assuming ES does a good job, the complexity would be similar.

invenio_app_ils/holidays/api.py
FlorianCassayre: Thanks for that!
It's more or less what RERO does, this procedure I wrote is essentially a more efficient implementation of that although I agree that will not make much of a difference. One must also take into account the special case where all the weekdays are being closed and the library only relies on opening exceptions (this is already handled in the current procedure).

invenio_app_ils/locations/loaders/jsonschemas/location.py
FlorianCassayre: :+1: for the enum, I forgot it existed.

invenio_app_ils/circulation/api.py
```diff
@@ -210,6 +212,14 @@ def checkout_loan(
         loan, **dict(params, trigger="checkout")
     )
 
+    current_end_date = arrow.get(loan["end_date"]).date()
+    new_end_date = next_open_after(loan["pickup_location_pid"], current_end_date)
+    if new_end_date and new_end_date != current_end_date:
+        loan.update(end_date=new_end_date.isoformat())
+        loan.commit()
+        db.session.commit()
+        current_app_ils.location_indexer.index(loan)
```
FlorianCassayre: Instead of updating/reindexing the record twice, I believe it would be better to define a new `Transition` class that inherits from `ToItemOnLoan` and executes this procedure in the `before` method (after calling `super().before`).

invenio_app_ils/config.py
```diff
@@ -177,6 +177,10 @@ def _(x):
         "task": "invenio_app_ils.circulation.mail.tasks.send_overdue_loans_mail_reminder",
         "schedule": timedelta(days=1),
     },
+    "extend_active_loans_holiday": {
```
ntarocco: I find this a bit dangerous maybe?
Probably better to discuss it IRL, but I would have instead probably triggered this new task (with a delay of 2/3 hours) after a change to the closures dates of a location, instead of running this every day.
Hopefully, the closures are updated very rarely...

invenio_app_ils/holidays/tasks.py
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Holidays tasks."""
```
ntarocco: Can you please add logging to both tasks here? It would be great to have JSON formatted logging as done in `ReferencedRecordsIndexer` class.
The idea is to be able to understand what happened in case of an error by checking logs:

- which past exceptions were automatically deleted? (if you get a support email asking "my previous exceptions disappeared, can you tell me which ones were there?")
- for which active loans we have automatically change the end date? What are the previous and the new end date? (imagine you get a support ticket asking "for the loan of Mr. Mario, the end date changed" and you need to find in the logs what happened)

invenio_app_ils/locations/loaders/jsonschemas/location.py
```diff
@@ -8,8 +8,53 @@
...
 
 from invenio_records_rest.schemas import RecordMetadataSchemaJSONV1
-from invenio_records_rest.schemas.fields import PersistentIdentifier
-from marshmallow import fields
+from invenio_records_rest.schemas.fields import (DateString,
+                                                 PersistentIdentifier)
+from marshmallow import (EXCLUDE, Schema, ValidationError, fields, post_load,
+                         pre_load, validates)
+
+_weekday_names = ["monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday"]
```
ntarocco: Uppercase? Since we are using this in cli.py too, what about putting it in common somewhere? e.g. in the closure module, in `api.py` or `constants.py`?

invenio_app_ils/locations/loaders/jsonschemas/location.py
```diff
@@ -21,3 +66,37 @@ class LocationSchemaV1(RecordMetadataSchemaJSONV1):
...
+        exists_open = False
+        for weekday in weekdays:
+            name = weekday["weekday"]
+            exists_open = exists_open or weekday["is_open"]
+            index = _weekday_indices[name]
+            if new_weekdays[index]:
+                raise ValidationError("Duplicate weekday.", field_names=["opening_weekdays"])
+            new_weekdays[index] = weekday
+        if len(weekdays) != len(new_weekdays):
+            raise ValidationError("Missing weekdays.", field_names=["opening_weekdays"])
```
ntarocco: Can you improve this error message?

invenio_app_ils/locations/loaders/jsonschemas/location.py
```diff
@@ -21,3 +66,37 @@ class LocationSchemaV1(RecordMetadataSchemaJSONV1):
...
+        exceptions = data["opening_exceptions"]
+        exceptions.sort(key=lambda ex: ex["start_date"])
+        previous = None
+        for exception in exceptions:
+            if previous:
+                if previous["end_date"] >= exception["start_date"]:
+                    raise ValidationError("Exceptions must not overlap.",
+                                          field_names=["opening_exceptions"])
+            previous = exception
+
```
ntarocco, line 161: here you could add an extra check, since you have the record in the context with the previous exceptions and the new exceptions and weekdays to check if there is at least a day open in the next 365 days, if you want (see my comment before for the while True).

tests/api/ils/test_holidays.py
```diff
@@ -0,0 +1,195 @@
+# -*- coding: utf-8 -*-
```
ntarocco: 👍 👏

invenio_app_ils/circulation/transitions/transitions.py
```diff
@@ -0,0 +1,42 @@
...
+from invenio_circulation.transitions.transitions import (
+    ItemOnLoanToItemOnLoan, ToItemOnLoan)
+
+from invenio_app_ils.holidays.api import next_open_after
+
+
+def _update_loan_end_date(loan):
+    """Update the end date to the next opening day."""
+    current_end_date = arrow.get(loan["end_date"]).date()
+    new_end_date = next_open_after(loan["pickup_location_pid"], current_end_date)
```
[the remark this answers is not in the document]
FlorianCassayre: The branch is indeed not needed anymore, it used to be in the case a new date would not be found (all weekdays closed).

invenio_app_ils/locations/loaders/jsonschemas/location.py
FlorianCassayre: (won't fix)

invenio_app_ils/holidays/api.py
```diff
@@ -0,0 +1,53 @@
...
+    """Finds the next day where this location is open."""
+    location = current_app_ils.location_record_cls.get_record_by_pid(location_pid)
+    while True:
```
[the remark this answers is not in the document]
FlorianCassayre: I will also sleep better with a loop guard. Moreover my assertion only holds under additional assumptions (ie. all the data was modified through REST).

invenio_app_ils/locations/loaders/jsonschemas/location.py
FlorianCassayre: (won't fix)

invenio_app_ils/holidays/tasks.py
```diff
@@ -0,0 +1,60 @@
...
+    """Deletes all past exceptions."""
+    today = date.today()
+    search_cls = current_app_ils.location_search_cls()
+    location = current_app_ils.location_record_cls
+    for hit in search_cls.scan():
+        index = 0
+        location_pid = hit.pid
+        record = location.get_record_by_pid(location_pid)
+        record_copy = deepcopy(record)
+        for item in record_copy["opening_exceptions"]:
```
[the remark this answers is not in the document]
FlorianCassayre: Agreed, would have done the same (also more efficient I think).

FlorianCassayre:
Compare `9829f6c` (old) and `0009332` (new).

FlorianCassayre:
Please review `8217205` (update active loans following a location update), after which it will integrated into `83fdc56`.

invenio_app_ils/circulation/config.py
```diff
@@ -9,13 +9,10 @@
...
 from invenio_circulation.api import Loan
 from invenio_circulation.pidstore.pids import (_LOANID_CONVERTER,
-                                               CIRCULATION_LOAN_FETCHER,
-                                               CIRCULATION_LOAN_MINTER,
                                                CIRCULATION_LOAN_PID_TYPE)
 from invenio_circulation.search.api import LoansSearch
 from invenio_circulation.transitions.transitions import (
-    CreatedToPending, ItemOnLoanToItemOnLoan, ItemOnLoanToItemReturned,
-    ToCancelled, ToItemOnLoan)
+    CreatedToPending, ItemOnLoanToItemReturned, ToCancelled)
```
ntarocco, line 15: you might have run `isort` here to fix imports

invenio_app_ils/circulation/transitions/transitions.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# Copyright (C) 2020 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Circulation custom transitions."""
+
+import arrow
+from invenio_circulation.transitions.transitions import (
```
ntarocco, line 11: isort?

invenio_app_ils/closures/api.py
```diff
@@ -0,0 +1,56 @@
...
+    """Finds the next day where this location is open."""
+    location = current_app_ils.location_record_cls.get_record_by_pid(location_pid)
+    _infinite_loop_guard = date + timedelta(days=365)
+    while date < _infinite_loop_guard:
+        if _is_open_on(location, date):
+            return date
+        date += _ONE_DAY_INCREMENT
+
+    # Termination is normally guaranteed if there is at least one weekday open
+    raise IlsException(description="Reached maximum iterations allowed to extend loan.")
```
ntarocco: Great! 👍 
Minor suggestion, I would also print the params:
```suggestion
    raise IlsException(description="Reached maximum iterations allowed to extend loan. Cannot find any date for which the location #{} is open after the given date {}. Please check opening/closures dates.".format(location["pid"], date.isoformat()))
```

invenio_app_ils/closures/api.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def _is_exceptionally_open(location, date):
+    """Checks if the location is exceptionally opened on a given date, or None if undefined."""
+    for exception in location["opening_exceptions"]:
+        if date <= arrow.get(exception["end_date"]).date():
+            if arrow.get(exception["start_date"]).date() <= date:
+                return exception["is_open"]
+            else:
+                return None  # Early return
```
ntarocco: I am not sure about this early return... isn't it working only if exceptions are sorted by `end_date` `ASC`?

Otherwise, I would do:
What about:
```
for exception in location["opening_exceptions"]:
    start_date = arrow.get(exception["start_date"]).date()
    end_date = arrow.get(exception["end_date"]).date()
    date_in_range = start_date <= date <= end_date
    if date_in_range:
        return exception["is_open"]
return False
```
I am also not sure why returning `None` instead of `False`

invenio_app_ils/closures/api.py
```diff
@@ -0,0 +1,56 @@
...
+    return None  # Exhaustion
+
+
+def _is_open_on(location, date):
+    """Checks if the location is open or closed on a given date."""
+    exceptionally_open = _is_exceptionally_open(location, date)
+    if exceptionally_open is not None:
```
ntarocco, line 46:
```suggestion
    if exceptionally_open:
        return True
```

invenio_app_ils/config.py
```diff
@@ -187,8 +188,15 @@ def _(x):
         "schedule": timedelta(hours=3),
         "args": [("record-view-agg", "file-download-agg")],
     },
+    "clean_locations_past_closures_exceptions": {
+        "task": "invenio_app_ils.closures.tasks.clean_locations_past_closures_exceptions",
+        "schedule": timedelta(weeks=4),
```
ntarocco: I am still doubting about this, ping other reviewers @kprzerwa @topless @KonstantinaStoikou.
Isn't it maybe weird that in June I cannot see the closures I had set up for last April or May? They will be deleted.
Wondering if we should run this once a year? Which it also sounds weird! :D

invenio_app_ils/locations/schemas/locations/location-v1.0.0.json
```diff
@@ -34,6 +34,110 @@
     "notes": {
       "type": "string",
       "title": "Extra notes"
+    },
+    "opening_weekdays": {
```
ntarocco, line 38: In the end, the last discussion we had, was it to move this to another record in the end or not?
The drawback is that if we change the name of the location, or anything else, the code will run... Even if it should not change any loan right?

invenio_app_ils/circulation/config.py
FlorianCassayre: I indeed ran `isort .` but note how the imports are different (one group comes from the circulation module while this other comes from the current project).

invenio_app_ils/closures/api.py
FlorianCassayre: The function can return three different results:
- `True | False`: an interval was found and its value is returned
- `None`: no interval was found

It's important to make the distinction between `False` and `None` because they should not be treated in the same way. More precisely if no interval matches, we should branch to the weekdays procedure.

invenio_app_ils/closures/api.py
FlorianCassayre: ([see this](https://github.com/inveniosoftware/invenio-app-ils/pull/940#discussion_r495882380))

invenio_app_ils/locations/schemas/locations/location-v1.0.0.json
FlorianCassayre: Yes, the code will run anytime a change is detected. It may be possible to detect what change has occurred using the signal, but personally I don't think it is worth it.

invenio_app_ils/closures/api.py
kpsherva: I find it  a bit dangerous, because we need to remember about it in all the conditional statements, which are to come based on this function. The question is if it is exceptionally open, intuition says the answer should be yes/no. Is the fact of not finding the exception changing anything? We still know that it is not exceptionally open.

FlorianCassayre: Because it can be either exceptionally open or exceptionally closed, in which case we would completely ignore the weekly schedule.

invenio_app_ils/ext.py
```diff
@@ -256,13 +257,15 @@ class InvenioAppIlsREST(InvenioAppIls):
...
-        self._register_signals()
+        self._register_signals(app)
         app.errorhandler(RESTException)(handle_rest_exceptions)
 
-    def _register_signals(self):
+    def _register_signals(self, app):
         """Register signals."""
         register_circulation_signals()
         register_files_signals()
+        if app.config["EXTEND_LOANS_LOCATION_UPDATED"]:
```
ntarocco, line 267: For the next PR, not this one, I would propose to remove the config var `EXTEND_LOANS_LOCATION_UPDATED` from the `config.py` since it is used only for tests and change this to:
```suggestion
        if app.config.get("EXTEND_LOANS_ON_LOCATION_UPDATED", True):
```
(see the rename `EXTEND_LOANS_ON_LOCATION_UPDATED`)

ntarocco (approved):

invenio_app_ils/closures/api.py
kpsherva: I still do not see the need for `None`. I think:
```
is_normally_open XOR is_exceptionally_closed XOR is_exceptionally_open 
```
covers all the cases, no?

PR #960, opened 2020-10-21
email: Adds overridable footer

jrcastro2:
Needed for https://github.com/CERNDocumentServer/cds-ils/pull/185


kpsherva:
Can you explain in a bit more details why the change was needed?

jrcastro2:
> Can you explain in a bit more details why the change was needed?

Sure ! Thanks for asking. Currently it's not possible to override the [footer.html](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/templates/invenio_app_ils/mail/footer.html) of `invenio-app-ils`. To be able to override it in `cds-ils`I just changed it (as it's done for the other `html` files) to make it overridable through the `ILS_GLOBAL_MAIL_TEMPLATES`, this way we can add [this](https://github.com/CERNDocumentServer/cds-ils/pull/185/files#diff-ae7be4653c3ffa44328860b92d374ec75aa361bf1561a7e5bc2ed9544d883eb0R464) in cds-ils and it overrides the footer for the emails.

kpsherva (approved):

invenio_app_ils/mail/messages.py
```diff
@@ -20,10 +20,16 @@
...
+
+    default_footer_template = dict(
+        footer="footer.html",
+    )
+
     def __init__(
         self,
         template,
-        footer_template="invenio_app_ils/mail/footer.html",
+        footer_template={},
```
ntarocco:
```suggestion
        footer_template="",
```

invenio_app_ils/mail/messages.py
```diff
@@ -50,7 +56,13 @@ def __init__(
             )
         )
         self.ctx = ctx
-
+        footer_template = dict(
```
ntarocco: Maybe better to keep it a string?
```suggestion
        footer_template = current_app.config.get("ILS_GLOBAL_MAIL_TEMPLATES", {})["footer"] or self.default_footer_template
```

invenio_app_ils/mail/messages.py
```diff
@@ -20,10 +20,16 @@
 class BlockTemplatedMessage(Message):
     """Templated message using Jinja2 blocks."""
 
+    templates_footer_base_dir = "invenio_app_ils/mail"
+
+    default_footer_template = dict(
```
ntarocco: string instead of a dict?

invenio_app_ils/mail/messages.py
```diff
@@ -59,7 +71,7 @@ def __init__(
         except TemplateError:
             kwargs["html"] = kwargs["body"]
 
-        footer_tmpl = current_app.jinja_env.get_template(footer_template)
+        footer_tmpl = current_app.jinja_env.get_template(footer_value)
```
ntarocco: it looks to me a little too complex maybe... As for the others already there, we could define in the config:
```
ILS_GLOBAL_MAIL_TEMPLATES = {
    "footer": "footer.html"
}
```

and then when using it, simply:
```
class BlockTemplatedMessage(Message):

    TEMPLATES_DIR = "invenio_app_ils/mail"

    def __init__(
        self,
        template,
        ctx={},
        **kwargs
    ):

    ...
    footer_template = current_app.config.get("ILS_GLOBAL_MAIL_TEMPLATES", {}).get("footer")
    if footer_template:
        ...
        footer_tmpl = current_app.jinja_env.get_template(os.path.join("invenio_app_ils/mail/", footer_template))
        ...
```
and not declare any default in this class

ntarocco:
Can you please also fix all the:
```
template="{}/{}".format(self.TEMPLATES_DIR, ...),
```
they should be:
```
template=os.path.join(self.TEMPLATES_DIR, ...),
```

invenio_app_ils/mail/messages.py
```diff
@@ -59,7 +60,8 @@ def __init__(
         except TemplateError:
             kwargs["html"] = kwargs["body"]
 
-        footer_tmpl = current_app.jinja_env.get_template(footer_template)
+        footer_template = current_app.config.get("ILS_GLOBAL_MAIL_TEMPLATES", {}).get("footer")
```
ntarocco, line 65: Just pay attention that when you use `dict.get` without default value, you might get `None`, so you need to handle that case:
```
footer_template = current_app.config.get(ILS_GLOBAL_MAIL_TEMPLATES, {}).get("footer")
if footer_template:
    footer_tmpl = current_app.jinja_env.get_template(os.path.join(self.GLOBAL_MAIL_TEMPLATES_BASE_DIR, footer_template))
    ... handle the footer ...
```
Watch out that footer is used below but it might None.

kpsherva (approved):

PR #969, opened 2020-10-26
vocabulary: Adds document related vocabularies

jrcastro2:
* Adds accelerator, experiments and
  standard_reviews vocabularies


jrcastro2:
Does it makes sense to add these vocabularies to invenio-app-ils ?

jrcastro2:
These vocabularies do not belong to invenio-app-ils, added in https://github.com/CERNDocumentServer/cds-ils/pull/199

PR #974, opened 2020-10-29
put back db.session.commit after transactions

ntarocco:
Put back the missing db.session.commit after a transaction. This is the only place that should be missing.

### DIFF
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -113,8 +113,6 @@ BABEL_DEFAULT_TIMEZONE = "Europe/Zurich"
 ###############################################################################
 #: Email address for support.
 SUPPORT_EMAIL = "info@inveniosoftware.org"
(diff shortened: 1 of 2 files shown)


invenio_app_ils/config.py
```diff
@@ -113,8 +113,6 @@ def _(x):
 ###############################################################################
 #: Email address for support.
 SUPPORT_EMAIL = "info@inveniosoftware.org"
-#: Management email for internal notifications.
-MANAGEMENT_EMAIL = "internal@inveniosoftware.org"
```
kpsherva, line 117: there were dependencies of this config var if I'm not wrong, but I don't see you removing any?

ntarocco: dead code, not used any more because the email has been moved to CDS-ILS.

kpsherva (approved):

PR #975, opened 2020-10-30
relations: validate missing language and edition fields

kpsherva:


ntarocco (commented):
Thank you for adding tests!

jrcastro2 (approved):
LGTM! Just a minor question, would make sense to also validate that the records we are relating don't have the same language or edition ?

ntarocco (approved):

Issue #980, opened 2020-11-05
The location opening hours do not work without lunch break


kpsherva:
When you try to save the opening hours without the lunch break the validation fails - it should be possible to have only one opening hour range

PR #981, opened 2020-11-06
location: relax constraint on opening hours

FlorianCassayre:
* addresses #980

**Before**: exactly two opening hours period had to be defined
**After**: there can be either one or two opening hours periods

This is just to set a frame for the backoffice form. The validation works for any number of time periods (though there should be a least one because it wouldn't make sense otherwise).

### DIFF
--- a/invenio_app_ils/locations/loaders/jsonschemas/location.py
+++ b/invenio_app_ils/locations/loaders/jsonschemas/location.py
@@ -89,11 +89,15 @@ class OpeningWeekdaySchema(Schema):
                     "Time periods must be defined on an opened weekday.",
                     "times",
                 )
-            if len(data["times"]) != 2:
-                raise ValidationError(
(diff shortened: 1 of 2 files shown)


FlorianCassayre:
(some tests may have to be updated, wait and see)

jrcastro2 (approved):

kpsherva (commented):
does it make sense to have more than 2 opening periods? Just theoretically, because the assumption with 2 was wrong, maybe there is a use case for 3?

ntarocco (approved):
Impressive how "easy" was to fix, so it was well done even before! 👏

FlorianCassayre:
@kprzerwa It would make sense to have more than 2, although rare to be found in practice. The main argument in favor of having at most 2 is to be able to inline the field in the backoffice. Otherwise we can very well use an array of fields, that works too.

PR #983, opened 2020-11-06
Global: change email tokenizer

### LINKED ISSUE #922: search: patron email tokenizer
## Description

When we are searching for patrons by email, elasticsearch uses the standard tokenizer for strings, and thus our search results  are exact matches. In patronApi we have added the `*` wildcard in our searches in order to match everything that starts with our search term, which is not the desired behaviour.

For the `email` field of the patron we need to use `uax_url_email` instead of the default. You can read more about it  https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-uaxurlemail-tokenizer.html

and there are a couple of answers in SO which provide greater in depth analysis on how it works

https://stackoverflow.com/questions/30115867/elasticsearch-analyzer-and-tokenizer-for-emails
https://stackoverflow.com/questions/38246436/elasticsearch-search-email-by-full-match (mostly 2nd answer)

KonstantinaStoikou:
Closes #922  (the wildcard from patronApi email is removed in [this PR)](https://github.com/inveniosoftware/react-invenio-app-ils/pull/275).

[Here](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-uaxurlemail-tokenizer.html) is the difference between the standard tokenizer and the `uax_url_email`.
The `uax_url_email` matches only the exact email address with the results, so for the email "john.doe@mail.com" the search john.doe or mail.com has no results.

Instead of defining settings in each mapping maybe we could use [elasticsearch index templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html).

Searches tested and working with the new tokenizer:
- checkout (search for patron by email)
- patrons search page
- libraries search page (search for library by email)
- vendors search page (search for vendor by email)
- borrowing request search (search by patron email)
- orders search page (search for vendor/patron by email)
- items search page (search by patron email)

Generally when you don't specify `email:` in the search it will tokenize the string and might return results if a field(except the email field) matches a token.

### DIFF
--- a/invenio_app_ils/internal_locations/jsonresolvers/internal_location.py
+++ b/invenio_app_ils/internal_locations/jsonresolvers/internal_location.py
@@ -12,9 +12,10 @@ from werkzeug.routing import Rule
 
 from invenio_app_ils.internal_locations.api import InternalLocation
 from invenio_app_ils.locations.api import Location
-from invenio_app_ils.records.jsonresolvers.api import \
-    get_field_value_for_record as get_field_value
-from invenio_app_ils.records.jsonresolvers.api import get_pid_or_default
+from invenio_app_ils.records.jsonresolvers.api import (
+    get_field_value_for_record as get_field_value,
+)
+from invenio_app_ils.records.jsonresolvers.api import get_pid_or_default, pick
 
 # Note: there must be only one resolver per file,
 # otherwise only the last one is registered
@@ -28,12 +29,18 @@ def jsonresolver_loader(url_map):
     @get_pid_or_default(default_value=dict())
     def location_resolver(internal_loc_pid):
         """Return the Location record for the given Internal Loc. or raise."""
-        location_pid = get_field_value(InternalLocation, internal_loc_pid,
-                                       "location_pid")
+        location_pid = get_field_value(
+            InternalLocation, internal_loc_pid, "location_pid"
+        )
         location = Location.get_record_by_pid(location_pid)
-        del location["$schema"]
 
-        return location
+        return pick(
+            location,
+            "name",
+            "opening_exceptions",
+            "opening_weekdays",
+            "pid",
+        )
 
     url_map.add(
         Rule(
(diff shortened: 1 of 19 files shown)


FlorianCassayre (approved):
Tested on my environment, looks good to me!
About the side-effect you mentioned, I honestly don't think there is any useful use case associated to this legacy feature.

invenio_app_ils/items/mappings/v6/items/item-v1.0.0.json
```diff
@@ -178,18 +189,6 @@
                 "name": {
                   "type": "keyword"
                 },
-                "address": {
```
kpsherva, line 181: for the reviewers: I requested the change, since it is highly unlikely that we will search items using email of the location and the other removed fields

kpsherva (approved):

invenio_app_ils/items/mappings/v6/items/item-v1.0.0.json
ntarocco: Ok, but then we should change the resolvers too? This is only removing the mappings. The will be still indexed, but ES will auto-guess the type.

ntarocco (commented):
Nice! It would be great if you demo the before and after this change and explain to everyone! :)

invenio_app_ils/items/mappings/v6/items/item-v1.0.0.json
kpsherva: true,  we shouldn't resolve those fields IMO, @KonstantinaStoikou could you remove the fields from the resolver too?

invenio_app_ils/internal_locations/jsonresolvers/internal_location.py
```diff
@@ -28,12 +28,18 @@ def jsonresolver_loader(url_map):
...
+        )
         location = Location.get_record_by_pid(location_pid)
-        del location["$schema"]
 
-        return location
+        return pick(
+            location,
+            "name",
+            "opening_exceptions",
+            "opening_weekdats",
```
ntarocco: Watch out, spelling issue here!
```suggestion
            "opening_weekdays",
```

ntarocco (approved):

PR #985, opened 2020-11-09
setup: Fixes relations from setup script

jrcastro2:
Currently if we run the setup script there are some chances that it will fail because it will try to create a relation between 2 documents with the same edition or language, this PR just prevents the script to fail in these cases.


KonstantinaStoikou (approved):

jrcastro2:
Already fixed in https://github.com/inveniosoftware/invenio-app-ils/pull/984

PR #990, opened 2020-11-16
records: split records db tables

ntarocco:
pgAdmin_4

There is still an issue with this to be solved before merging:

```
....lib/python3.6/site-packages/sqlalchemy/ext/declarative/clsregistry.py:129: SAWarning: This declarative base already contains a class with the same class name and module name as sqlalchemy_continuum.m
odel_builder.SeriesMetadataVersion, and will be replaced in the string-lookup table.
  % (item.__module__, item.__name__)

...lib/python3.6/site-packages/sqlalchemy/orm/properties.py:249: SAWarning: On mapper mapped class RecordMetadataVersion->records_metadata_version, primary key column 'records_metadata_version.transacti
on_id' is being combined with distinct primary key column 'records_metadata_version.transaction_id' in attribute 'transaction_id'. Use explicit properties to give each column its own mapped attribute name.
  % (self.parent, self.columns[1], self.columns[0], self.key)
```

Related to https://github.com/inveniosoftware/invenio-db/issues/96

### DIFF
--- a/invenio_app_ils/acquisition/api.py
+++ b/invenio_app_ils/acquisition/api.py
@@ -13,6 +13,7 @@ from flask import current_app
 from invenio_pidstore.models import PIDStatus
 from invenio_pidstore.providers.recordid_v2 import RecordIdProviderV2
 
+from invenio_app_ils.acquisition.models import OrderMetadata, VendorMetadata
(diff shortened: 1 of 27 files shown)


ntarocco:
Unfortunately we cannot put more effort to complete this PR.

PR #994, opened 2020-11-20
DocumentRequest: validate accepted requests

KonstantinaStoikou:
Validate if provider and document is given to a request with accepted status. 
The problem that occured before had to do with the demo data from cli: in the frontend the step to add a provider was not disabled (because the accepted requests didn't have a provider) and resulted in a server error when a provider was added.
Closes partially inveniosoftware/react-invenio-app-ils#290

### DIFF
--- a/invenio_app_ils/document_requests/api.py
+++ b/invenio_app_ils/document_requests/api.py
@@ -74,6 +74,16 @@ class DocumentRequestValidator(RecordValidator):
                 "without providing a document_pid."
             )
 
+    def validate_acceptance(self, document_pid, physical_item_provider, state):
+        """Validate rejection is correct."""
+        if state == "ACCEPTED" and (
+            not document_pid or not physical_item_provider
+        ):
+            raise DocumentRequestError(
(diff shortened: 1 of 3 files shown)


kpsherva (approved):
LGTM 🚀 needs one more reviewer who knows the use cases of the module better

invenio_app_ils/document_requests/api.py
```diff
@@ -74,6 +74,16 @@ def validate_rejection(self, document_pid, state, reject_reason):
                 "without providing a document_pid."
             )
 
+    def validate_acceptance(self, document_pid, physical_item_provider, state):
+        """Validate rejection is correct."""
+        if state == "ACCEPTED" and (
```
ntarocco, line 79: I think we should quickly chat maybe?
The endpoint is requiring both (doc and provider), see [here](https://github.com/inveniosoftware/invenio-app-ils/blob/0280d21e755b6f68c0e4f501a06f7539333bde64/invenio_app_ils/document_requests/views.py#L195) and it is not great that we split the validation.

ntarocco (approved):

PR #998, opened 2020-11-27
circulation: fix item availability

### LINKED ISSUE #966: circulation: incorrect item availability count with ILLs
When creating a borrowing request on a document that doesn't have any associated item, the field `circulation.available_items_for_loan_count` (on document) becomes `-1`:

**How to reproduce it:**
1. create borrowing request and set it to `Requested`
2. on the borrowing request details page create a loan `New loan`
3. the field on the document becomes -1

Please review how [available_items_for_loan_count](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/documents/jsonresolvers/document_circulation.py#L85) is computed

KonstantinaStoikou:
Closes #966 
Now it looks like this if there is an active ILL loan but the document doesn't have any physical copies in the library:
Screenshot 2020-12-14 at 11 03 35

### DIFF
--- a/invenio_app_ils/documents/jsonresolvers/document_circulation.py
+++ b/invenio_app_ils/documents/jsonresolvers/document_circulation.py
@@ -14,6 +14,7 @@ from werkzeug.routing import Rule
 from invenio_app_ils.circulation.search import (get_loan_next_available_date,
                                                 get_loans_aggregated_by_states,
                                                 get_overdue_loans_by_doc_pid)
+from invenio_app_ils.ill.api import BORROWING_REQUEST_PID_TYPE
 from invenio_app_ils.items.search import get_items_aggregated_by_statuses
 
 lt_es7 = ES_VERSION[0] < 7
@@ -43,8 +44,7 @@ def jsonresolver_loader(url_map):
         loans_search = get_loans_aggregated_by_states(
             document_pid, loan_states
         )
-        # No need for the loan hits
-        loans_search = loans_search[:0]
+
         loans_result = loans_search.execute()
 
         for bucket in loans_result.aggregations.states.buckets:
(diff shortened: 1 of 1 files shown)


invenio_app_ils/documents/jsonresolvers/document_circulation.py
```diff
@@ -82,8 +83,16 @@ def circulation_resolver(document_pid):
             if bucket["key"] in "FOR_REFERENCE_ONLY":
                 items_for_reference_count = bucket["doc_count"]
 
+        active_ill_loans = (
+            BorrowingRequestsSearch()
+            .get_active_ill_loans_for_document(document_pid)
+            .count()
+        )
```
FlorianCassayre: Looks good to me :+1:

I was wondering if it is possible to avoid an extra search, but it might not be reliable. Given an active loan, one of the only way we can know if it's an ILL loan is to check the metadata's `item_pid.type`: in case of an ILS item it would be `pitmid`, and otherwise `illbid`. So the idea would be to filter out the loans by their item before aggregating them, so we are sure we are dealing with loans on ILS items. (if a loan has no item we should not filter it out)

invenio_app_ils/documents/jsonresolvers/document_circulation.py
```diff
@@ -82,8 +82,20 @@ def circulation_resolver(document_pid):
             if bucket["key"] in "FOR_REFERENCE_ONLY":
                 items_for_reference_count = bucket["doc_count"]
 
+        active_ill_loans_count = 0
+
+        for hit in loans_result.hits:
```
FlorianCassayre, line 87: Note: the hits count will not exceed the window size, so it's possible that we will miss some. I don't remember how many results are returned at most (25? 10k?) but it might not ever happen in practice anyway because there are usually only a few items linked to the same document. In case that happens the`available_items_for_loan_count` could be negative, but unlikely, and not with a big impact.

jrcastro2 (approved):

invenio_app_ils/documents/jsonresolvers/document_circulation.py
kpsherva, line 87: you don't need to count the loans in the `for` loop. You could do something like (pseudocode):
```python

active_ill_loans_count = loans_search.filter(item_pid__type=BORROWING_REQUEST_PID_TYPE).count()
```
and the ES would do everything for you

FlorianCassayre: Actually it was myself who suggested - if possible - to get rid of the extra search, since it adds more delay to circulation. What do you think?

kpsherva: discussed IRL, agreed that this solution is more efficient than the extra search

kpsherva (approved):

PR #1000, opened 2020-12-17
Elasticsearch: boost fields

KonstantinaStoikou:
[CodiMD](https://codimd.web.cern.ch/TvrR3CJDTP-SZD75xC-sog)

### DIFF
--- a/invenio_app_ils/documents/search.py
+++ b/invenio_app_ils/documents/search.py
@@ -16,6 +16,15 @@ from invenio_app_ils.search_permissions import search_filter_record_permissions
 class DocumentSearch(RecordsSearch):
     """RecordsSearch for documents."""
 
+    boosted_fields = [
+        "title^8",
+        "author^6",
+        "imprint.publisher^4",
+        "edition^4",
+        "keywords^2",
+        "abstract^2",
+    ]
+
     class Meta:
         """Search only on documents index."""
 
--- a/invenio_app_ils/search_permissions.py
+++ b/invenio_app_ils/search_permissions.py
@@ -55,17 +55,19 @@ def search_filter_record_permissions():
     return Q("bool", filter=[combined_filter])
 
 
-def _ils_search_factory(self, search, validator):
+def ils_search_factory(self, search, validator=None):
     """Search factory with Query String validator.
 
     :param self: REST view.
     :param search: Elastic search DSL search instance.
     :returns: Tuple with search instance and URL arguments.
     """
+
     def query_parser(qstr=None):
         """Default parser that uses the Q() from elasticsearch_dsl."""
+        fields = getattr(search, "boosted_fields", []) + ["*"]
         if qstr:
-            return Q('query_string', query=qstr)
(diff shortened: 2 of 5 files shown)


invenio_app_ils/search_permissions.py
```diff
@@ -62,10 +62,15 @@ def _ils_search_factory(self, search, validator):
     :param search: Elastic search DSL search instance.
     :returns: Tuple with search instance and URL arguments.
     """
+
     def query_parser(qstr=None):
         """Default parser that uses the Q() from elasticsearch_dsl."""
         if qstr:
-            return Q('query_string', query=qstr)
+            return Q(
```
ntarocco: I think that the idea is good, now we should maybe make it applicable to all type of records.

In `elasticsearch-py`, in the search class, we have:
```
class Meta:
    ...
    fields = ('*', )
```
See it [here](https://github.com/inveniosoftware/invenio-search/blob/master/invenio_search/api.py#L77).
What happens if we that in each Meta of each each search class in ILS?
Can you check if they are taken into account when performing the search?

Otherwise, we could try to change the DocumentSearch class in `documents/search.py` (and the same for other records):
```
DocumentSearch():
    ...
    _boosted_fields = [...]
```

and then here, in this generic query parser, if the `search` var is the class above, something like:

```
if ...
    return Q("query_string",
                    query=qstr,
                    fields=search._boosted_fields or search.fields or ["*"],  # whatever makes sense as or
return Q(fields=...as above...)
```

I would avoid the copy/paste of the entire search_factory as much as possible.

invenio_app_ils/documents/search.py
```diff
@@ -16,6 +16,16 @@
...
     """RecordsSearch for documents."""
 
+    boosted_fields = [
+        "title^8",
+        "author^6",
+        "imprint.publisher^4",
+        "edition^4",
+        "keywords^2",
+        "abstract^2",
+        "*",
```
ntarocco: suggestion: can we move the `*` to the code that is using this? See below

invenio_app_ils/search_permissions.py
```diff
@@ -55,37 +55,44 @@ def search_filter_record_permissions():
...
 
     :param self: REST view.
     :param search: Elastic search DSL search instance.
     :returns: Tuple with search instance and URL arguments.
     """
+
     def query_parser(qstr=None):
         """Default parser that uses the Q() from elasticsearch_dsl."""
+        fields = (
+            search.boosted_fields
```
ntarocco: Suggestion (from comment above):
```
all_fields = ['*']
fields =  getattr(search, "boosted_fields", []) + all_fields
```
or simply
```
fields =  getattr(search, "boosted_fields", []) + ['*']
```

ntarocco (commented):
LGTM! 🚀 👍🏼

kpsherva (approved):
🚀 great job!!

ntarocco (approved):

PR #1005, opened 2021-01-07
relations: allow documents to have max one multipart monograph

### LINKED ISSUE #601: Document can have only one multipart monograph
Currently it is possible to add many multipart monographs to a document - which is incorrect. The `metadata.relations.multipart_monograph` should become an object.

KonstantinaStoikou:
Closes #601

### DIFF
--- a/invenio_app_ils/documents/schemas/documents/document-v1.0.0.json
+++ b/invenio_app_ils/documents/schemas/documents/document-v1.0.0.json
@@ -590,21 +590,18 @@
           }
         },
         "multipart_monograph": {
-          "items": {
-            "properties": {
-              "pid": {
-                "type": "string"
-              },
-              "pid_type": {
-                "type": "string"
-              },
-              "volume": {
-                "type": "string"
-              }
+          "properties": {
+            "pid": {
+              "type": "string"
             },
-            "type": "object"
+            "pid_type": {
+              "type": "string"
+            },
+            "volume": {
+              "type": "string"
+            }
           },
-          "type": "array"
+          "type": "object"
         },
         "other": {
           "items": {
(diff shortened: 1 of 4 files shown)


jrcastro2 (commented):
LGTM!

jrcastro2 (approved):

invenio_app_ils/documents/schemas/documents/document-v1.0.0.json
```diff
@@ -590,21 +590,18 @@
...
-              "pid": {
-                "type": "string"
-              },
-              "pid_type": {
-                "type": "string"
-              },
-              "volume": {
-                "type": "string"
-              }
+          "properties": {
```
kpsherva, line 593: did you check if it is breaking change for the frontend?

invenio_app_ils/records_relations/api.py
```diff
@@ -158,11 +158,20 @@ def _validate_relation_between_records(self, parent, child, relation_name):
...
 
-        if not (is_series_doc or is_serial_mm):
+        child_has_mm = False
+        if relation_type.name == "multipart_monograph":
+            pcr = ParentChildRelation(relation_type)
+            relations = pcr.get_relations_by_child(child.pid)
+            if len(relations) > 0:
+                child_has_mm = True
+
+        if not (is_series_doc or is_serial_mm) or child_has_mm:
```
kpsherva: if we check if child has mm on the beginning, we can avoid checking the rest of the conditions, but also provide more useful error message (f.e. `Record {pid} has already a multipart monograph`)
but it could also stay like this.
Please add a test case

invenio_app_ils/documents/schemas/documents/document-v1.0.0.json
KonstantinaStoikou: it doesn't break anything

tests/api/ils/documents/test_multipart_relations.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (C) 2021 CERN.
```
ntarocco: There are already tests for record relations, maybe to be moved there?

ntarocco (approved):

kpsherva (approved):

PR #1006, opened 2021-01-07
postFilters: Updates eitems "has files" filter

### LINKED ISSUE #22: fix filtering for e-items search page
* fix the access filter isSelected function
* implement has_files filter (only frontend,  preferably in the react-searchkit)

jrcastro2:
* Adds return statement for eitems with no files
* closes https://github.com/inveniosoftware/react-invenio-app-ils/issues/22

### DIFF
--- a/invenio_app_ils/facets.py
+++ b/invenio_app_ils/facets.py
@@ -48,16 +48,17 @@ def default_value_when_missing_filter(field, missing_val):
     return inner
 
 
-def not_empty_object_or_list_filter(field):
-    """Create a custom exists filter.
+def exists_value_filter(field, filter_value):
+    """Create a custom filter that filters by existing value.
 
     :param field: Field name.
-    :missing_val
-    :returns: Function that returns the Terms query.
+    :param filter_value: Filter value.
     """
     def inner(values):
-        return Bool(**{'must': {'exists': {'field': field}}})
-
+        if filter_value in values:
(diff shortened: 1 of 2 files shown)


invenio_app_ils/facets.py
```diff
@@ -48,16 +48,16 @@ def inner(values):
...
-    :missing_val
-    :returns: Function that returns the Terms query.
     """
     def inner(values):
-        return Bool(**{'must': {'exists': {'field': field}}})
-
+        if 'has_files' in values:
+            return Bool(**{'must': {'exists': {'field': field}}})
+        else:
+            return Bool(**{'must_not': {'exists': {'field': field}}})
```
kpsherva: we have the access to the field name from the outer function, so instead of making it specialised to only e-item, we can do
```suggestion
        if field in values:
            return Bool(**{'must': {'exists': {'field': field}}})
        else:
            return Bool(**{'must_not': {'exists': {'field': field}}})
```
and it stays as a generic empty/list filter.

jrcastro2: Yes! You are right, we can keep it as a generic filter, but we need to pass the filter value as an argument (since field value is `files.file_id`, see [here](https://github.com/inveniosoftware/invenio-app-ils/pull/1006/files#diff-5cb3b7e47a3af9c319af5e812e25d1297eab7f43aa4c9671be71a481b3fa8d9bR865))

ntarocco (approved):

PR #1010, opened 2021-01-13
ES: improves ES mappings

### LINKED ISSUE #355: Improve ElasticSearch index mappings
1. For each records, only a subset of fields should be in ES.
For example, for eitems, the urls field should probably not be indexed, there is no point to search for it.

2. Ensure "private" fields are not indexed, e.g. private notes

3. Give more weight to some fields: probably title should be the most relevant field when searching. This will improve relevance of search results.

4. Right now some indexes have sub-optimal mappings. In order to improve search results we should go over the mappings of each index to make sure they are optimal.

Discussion: Should we provide language optimized indexing using the [language analyzers](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html) in ES?

Some proposed changes:
- [x] documents-document-v1.0.0
  - `title`: multi-field with `text` and `keyword`
  - `abstracts`: `text`
  - `conference.series`: multifield
  - `keywords.value`: is it enough with `keyword` or would it be useful to search for parts of a keyword?
- [ ] items-item-v1.0.0
  - `document.*`: same as `documents-document-v1.0.0`
  - `internal_location.*`: should be optimized - uses dynamic mapping now
- [x] All vocabulary fields should probably be `keyword` only (e.g. `series.languages`)
- [x] Most names (e.g. author `full_name`, `alternative_names`, etc) might be useful to have multifields so we can efficiently search for the full name
- [x] `series.title` and `document.title`: multifield would be useful if we need to match by the full title

REMEMBER: about resolvers, in the resolved object rather pick keys than delete.

jrcastro2:
* title and name fields now have multifields
* adds conference_info to document mapping
* prevents some field from being indexed
* closes #355

### DIFF
--- a/invenio_app_ils/eitems/mappings/v6/eitems/eitem-v1.0.0.json
+++ b/invenio_app_ils/eitems/mappings/v6/eitems/eitem-v1.0.0.json
@@ -48,16 +48,31 @@
                       "type": "object"
                     },
                     "name": {
-                      "type": "keyword"
+                      "type": "text",
+                      "fields": {
+                        "keyword": {
+                          "type": "keyword"
+                        }
+                      }
                     }
                   },
                   "type": "object"
                 },
                 "alternative_names": {
-                  "type": "keyword"
+                  "type": "text",
+                  "fields": {
+                    "keyword": {
+                      "type": "keyword"
+                    }
+                  }
                 },
                 "full_name": {
-                  "type": "text"
+                  "type": "text",
+                  "fields": {
+                    "keyword": {
+                      "type": "keyword"
+                    }
+                  }
                 },
                 "identifiers": {
                   "properties": {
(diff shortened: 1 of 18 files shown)


kpsherva (approved):

invenio_app_ils/eitems/mappings/v6/eitems/eitem-v1.0.0.json
```diff
@@ -48,16 +48,31 @@
...
+                  "type": "text",
+                  "fields": {
+                    "keyword": {
+                      "type": "keyword"
+                    }
+                  }
                 },
                 "full_name": {
-                  "type": "text"
+                  "type": "text",
```
ntarocco, line 70: We can keep it like this, but in general this is useful when we want to create a search filter on a text field (you need a `keyword` type). I am not sure we will do such a thing for `names` in general...

ntarocco (approved):

invenio_app_ils/eitems/mappings/v6/eitems/eitem-v1.0.0.json
jrcastro2: Yes, apart from the search filter, it's also usefull for efficiently search for the full name

KonstantinaStoikou (approved):

PR #1015, opened 2021-01-22
legacy_id: Updates legacy_id type

jrcastro2:


KonstantinaStoikou (approved):

invenio_app_ils/ill/schemas/ill_libraries/library-v1.0.0.json
```diff
@@ -13,7 +13,7 @@
       "type": "string"
     },
     "legacy_id": {
-      "type": "string",
+      "type": ["array", "integer"],
```
ntarocco: Are we sure we want this? The loader accepts only one string, not even an int.
What is the need? We should really try to avoid to put in the generic ILS anything related to our migration...

jrcastro2: After merging libraries (as requested in this [google sheet](https://docs.google.com/spreadsheets/d/1UquTj5MrR8Nh3gYKOshFxbb6nJEg_hchmxDuW5PPA-A/edit#gid=440103590)) we will have both, arrays and strings as legacy_id, thats the motivation behind this change.
The libraries will look like this:
```
...
{
    "name": "Other", 
    "notes": "", 
    "phone": "", 
    "legacy_id": [42,66,65,67,70,71,69,48], 
    "address": "", 
    "type": "external", 
    "email": ""
  }, 
  {
    "name": "Legal Service Library", 
    "notes": "", 
    "phone": "", 
    "legacy_id": 3, 
    "address": "", 
    "type": "internal", 
    "email": ""
  }, 
...
```
I removed it from the loader for libraries (since it was not present in the loader in internal_locations either) As I undersand it, we don't modify this value, it's only used by the migration at importing time.
I agree with keeping the migration away from generic ILS, but just having the legacy_id is already part of the migration, or not?

kpsherva: I've checked, and the legacy_id was added because the schema is strict, lets discuss in person ?

ntarocco (approved):

kpsherva (approved):

kpsherva (commented):
what about the items, for consistency?
I know our use case doe not have them, but maybe it makes sense to keep the same type: array for all

jrcastro2:
> what about the items, for consistency?
> I know our use case doe not have them, but maybe it makes sense to keep the same type: array for all

Yes, but if I am not mistaken, that would imply to also change the legacy_id from ACQ orders, document request and borrowing request (apart from the items). Should we go with the change ?

kpsherva (approved):

PR #1017, opened 2021-01-25
global: update invenio packages to 3.4

kpsherva:
* change IlsRecord validate method signature
* fix pycodestyle test
* update isort settings
* fix sqlite test

for reviewers:  don't worry, I will tag the real changes with comments :)

### DIFF
--- a/docker-services.yml
+++ b/docker-services.yml
@@ -70,6 +70,7 @@ services:
       - bootstrap.memory_lock=true
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
       - discovery.type=single-node
+      - cluster.routing.allocation.disk.threshold_enabled=false
     ulimits:
       memlock:
         soft: -1
--- a/invenio_app_ils/acquisition/api.py
+++ b/invenio_app_ils/acquisition/api.py
@@ -75,7 +75,7 @@ class Order(IlsRecord):
     _vendor_resolver_path = (
         "{scheme}://{host}/api/resolver/acquisition/orders/{order_pid}/vendor"
     )
-    _order_lines_resolver_path = "{scheme}://{host}/api/resolver/acquisition/orders/{order_pid}/order-lines"
+    _order_lines_resolver_path = "{scheme}://{host}/api/resolver/acquisition/orders/{order_pid}/order-lines"  # noqa
 
     STATUSES = ["PENDING", "ORDERED", "RECEIVED", "CANCELLED"]
 
--- a/invenio_app_ils/document_requests/mail/messages.py
+++ b/invenio_app_ils/document_requests/mail/messages.py
@@ -20,7 +20,7 @@ class DocumentRequestMessage(BlockTemplatedMessage):
 
     DEFAULT_TEMPLATES = dict(
         request_accepted="document_request_accept.html",
-        request_rejected_user_cancel="document_request_reject_user_cancel.html",
+        request_rejected_user_cancel="document_request_reject_user_cancel.html",  # noqa
         request_rejected_in_catalog="document_request_reject_in_catalog.html",
         request_rejected_not_found="document_request_reject_not_found.html",
     )
--- a/invenio_app_ils/items/api.py
+++ b/invenio_app_ils/items/api.py
@@ -12,13 +12,17 @@ from functools import partial
 from elasticsearch import VERSION as ES_VERSION
 from flask import current_app
 from invenio_circulation.search.api import search_by_pid
-from invenio_pidstore.errors import (PersistentIdentifierError,
-                                     PIDDoesNotExistError)
+from invenio_pidstore.errors import (
+    PersistentIdentifierError,
+    PIDDoesNotExistError,
+)
 from invenio_pidstore.models import PIDStatus
 from invenio_pidstore.providers.recordid_v2 import RecordIdProviderV2
 
-from invenio_app_ils.errors import (ItemDocumentNotFoundError,
-                                    ItemHasPastLoansError)
+from invenio_app_ils.errors import (
+    ItemDocumentNotFoundError,
+    ItemHasPastLoansError,
+)
 from invenio_app_ils.fetchers import pid_fetcher
 from invenio_app_ils.minters import pid_minter
 from invenio_app_ils.records.api import IlsRecord, RecordValidator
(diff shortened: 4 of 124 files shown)


docker-services.yml
```diff
@@ -70,6 +70,7 @@ services:
       - bootstrap.memory_lock=true
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
       - discovery.type=single-node
+      - cluster.routing.allocation.disk.threshold_enabled=false
```
kpsherva, line 73: on my local machine ES container was overloaded, I had to add this config

invenio_app_ils/acquisition/api.py
```diff
@@ -75,7 +75,7 @@ class Order(IlsRecord):
     _vendor_resolver_path = (
         "{scheme}://{host}/api/resolver/acquisition/orders/{order_pid}/vendor"
     )
-    _order_lines_resolver_path = "{scheme}://{host}/api/resolver/acquisition/orders/{order_pid}/order-lines"
+    _order_lines_resolver_path = "{scheme}://{host}/api/resolver/acquisition/orders/{order_pid}/order-lines"  # noqa
```
kpsherva, line 78: it fails if line breaks

invenio_app_ils/document_requests/mail/messages.py
```diff
@@ -20,7 +20,7 @@ class DocumentRequestMessage(BlockTemplatedMessage):
 
     DEFAULT_TEMPLATES = dict(
         request_accepted="document_request_accept.html",
-        request_rejected_user_cancel="document_request_reject_user_cancel.html",
+        request_rejected_user_cancel="document_request_reject_user_cancel.html",  # noqa
```
kpsherva, line 23: does anyone have a better solution? it was failing only on the comma...

invenio_app_ils/items/api.py
```diff
@@ -57,20 +61,29 @@ def ensure_document_exists(self, document_pid):
     def ensure_item_can_be_updated(self, record):
         """Raises an exception if the item's status cannot be updated."""
         latest_version = record.revisions[-1]
+        document_changed = False
```
kpsherva, line 64: it was used before assignment

invenio_app_ils/patrons/api.py
```diff
@@ -166,6 +166,6 @@ def get_patron_or_unknown_dump(patron_pid):
     try:
         Patron = current_app_ils.patron_cls
         return Patron.get_patron(patron_pid).dumps_loader()
-    except:
+    except Exception as e:
```
kpsherva, line 169: bare except is a bad practice

invenio_app_ils/records/api.py
```diff
@@ -79,11 +79,11 @@ def clear(self):
         super().clear()
         self["$schema"] = current_jsonschemas.path_to_url(self._schema)
 
-    def validate(self, **kwargs):
+    def _validate(self, **kwargs):
```
kpsherva, line 82: signature change because of latest changes in invenio-records

invenio_app_ils/records/api.py
```diff
@@ -96,3 +96,4 @@ def validate(self, **kwargs):
         # Custom record validation
         if self._validator:
             self._validator.validate(self, **kwargs)
+        return json
```
kpsherva, line 99: _validate must return json from now on, see https://github.com/inveniosoftware/invenio-records/blob/v1.4.0/invenio_records/api.py#L173

invenio_app_ils/records_relations/api.py
```diff
@@ -84,10 +89,13 @@ def remove_extra_metadata_from(
         """Remove any presence of the given PID in extra metadata."""
         field = cls.field_name()
         if field in record and relation_name in record[field]:
-            keep_pid_func = lambda m: not (
-                m.get("pid_value", "") == pid_value
-                and m.get("pid_type", "") == pid_type
-            )
+            def keep_pid_func(m):
```
kpsherva, line 92: it was marked by pep8 as a bad practice to assign lambda

pytest.ini
```diff
@@ -7,4 +7,4 @@
 
 [pytest]
 addopts = --isort --pydocstyle --pycodestyle --doctest-glob="*.rst" --doctest-modules --cov=invenio_app_ils --cov-report=term-missing --cov-append
-testpaths = docs tests invenio_app_ils
+testpaths = docs invenio_app_ils
```
kpsherva, line 10: change to add the pydocstyle tests

setup.cfg
```diff
@@ -20,10 +20,17 @@ universal = 1
 add_ignore = D401
 
 [pycodestyle]
-exclude = docs/conf.py
+exclude = docs/conf.py, invenio_app_ils/config.py
```
kpsherva, line 23: easier than adding # noqa  in every string module path

setup.cfg
```diff
@@ -20,10 +20,17 @@ universal = 1
...
 
 [isort]
 skip_glob=node_modules
+# black compatible sorting
```
kpsherva, line 27: with this settings, no more black/isort conflicts, WDYT?

setup.py
```diff
@@ -58,22 +64,25 @@
 
 install_requires = [
     # --- Invenio ----------------------------------------------------------
-    "invenio[base,auth]>=3.3.0,<3.4",
+    "invenio[base,auth]>=3.4.0,<3.5",
```
kpsherva, line 67: 3.4 updates here and below

tests/conftest.py
```diff
@@ -25,7 +25,10 @@ def users(app, db):
     """Create admin, librarians and patrons."""
     # with Postgresql, when dropping the User table, the sequence is not
     # automatically reset to 1, causing issues with the tests demo data.
-    db.session.execute("ALTER SEQUENCE IF EXISTS accounts_user_id_seq RESTART")
+    if db.engine.driver != "pysqlite":
```
kpsherva, line 28: I had to add this, the tests were failing on the incorrect syntax for sqlite

run-tests.sh
```diff
@@ -31,6 +31,7 @@ trap cleanup EXIT
 
 python -m check_manifest --ignore ".*-requirements.txt"
 python -m sphinx.cmd.build -qnNW docs docs/_build/html
+python -m pytest
```
ntarocco, line 34: I would add a comment here to explain that, without the path to the tests folder, it is simply checking style in docs and code folders

setup.cfg
ntarocco: true, but the config.py is a massive file... can we simply exclude the line-length in that file?

setup.cfg
ntarocco: What about [this](https://pycqa.github.io/isort/docs/configuration/black_compatibility/) instead?

tests/conftest.py
ntarocco: Yes, but tests should not run on `sqllite`... why do you need it?

ntarocco (approved):
Good job! Thanks for doing it! 🚀 
I would propose to merge this and the take comments in another PR maybe? WDYT?

tests/conftest.py
kpsherva: I thought they suppose to be running on sqlite too, in this case something is wrong with the tests configuration, because it runs sqlite for me and it raises the syntax exception

PR #1021, opened 2021-01-27
loans: available items filter on search

### LINKED ISSUE #804: Loans search: Filter by items available for loan
@topless @ntarocco  this was one of the requirements, it made a lot of sense to have this filter, so I think we should discuss about alternative solutions

_Originally posted by @kprzerwa in https://github.com/inveniosoftware/invenio-app-ils/pull/801_

Nice to have, from old requirements: combined filter of  'has items available for loan` and 'has ebook` described as `Immediate availability` filter

kpsherva:
* closes #804

### DIFF
--- a/invenio_app_ils/circulation/indexer.py
+++ b/invenio_app_ils/circulation/indexer.py
@@ -12,6 +12,7 @@ from datetime import datetime
 from celery import shared_task
 from flask import current_app
 from invenio_circulation.pidstore.pids import CIRCULATION_LOAN_PID_TYPE
+from invenio_circulation.proxies import current_circulation
 from invenio_indexer.api import RecordIndexer
 
 from invenio_app_ils.circulation.utils import resolve_item_from_loan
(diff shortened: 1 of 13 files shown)


invenio_app_ils/circulation/indexer.py
```diff
@@ -37,7 +38,14 @@ def index_referenced_records(loan):
         item = resolve_item_from_loan(loan["item_pid"])
         referenced.append(dict(pid_type=loan["item_pid"]["type"], record=item))
 
-    # index the document
+    # index the loan itself
```
ntarocco: IMHO this comment is not enough, in one month I will be probably asking myself why we are doing this :)
```suggestion
    # index the loan itself: this is needed because of the extra field `available_items_for_loan_count`
    # added when indexing. To calculate the value of this field, a search on the `loans`
    # indexed is performed and this loan has to be already indexed with its latest data.
    # At the first indexing, `available_items_for_loan_count` value might be wrong and corrected
    # at the second re-indexing. 
```

invenio_app_ils/circulation/indexer.py
```diff
@@ -49,3 +57,19 @@ def index(self, loan, arguments=None, **kwargs):
         super().index(loan)
         eta = datetime.utcnow() + current_app.config["ILS_INDEXER_TASK_DELAY"]
         index_referenced_records.apply_async((loan,), eta=eta)
+
+
+def index_item_availability_for_loan(loan_dict):
+    """Indexer hook to modify the loan record dict before indexing."""
```
ntarocco:
```suggestion
"""Indexer hook to modify the loan record dict before indexing.

     The `available_items_for_loan_count` field is added to the loan only on the search index
     because it is needed for search aggregation/filtering. It is not needed when fetching the
     loan details.
"""
```

ntarocco (commented):
Can you please improve the comments? :)

tests/conftest.py
```diff
@@ -25,10 +25,9 @@ def users(app, db):
     """Create admin, librarians and patrons."""
     # with Postgresql, when dropping the User table, the sequence is not
     # automatically reset to 1, causing issues with the tests demo data.
-    if db.engine.driver != "pysqlite":
-        db.session.execute(
-            "ALTER SEQUENCE IF EXISTS accounts_user_id_seq RESTART"
-        )
+    db.session.execute(
```
kpsherva, line 28: @ntarocco it worked in the end as it was before, I probably had a wrong configuration locally

ntarocco (approved):

PR #1024, opened 2021-02-05
schema: physical_description

jrcastro2:
* Deletes physical_description from item schema
* Adds physical_description to document schema
* Requested in for https://github.com/CERNDocumentServer/cds-dojson/issues/259

### DIFF
--- a/invenio_app_ils/documents/schemas/documents/document-v1.0.0.json
+++ b/invenio_app_ils/documents/schemas/documents/document-v1.0.0.json
@@ -728,6 +728,10 @@
       "minItems": 1,
       "type": "array",
       "uniqueItems": true
+    },
+    "physical_description": {
+      "title": "Physical description of a document",
+      "type": "string"
     }
   }
 }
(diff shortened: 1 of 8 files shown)


kpsherva (approved):

invenio_app_ils/documents/schemas/documents/document-v1.0.0.json
```diff
@@ -728,6 +728,10 @@
       "minItems": 1,
       "type": "array",
       "uniqueItems": true
+    },
+    "physical_description": {
```
ntarocco, line 732: :( this should ideally go to the Item and not here, but if this change is needed...
We will have also to add this to the React-JSONSchema form schema.

kpsherva: the problem is that if we have `physical_description` it will be copied over each instance of item but in the same time it is more flexible if one of the items is different than the others, and I was actually rethinking what is better. Conceptually it seems that each item should have its own physical_description. WDYT @ntarocco ?

PR #1028, opened 2021-02-08
search factory: validate correct search instance

kpsherva:


KonstantinaStoikou (approved):

invenio_app_ils/search_permissions.py
```diff
@@ -63,11 +63,18 @@ def ils_search_factory(self, search, validator=None):
...
+    def query_parser(search, qstr=None):
         """Default parser that uses the Q() from elasticsearch_dsl."""
-        fields = getattr(search, "boosted_fields", []) + ["*"]
         if qstr:
-            return Q("query_string", query=qstr, fields=fields)
+            boosted = getattr(search, "boosted_fields", [])
+            extra_params = {}
+            if boosted:
+                extra_params["fields"] = boosted + ["*"]
+                # add lenient parameter in order to fix
```
ntarocco, line 73: @KonstantinaStoikou even better this comment for `invenio-records-rest`...

invenio_app_ils/search_permissions.py
```diff
@@ -63,11 +63,18 @@ def ils_search_factory(self, search, validator=None):
     :returns: Tuple with search instance and URL arguments.
     """
 
-    def query_parser(qstr=None):
+    def query_parser(search, qstr=None):
```
ntarocco, line 66: You can eventually use the `search` param from outside... not really important
```suggestion
    def query_parser(qstr=None):
```

ntarocco (approved):

invenio_app_ils/search_permissions.py
kpsherva: I changed it because it passes through a validator before query_parser is even called and basically validator can return changed search instance

PR #1032, opened 2021-02-12
extensions: facet

jrcastro2:
* Adds extensions facet
* Allows indexing of extensions

### DIFF
--- a/invenio_app_ils/documents/mappings/v6/documents/document-v1.0.0.json
+++ b/invenio_app_ils/documents/mappings/v6/documents/document-v1.0.0.json
@@ -259,9 +259,7 @@
           "type": "object"
         },
         "extensions": {
-          "type": "object",
-          "dynamic": false,
-          "enabled": false
+          "type": "object"
         },
         "extensions_keywords": {
           "type": "object",
--- a/invenio_app_ils/documents/mappings/v7/documents/document-v1.0.0.json
+++ b/invenio_app_ils/documents/mappings/v7/documents/document-v1.0.0.json
@@ -258,9 +258,7 @@
         "type": "object"
       },
       "extensions": {
-        "type": "object",
-        "dynamic": false,
-        "enabled": false
+        "type": "object"
       },
       "extensions_keywords": {
         "type": "object",
--- a/invenio_app_ils/records/metadata_extensions.py
+++ b/invenio_app_ils/records/metadata_extensions.py
@@ -60,10 +60,7 @@ class MetadataExtensions(object):
             contexts.add(settings["@context"])
 
         # Validate extensions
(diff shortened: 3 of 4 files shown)


invenio_app_ils/config.py
```diff
@@ -831,6 +832,9 @@ def _(x):
             ),
             medium=dict(terms=dict(field="stock.mediums")),
         ),
+        filters=dict(
+            extensions=custom_metadata_filter('extensions'),
```
ntarocco: remove?

invenio_app_ils/documents/facets.py
```diff
@@ -0,0 +1,94 @@
...
+"""ILS Document indexer APIs."""
+
+import re
+
+from elasticsearch_dsl import Q
+from flask import current_app
+from invenio_rest.errors import FieldError, RESTValidationError
+
+
+def custom_metadata_filter(field):
```
ntarocco: is this to be removed after our chat?  
I am sorry if you spent a lot of time on this code :(

invenio_app_ils/documents/mappings/v6/documents/document-v1.0.0.json
```diff
@@ -259,9 +259,7 @@
           "type": "object"
         },
         "extensions": {
-          "type": "object",
-          "dynamic": false,
```
ntarocco, line 263: I would probably have left `dynamic: false`... any particular reason why you have removed it?

ntarocco (commented):
Did you check if there is nothing to change in the React app when rendering the custom fields, both in the details page (frontsite and backend) and also in the edit/create form?
The `:` might be hardcoded somewhere...

invenio_app_ils/documents/mappings/v6/documents/document-v1.0.0.json
jrcastro2: If we keep `dynamic: false` we cannot search for the extensions in the search-box. I didn't check further.

jrcastro2:
> Did you check if there is nothing to change in the React app when rendering the custom fields, both in the details page (frontsite and backend) and also in the edit/create form?
> The `:` might be hardcoded somewhere...

I just double checked and it is working correctly both, frontsite and backoffice, but I noticed that there are 2 fields, [`institution`](https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/config.py#L439-L441) and [`study`](https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/config.py#L447-L450) that are displayed as text but expect a list. (This values are not even migrated as they do not exist in legacy data if I am not mistake. Also this "bug" is not motivated by this change).

image

ntarocco (approved):

PR #1037, opened 2021-02-19
emails: create email log db table and entrypoints

### LINKED ISSUE #523: Overdue loan reminder: show last sent
Show in the UI last sent email manual or automatic so that the librarian can know when it was last sent.

## Solution

Sent emails (not the body) will be stored in the database. The UI will fetch data from a new API endpoint.

**DB table cols:**

| col name | type | description |
|-|-|-|
| id | number | autoincrement primary key |
| message_id | UUID | the UUID of the mail message |
| triggered_on | datetime | when the email was sent |
| triggered | string | if the email was triggered by a librarian (manually) or a celery task (automatically) |
| recipient_user_id | number | a real user id, the type of the col should be the same as the one in the accounts tables |
| email_action | string | keep track of the action that triggered the email, e.g. checkout, checkin |
| send_log | long string | information from log_successful_mail/log_error_mail (they are the celery task callback functions) |
| loan_pid| string | the pid of the loan if the email is related to a loan (optional) |

**Tasks**

- [x] create db table with the fields above
- [x] create admin panel display
- [x] anonymize rows in the db table when users are anonymized
- [x] create api endpoint with list/item views:
  - [x] list: `/api/emails`
  - [x] list by user: `/api/emails?recipient_user_id=<>` and by loan_pid and email_action
  - [x] item: `/api/emails/<id>`
- [x] test endpoint protection (admins/librarians should have access)
- [x] display in frontend backoffice how many reminders were sent for each overdue loan

KonstantinaStoikou:
* add item/list entrypoint for emails (for list view: recipient_user_id, loan_pid and email_action can be given as parameters)
* add admin panel for email logs
* anonymization: 
    * anonymize patron recipients in emails
    * loans stop sending cancellation emails
    * delete rows from transaction table
    * fix bug in `anonymize_patron_data` because it was trying to cancel completed loans
    * fix anonymization of orders and fail anonymization if there are pending orders

Closes #523

### DIFF
--- a/invenio_app_ils/circulation/receivers.py
+++ b/invenio_app_ils/circulation/receivers.py
@@ -54,5 +54,7 @@ def index_after_loan_replace_item(_, old_item_pid, new_item_pid):
 def send_email_after_loan_change(_, initial_loan, loan, trigger):
     """Send email notification when the loan changes."""
     send_loan_mail(
-        action=trigger, loan=loan, message_ctx=dict(initial_loan=initial_loan)
+        action=trigger,
+        loan=loan,
+        message_ctx=dict(initial_loan=initial_loan),
     )
--- a/invenio_app_ils/document_requests/mail/messages.py
+++ b/invenio_app_ils/document_requests/mail/messages.py
@@ -10,6 +10,7 @@ import os
 
 from flask import current_app
 
+from invenio_app_ils.document_requests.api import DOCUMENT_REQUEST_PID_TYPE
 from invenio_app_ils.mail.messages import BlockTemplatedMessage
 
 
@@ -28,6 +29,7 @@ class DocumentRequestMessage(BlockTemplatedMessage):
     def __init__(self, request, action=None, message_ctx={}, **kwargs):
         """Create an e-mail message based on the new doc request record."""
         self.request = request
+        self.action = action
 
         templates = dict(
             self.DEFAULT_TEMPLATES,
@@ -56,5 +58,7 @@ class DocumentRequestMessage(BlockTemplatedMessage):
     def dump(self):
         """Dump document request email data."""
         data = super().dump()
-        data["document_request_pid"] = self.request["pid"]
+        data["pid_value"] = self.request["pid"]
+        data["pid_type"] = DOCUMENT_REQUEST_PID_TYPE
+        data["action"] = self.action
         return data
--- a/invenio_app_ils/mail/admin.py
+++ b/invenio_app_ils/mail/admin.py
@@ -1,0 +1,57 @@
+# Copyright (C) 2021 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Admin panel for sent emails."""
+
+from flask import Blueprint
+from flask_admin.contrib.sqla import ModelView
+from flask_babelex import gettext as _
+
+from invenio_app_ils.mail.models import EmailLog
+
+
+class EmailLogModelView(ModelView):
+    """Invenio admin view for sent emails."""
+
+    column_list = (
+        "id",
+        "is_manually_triggered",
+        "recipient_user_id",
+        "triggered_on",
+        "send_log",
+        "message_id",
+        "email_action",
+        "pid_type",
+        "pid_value",
+    )
+
+    # Entries are read-only
+    can_create = False
+    can_edit = False
+    can_delete = False
+
+    can_view_details = True
+
+    form_excluded_columns = EmailLog.__table__.columns
+    column_default_sort = (EmailLog.triggered_on, True)
+
+    column_filters = column_list
+
+
+blueprint = Blueprint(
+    "ils_admin",
+    __name__,
+    template_folder="templates",
+    static_folder="static",
+)
+
+email_log = {
+    "model": EmailLog,
+    "modelview": EmailLogModelView,
+    "name": "Email Log",
+    "category": _("User Management"),
+}
+
+__all__ = ("email_log",)
--- a/invenio_app_ils/mail/tasks.py
+++ b/invenio_app_ils/mail/tasks.py
@@ -12,12 +12,16 @@ import json
 from celery import shared_task
 from celery.utils.log import get_task_logger
 from flask import current_app
+from invenio_accounts.models import User
 from invenio_mail.tasks import send_email
+from sqlalchemy.orm.exc import NoResultFound
+
+from invenio_app_ils.mail.models import EmailLog
 
 celery_logger = get_task_logger(__name__)
 
 
-def send_ils_email(message, name="ils_mail"):
+def send_ils_email(message, is_manually_triggered=False, name="ils_mail"):
     """Send an email async with Invenio-Mail and log success / errors.
 
     :param message: BlockTemplatedMessage mail message.
@@ -30,7 +34,10 @@ def send_ils_email(message, name="ils_mail"):
     log_msg = dict(
         name=name, action="before_send", message_id=dump["id"], data=dump
     )
-    celery_logger.debug(json.dumps(log_msg, sort_keys=True))
+    celery_logger.info(json.dumps(log_msg, sort_keys=True))
+
+    dump["is_manually_triggered"] = is_manually_triggered
+
     send_email.apply_async(
         (full_data,),
         link=log_successful_mail.s(dump),
@@ -45,6 +52,30 @@ def get_recipients(recipients):
     return recipients
 
 
+def create_email_log_entry(data, log_message):
+    """Create an entry in the email log db."""
+    try:
+        patron = User.query.filter_by(email=data["recipients"][0]).one()
+    except NoResultFound as ex:
+        current_app.logger.exception(ex)
+        return
+    recipient_id = str(patron.id)
+
+    data_dict = dict(
+        email_action=data.get("action", None),
+        recipient_user_id=recipient_id,
+        is_manually_triggered=data["is_manually_triggered"],
+        message_id=data["id"],
+        send_log=log_message,
+    )
+
+    pid_type = data.get("pid_type", None)
+    data_dict["pid_type"] = pid_type
+    pid_value = data.get("pid_value", None)
+    data_dict["pid_value"] = pid_value
+    EmailLog.create(data_dict)
+
+
 @shared_task
 def log_successful_mail(_, data):
     """Log successful email task."""
@@ -54,6 +85,9 @@ def log_successful_mail(_, data):
         message_id=data["id"],
         data=data,
     )
+
+    create_email_log_entry(data, "Success")
+
     celery_logger.info(json.dumps(log_msg, sort_keys=True))
 
 
@@ -69,4 +103,7 @@ def log_error_mail(request, exc, traceback, data, **kwargs):
         exception=repr(exc),
         data=data,
     )
+
+    create_email_log_entry(data, "Error: " + repr(exc))
+
     celery_logger.exception(json.dumps(error, sort_keys=True), exc_info=exc)
--- a/invenio_app_ils/mail/views.py
+++ b/invenio_app_ils/mail/views.py
@@ -1,0 +1,90 @@
+# Copyright (C) 2021 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio App ILS email views."""
+
+from flask import Blueprint, abort
+from sqlalchemy.orm.exc import NoResultFound
+from webargs import fields, validate
+from webargs.flaskparser import use_kwargs
+
+from invenio_app_ils.mail.models import EmailLog
+from invenio_app_ils.permissions import need_permissions
+
+
+def serialize_email(email):
+    """Serialize email from db."""
+    if email is None:
+        return {}
+    email.triggered_on = email.triggered_on.isoformat()
+    email = email.__dict__
+    del email["_sa_instance_state"]
+    return email
+
+
+def get_emails_list_blueprint(_):
+    """Add email list view to the blueprint."""
+    blueprint = Blueprint("invenio_app_ils_emails_list", __name__)
+
+    @blueprint.route("/emails")
+    @use_kwargs(
+        {
+            "recipient_user_id": fields.Int(),
+            "pid_type": fields.Str(),
+            "pid_value": fields.Str(),
+            "email_action": fields.Str(),
+            "size": fields.Int(validate=[validate.Range(min=1, max=100)]),
+        }
+    )
+    @need_permissions("send-email-to-patron")
+    def get_emails(
+        recipient_user_id=None,
+        pid_type=None,
+        pid_value=None,
+        email_action=None,
+        size=20,
+    ):
+        filters = {}
+        if recipient_user_id:
+            filters.update({"recipient_user_id": recipient_user_id})
+        if pid_type:
+            filters.update({"pid_type": pid_type})
+        if pid_value:
+            filters.update({"pid_value": pid_value})
+        if email_action:
+            filters.update({"email_action": email_action})
+
+        emails = (
+            EmailLog.query.filter_by(**filters)
+            .order_by(EmailLog.triggered_on.desc())
+            .limit(size)
+            .all()
+        )
+
+        for i, email in enumerate(emails):
+            emails[i] = serialize_email(email)
+
+        return {"hits": emails, "total": len(emails)}
+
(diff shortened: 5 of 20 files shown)


invenio_app_ils/mail/tasks.py
```diff
@@ -30,7 +40,13 @@ def send_ils_email(message, name="ils_mail"):
     log_msg = dict(
         name=name, action="before_send", message_id=dump["id"], data=dump
     )
-    celery_logger.debug(json.dumps(log_msg, sort_keys=True))
+    celery_logger.info(json.dumps(log_msg, sort_keys=True))
+
+    if called_by_celery_task():
```
ntarocco: I think that we need to take a decision here (ping @jrcastro2 @kprzerwa too). When the email is sent after a checkout, do we consider it automatic or manual?

IMHO, manual is only when the librarian will click on the button "send email" but in all other cases it should be automatic.
If we go for this solution then I would change this code to be always automatic and then override the param to manual only when coming from the endpoint that is connected to the send mail buttons.
The manual email is sent via the `LoanMailResource` endpoint.

invenio_app_ils/mail/tasks.py
```diff
@@ -30,7 +40,13 @@ def send_ils_email(message, name="ils_mail"):
...
+        dump["triggered"] = "Automatically"
+    else:
+        dump["triggered"] = "Manually"
```
ntarocco: Can we have an ENUM for this (not in the db field, but just a string), probably in the model class?

invenio_app_ils/mail/tasks.py
```diff
@@ -45,6 +61,39 @@ def get_recipients(recipients):
     return recipients
 
 
+def create_email_log_entry(data, log_message):
+    """Create an entry in the email log db."""
+    sender_email = data["recipients"][0]
+    SystemAgent = current_app.config["ILS_PATRON_SYSTEM_AGENT_CLASS"]
+    AnonymousPatron = current_app.config["ILS_PATRON_ANONYMOUS_CLASS"]
+
+    if sender_email in current_app.config["ILS_MAIL_NOTIFY_TEST_RECIPIENTS"]:
```
ntarocco: I am not sure why this code here. Can we access to the mail data and access the `patron` field?

invenio_app_ils/mail/tasks.py
```diff
@@ -69,4 +121,7 @@ def log_error_mail(request, exc, traceback, data, **kwargs):
         exception=repr(exc),
         data=data,
     )
+
+    create_email_log_entry(data, "Error" + exc)
```
ntarocco: It would be nice to test this: I think it is missing the space between `Error` and `exc` and we are using the function `repr` above, might be worth to have a look how it looks like.

invenio_app_ils/mail/views.py
```diff
@@ -0,0 +1,69 @@
...
+
+    return blueprint
+
+
+def get_emails_item_blueprint(_):
+    """Add emails item view to the blueprint."""
+    blueprint = Blueprint("invenio_app_ils_emails_item", __name__)
+
+    @blueprint.route("/emails/<id>")
+    @need_permissions("emails-sent")
```
ntarocco:
```suggestion
    @need_permissions("send-email-to-patron")
```

invenio_app_ils/mail/views.py
```diff
@@ -0,0 +1,69 @@
...
+        return {"hits": emails, "total": len(emails)}
+
+    return blueprint
+
+
+def get_emails_item_blueprint(_):
+    """Add emails item view to the blueprint."""
+    blueprint = Blueprint("invenio_app_ils_emails_item", __name__)
+
+    @blueprint.route("/emails/<id>")
```
ntarocco: Missing param validation with `webargs`. Pay attention here because incoming values from the request should be always be sanitized.

invenio_app_ils/mail/views.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @blueprint.route("/emails")
+    @use_kwargs(
+        {
+            "recipient_user_id": fields.Int(),
+            "loan_pid": fields.Str(),
+            "email_action": fields.Str(),
+        }
+    )
+    @need_permissions("emails-sent")
```
ntarocco:
```suggestion
    @need_permissions("send-email-to-patron")
```

invenio_app_ils/mail/views.py
```diff
@@ -0,0 +1,69 @@
...
+    def get_emails(recipient_user_id=None, loan_pid=None, email_action=None):
+        filters = {}
+        if recipient_user_id:
+            filters.update({"recipient_user_id": recipient_user_id})
+        if loan_pid:
+            filters.update({"loan_pid": loan_pid})
+        if email_action:
+            filters.update({"email_action": email_action})
+
+        emails = EmailLog.query.filter_by(**filters).all()
```
ntarocco: Probably a good idea to have default size/limit of e.g. 20 and maybe a size param with a max to e.g. 100

ntarocco: what about sorting? I would add a sort by the triggered_on field, probably DESC? (most recent first?)

invenio_app_ils/mail/views.py
```diff
@@ -0,0 +1,69 @@
...
+from invenio_app_ils.mail.model import EmailLog
+from invenio_app_ils.permissions import need_permissions
+
+
+def serialize_email(email):
+    """Serialize email from db."""
+    if email is None:
+        return {}
+    email = email.__dict__
+    del email["_sa_instance_state"]
```
ntarocco, line 23: question: is the date of the triggered_on serialized back correctly is ISO format?

invenio_app_ils/patrons/anonymization.py
```diff
@@ -107,8 +108,11 @@ def anonymize_patron_data(patron_pid, force=False):
     for hit in patron_loans:
         loan = Loan.get_record_by_pid(hit.pid)
 
-        cancelled = current_app.config["CIRCULATION_STATES_LOAN_CANCELLED"]
-        if loan["state"] not in cancelled:
+        completed = (
+            current_app.config["CIRCULATION_STATES_LOAN_CANCELLED"]
```
ntarocco, line 122: Thank you for adding this extra check.

invenio_app_ils/patrons/anonymization.py
```diff
@@ -221,3 +228,18 @@ def delete_user_account(patron_pid):
         dropped += User.query.filter(User.id == patron_pid).delete() or 0
 
     return dropped
+
+
+def anonymize_patron_in_email_logs(patron_pid):
```
ntarocco, line 249: 👍🏼

tests/api/ils/test_mail_permissions.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (C) 2021 CERN.
```
ntarocco, line 1: It would be also nice to add some tests about inserting data in the db and reading it from the endpoint.

There are also some other emails tests: you could assert there that the email log exists in the DB.

ntarocco:
@KonstantinaStoikou great job! It looks very good! I have added a few comments, let me know if you need clarifications.

invenio_app_ils/mail/tasks.py
jrcastro2: I agree with you, I would consider it manual only if the librarian explicitly send it.

invenio_app_ils/mail/admin.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # Entries are read-only
+    can_create = False
+    can_edit = False
+    can_delete = False
+
+    can_view_details = True
+
+    form_excluded_columns = EmailLog.__table__.columns
+    column_default_sort = (EmailLog.triggered_on, True)
```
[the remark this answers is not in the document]
KonstantinaStoikou: I added the widget.

invenio_app_ils/document_requests/mail/messages.py
```diff
@@ -57,4 +58,5 @@ def dump(self):
         """Dump document request email data."""
         data = super().dump()
         data["document_request_pid"] = self.request["pid"]
+        data["action"] = self.action
```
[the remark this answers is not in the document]
KonstantinaStoikou: In the db model the `action` field uses the default value for `nullable` which is `True`

invenio_app_ils/mail/views.py
KonstantinaStoikou: The date has the format `"Mon, 22 Feb 2021 16:15:17 GMT"`

invenio_app_ils/mail/tasks.py
```diff
@@ -45,6 +54,27 @@ def get_recipients(recipients):
     return recipients
 
 
+def create_email_log_entry(data, log_message):
+    """Create an entry in the email log db."""
+    patron = User.query.filter_by(email=data["recipients"][0]).one_or_none()
```
ntarocco: unfortunately `patron` could be `None` (`one_or_none`)...
I would change this to
```
from sqlalchemy.orm.exc import NoResultFound
try:
    patron = User.query.filter_by(email=data["recipients"][0]).one()
except NoResultFound as ex:
    current_app.logger.exception(exception)
    return

```

invenio_app_ils/mail/tasks.py
```diff
@@ -45,6 +54,27 @@ def get_recipients(recipients):
...
+    recipient_id = patron.get_id()
```
ntarocco: isn't is just `str(user.id)`?

invenio_app_ils/mail/tasks.py
```diff
@@ -45,6 +54,27 @@ def get_recipients(recipients):
...
+    data_dict = dict(
+        email_action=data["action"],
+        recipient_user_id=recipient_id,
+        trigger=data["trigger"],
+        message_id=data["id"],
+        send_log=log_message,
+    )
+
+    pid_type = data.get("pid_type", None)
+    data_dict.update({"pid_type": pid_type})
```
ntarocco:
```suggestion
    data_dict["pid_type"] = pid_type
```
same below

invenio_app_ils/mail/views.py
ntarocco: we should always store dates in the ISO format to avoid mistakes

invenio_app_ils/mail/views.py
```diff
@@ -0,0 +1,78 @@
...
+
+def get_emails_item_blueprint(_):
+    """Add emails item view to the blueprint."""
+    blueprint = Blueprint("invenio_app_ils_emails_item", __name__)
+
+    @blueprint.route("/emails/<id>")
+    @use_kwargs({"id": fields.Int(required=True)})
+    @need_permissions("send-email-to-patron")
+    def get_email(id):
+        email = EmailLog.query.filter_by(id=id).one_or_none()
```
ntarocco: `email` can be `None`
```
try:
   email =  ...one()
except NoResultsFound:
    abort(404)
```

invenio_app_ils/patrons/anonymization.py
```diff
@@ -175,14 +197,20 @@ def anonymize_patron_data(patron_pid, force=False):
...
-    return dropped, indices
+    return dropped, indices, emails
 
 
 def delete_user_account(patron_pid):
     """Deletes a user account from the database, without updating the index."""
     dropped = 0
 
     with db.session.begin_nested():
+        # delete rows from transaction table
```
ntarocco: Lars was suggesting to set a global config `DB_VERSIONING_USER_MODEL = None` instead, see [here](https://github.com/inveniosoftware/invenio-db/blob/master/invenio_db/ext.py#L116-L124). WDYT?

tests/api/ils/test_mails.py
```diff
@@ -66,7 +66,7 @@ def test_log_successful_error_mail_task(app_with_mail, mocker):
 
     assert not succ.s.called
     assert not err.s.called
-    send_ils_email(TestMessage())
+    send_ils_email(TestMessage(), False)
```
ntarocco: it would be nice to add the name of the param her and below

invenio_app_ils/mail/views.py
```diff
@@ -0,0 +1,77 @@
...
+
+def get_emails_list_blueprint(_):
+    """Add email list view to the blueprint."""
+    blueprint = Blueprint("invenio_app_ils_emails_list", __name__)
+
+    @blueprint.route("/emails")
+    @use_kwargs(
+        {
+            "recipient_user_id": fields.Int(),
+            "loan_pid": fields.Str(),
```
ntarocco: This should be:
```suggestion
            "pid_type": fields.Str(),
            "pid_value": fields.Str(),
```

invenio_app_ils/circulation/receivers.py
```diff
@@ -53,6 +53,12 @@ def index_after_loan_replace_item(_, old_item_pid, new_item_pid):
 
 def send_email_after_loan_change(_, initial_loan, loan, trigger):
     """Send email notification when the loan changes."""
+    cancel_reason = loan.get("cancel_reason", None)
+    if cancel_reason == "Anonymization":
```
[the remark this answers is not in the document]
KonstantinaStoikou: In `anonymize` function we are first cancelling the loans and then changing the patrons to anonymous. If I change the `patron_pid` before the cancel trigger it throws an error `Working outside of application context.`

invenio_app_ils/mail/views.py
KonstantinaStoikou: I added a line above to transform it to ISO format.

invenio_app_ils/circulation/receivers.py
```diff
@@ -53,6 +53,11 @@ def index_after_loan_replace_item(_, old_item_pid, new_item_pid):
 
 def send_email_after_loan_change(_, initial_loan, loan, trigger):
     """Send email notification when the loan changes."""
+    cancel_reason = loan.get("cancel_reason", None)
+    if cancel_reason == "Anonymization.":
```
ntarocco: I am sorry but I am still wondering if we really need this.
Since the `patron` is `-2`, why this?

invenio_app_ils/mail/tasks.py
```diff
@@ -30,7 +34,13 @@ def send_ils_email(message, name="ils_mail"):
     log_msg = dict(
         name=name, action="before_send", message_id=dump["id"], data=dump
     )
-    celery_logger.debug(json.dumps(log_msg, sort_keys=True))
+    celery_logger.info(json.dumps(log_msg, sort_keys=True))
+
+    if is_manually_triggered:
```
ntarocco: why not:
```
dump["is_manually_triggered"] = is_manually_triggered
```
?

tests/api/ils/test_mails.py
```diff
@@ -92,3 +99,49 @@ def test_send_only_to_test_recipients(app_with_mail, mocker):
         assert outbox[0].recipients == fake_recipients
 
     app_with_mail.config["ILS_MAIL_ENABLE_TEST_RECIPIENTS"] = False
+
+
+def test_email_db_table_and_endpoint(
```
ntarocco, line 104: Nice thank you! 🚀 👏🏼

ntarocco (approved):

invenio_app_ils/circulation/receivers.py
KonstantinaStoikou: I answered above https://github.com/inveniosoftware/invenio-app-ils/pull/1037#discussion_r583677304

tests/api/ils/test_mail_permissions.py
KonstantinaStoikou: I have added the tests

invenio_app_ils/circulation/receivers.py
ntarocco: Thanks for the explanations. As discussed, let's remove this if and let's send the email.

invenio_app_ils/patrons/anonymization.py
```diff
@@ -107,19 +118,22 @@ def anonymize_patron_data(patron_pid, force=False):
...
+            current_app.config["CIRCULATION_STATES_LOAN_CANCELLED"]
+            + current_app.config["CIRCULATION_STATES_LOAN_COMPLETED"]
+        )
+        if loan["state"] not in completed:
             params = deepcopy(loan)
             params.update(
                 dict(
-                    cancel_reason="Loan request cancelled because of user "
-                    "deletion.",
+                    cancel_reason="Anonymization.",
```
ntarocco:
```suggestion
                    cancel_reason="Loan request cancelled by the system.",
```

ntarocco (approved):

ntarocco (approved):

PR #1041, opened 2021-02-22
document: add meta field to urls

kpsherva:


ntarocco (commented):
What about the ES mapping?

ntarocco (approved):

ntarocco:
Ups sorry... you need to add also the marshmallow loader otherwise on any update of the doc it will be lost

kpsherva:
in the end I added the meta field also for the series, I noticed that it will be consistent with the Document

ntarocco (approved):

PR #1044, opened 2021-02-23
[Merge with RIAI#372] Item fix circ restriction and authors limit

ntarocco:
* remove 4 weeks form item restrictions, already covered by the default
  no restriction value
* limit the number of authors for resolved documents

### DIFF
--- a/invenio_app_ils/documents/utils.py
+++ b/invenio_app_ils/documents/utils.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Document utils."""
+
+MAX_AUTHORS = 5
+
+
+def flatten_authors(authors):
+    """Flatten document authors array to a string."""
+    result = "; ".join([a["full_name"] for a in authors[: MAX_AUTHORS]])
+
+    if len(authors) > MAX_AUTHORS:
+        result += " et al."
+
+    return result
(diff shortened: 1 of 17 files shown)


invenio_app_ils/documents/utils.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Document utils."""
+
+MAX_AUTHORS = 5
```
kpsherva, line 10: nit: could it be overridable in the overlay?

invenio_app_ils/eitems/mappings/v6/eitems/eitem-v1.0.0.json
```diff
@@ -33,69 +33,7 @@
...
-                },
-                "roles": {
-                  "type": "keyword"
-                },
-                "type": {
-                  "type": "keyword"
-                }
-              },
-              "type": "object"
+              "type": "text"
```
kpsherva, line 36: I think it was a good decision not resolve the authors with the whole structure :)

kpsherva (approved):

invenio_app_ils/documents/utils.py
ntarocco: I thought to create a config for it, but then honestly I thought that it was just too much... do you prefer to have it a config?

kpsherva: no I'm fine, it's not a blocking issue, and very low impact

PR #1054, opened 2021-03-03
ES: replaces execute()

### LINKED ISSUE #1048: Fix email reminders
The implemented search query to retrieve the overdue loans is only taking into account the first `10` loans because the `execute` method is used instead of `scan`, see [here](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/circulation/mail/tasks.py#L66).

- [x] replace the `execute` method with the `scan` in the for loop:
```
overdue_loans = get_all_overdue_loans().execute()
for hit in overdue_loans.hits:
```
should be:
```
overdue_loans = get_all_overdue_loans().execute()
for hit in get_all_overdue_loans().scan():
```
- [x] for any usage of the `execute` method if it is ok or not to retrieve only the first 10 results, for example:
  - [here](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/circulation/mail/tasks.py#L87)
  - [here](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/circulation/tasks.py#L30)
  - [here](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/circulation/stats/api.py#L36)
- [x] fix the method [patron_has_active_loan_or_request_on_document](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/circulation/api.py#L88) (and its usage) that should return only the first result or `None`, and not the list of results, to avoid the same mistake.
- [x] check CDS-ILS code and migration to make sure that the right method is used (notice possible issue in importer)

jrcastro2:
* Replaces execute() with scan()
  when we retrieve more than 10 hits
* closes https://github.com/inveniosoftware/invenio-app-ils/issues/1048

### DIFF
--- a/invenio_app_ils/circulation/api.py
+++ b/invenio_app_ils/circulation/api.py
@@ -86,7 +86,7 @@ def _set_item_to_can_circulate(item_pid):
 
 
 def patron_has_active_loan_or_request_on_document(patron_pid, document_pid):
-    """Return True if patron has an active loan/request for given document."""
+    """Return loan/request if it's active for the given patron and document."""
     states = (
         current_app.config["CIRCULATION_STATES_LOAN_REQUEST"]
         + current_app.config["CIRCULATION_STATES_LOAN_ACTIVE"]
@@ -95,13 +95,13 @@ def patron_has_active_loan_or_request_on_document(patron_pid, document_pid):
         patron_pid=patron_pid, document_pid=document_pid, filter_states=states
     )
     search_result = search.execute()
-
-    return (
-        search_result,
-        search_result.hits.total > 0
-        if lt_es7
-        else search_result.hits.total.value > 0,
-    )
+    total = search_result.hits.total \
+        if lt_es7 \
+        else search_result.hits.total.value
+    if total > 0:
+        return search_result.hits[0]
+    else:
+        return None
 
 
 def request_loan(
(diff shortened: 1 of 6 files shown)


invenio_app_ils/circulation/stats/api.py
```diff
@@ -33,6 +33,7 @@ def fetch_most_loaned_documents(from_date, to_date, bucket_size):
     doc_search = current_app_ils.document_search_cls()
     doc_search = doc_search.with_preference_param().params(version=True)
     doc_search = doc_search.search_by_pid(*document_pids)
+    doc_search = doc_search[0:bucket_size]
```
jrcastro2, line 36: I am not sure about this but since bucket_size default value is 30 I kept it like this because if we add the scan(), we need to add a custom serializer for this or create the returning object with the fields needed by the serializer.

invenio_app_ils/circulation/api.py
```diff
@@ -95,13 +95,10 @@ def patron_has_active_loan_or_request_on_document(patron_pid, document_pid):
...
     )
     search_result = search.execute()
-
-    return (
-        search_result,
-        search_result.hits.total > 0
-        if lt_es7
-        else search_result.hits.total.value > 0,
-    )
+    if len(search_result) > 0:
```
ntarocco: I am not 100% sure this is correct because the length should be computed on the `hits` field. WDYT?

```suggestion
    total = search_result.hits.total
        if lt_es7
        else search_result.hits.total.value
    if total > 0:
```

invenio_app_ils/circulation/stats/api.py
ntarocco: I am not sure I understand your comment here, let's chat about this.

invenio_app_ils/circulation/api.py
jrcastro2: Yes, you are right I only tested it in my local environment and didn't consider other ES versions, I will change it, thanks!

ntarocco (approved):

PR #1057, opened 2021-03-05
Merge libraries and vendors

### LINKED ISSUE #1051: Merge libraries and vendors into one record type `providers`
Having 2 distinguished list of libraries and vendors is not ideal because often Acq/ILLs can be requested to Libraries/Vendors and it will require the librarian to duplicate records.

- [x] rename Vendors to Providers (schema, mapping, loader, config, filters, sort orders, APIs, demo data, tests). Should it be move to a standalone module/package or kept inside ACQ? -> was decided to be a standalone module
- [x] move providers to a standalone module/package
- [x] delete ILL Libraries and change the refs of Acq/ILLs to use providers instead
- [x] add a new field "type" in the provider model of type vocabulary which will contain `LIBRARY`, `VENDOR`, etc...
- [x] apply the same modifications to the frontend. The Create/Edit form should be implemented with RJSF
- [x] modify migration

KonstantinaStoikou:
- send email even if there is no user in db with that email

Closes #1051


invenio_app_ils/acquisition/config.py
```diff
@@ -142,38 +89,25 @@
...
 
 RECORDS_REST_FACETS = dict(
     acq_orders=dict(  # OrderSearch.Meta.index
         aggs=dict(
             status=dict(terms=dict(field="status")),
-            vendor=dict(
+            provider=dict(
                 terms=dict(
-                    field="vendor.name.keyword", size=FACET_VENDOR_LIMIT
+                    field="provider.name.keyword", size=FACET_PROVIDER_LIMIT
```
ntarocco, line 102: I didn't remember this. @kprzerwa maybe we can add this to limit the number of languages in the frontsite facet for the time being?

invenio_app_ils/providers/api.py
```diff
@@ -0,0 +1,58 @@
...
+
+    _pid_type = PROVIDER_PID_TYPE
+    _schema = "providers/provider-v1.0.0.json"
+
+    def delete(self, **kwargs):
+        """Delete record."""
+        if self["type"] == "VENDOR":
+            search_cls = current_ils_acq.order_search_cls
+            ref_type = "Order"
+        else:
```
ntarocco: as commented in the frontend PR, we are not sure how the librarians will use vendors or libraries, they might be mixing them.
Just change to do both searches: to delete a Provider, there must be `0` orders AND `0` brwreq attached.

ntarocco: Actually, is it better than to have a `enum` instead of a vocab? WDYT?

invenio_app_ils/providers/config.py
```diff
@@ -0,0 +1,96 @@
...
+        ),
+        bestmatch=dict(
+            fields=["-_score"],
+            title="Best match",
+            order=3,
+        ),
+    ),
+)
+
+FACET_TYPE_LIMIT = 2
```
ntarocco: Maybe 5 at least?
```suggestion
FACET_TYPE_LIMIT = 5
```

invenio_app_ils/providers/indexer.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@shared_task(ignore_result=True)
+def provider_index_referenced_records(provider):
+    """Index referenced records."""
+    indexer = ReferencedRecordsIndexer()
+    indexed = dict(pid_type=PROVIDER_PID_TYPE, record=provider)
+    referenced = []
+
+    if provider["type"] == "VENDOR":
```
ntarocco: Same here for the `if/elif/else`

ntarocco (commented):
Wow! Great work! 💯 🚀

invenio_app_ils/acquisition/config.py
kpsherva: we will push the languages down as a facet, but before we have "show more" for facets I wouldn't limit the languages on the frontsite, I think it will be confusing for the users

invenio_app_ils/mail/tasks.py
```diff
@@ -57,8 +57,8 @@ def create_email_log_entry(data, log_message):
     try:
         patron = User.query.filter_by(email=data["recipients"][0]).one()
     except NoResultFound as ex:
-        current_app.logger.exception(ex)
-        return
+        current_app.logger.warning(ex)
+        patron = current_app.config["ILS_PATRON_ANONYMOUS_CLASS"]
```
ntarocco, line 61: Thank you!

ntarocco (approved):

PR #1060, opened 2021-03-09
email: special characters

jrcastro2:
* Uses Markup ro prepare strings
  to be inserted in HTML


tests/api/circulation/test_mails.py
```diff
@@ -84,6 +85,34 @@ def test_email_on_loan_checkout(
...
 
+def test_email_with_special_characters(
+    client, app_with_mail, users, testdata, loan_params
+):
+    """Test that special characters are correctly displayed."""
+    loan_data = testdata["loans"][1]
+    loan = Loan.get_record_by_pid(loan_data["pid"])
+    doc = Document.get_record_by_pid(loan_data["document_pid"])
+    doc["title"] = \
+        "Spèciâl chäráctèrs: " \
```
kpsherva: :)

ntarocco:
I have tested this locally and it works fine:

Screenshot 2021-03-15 at 20 11 40

I am just wondering if there is maybe an easier solution... I will give it a thought and come back to you :)

kpsherva (approved):

jrcastro2:
I think we should also mark the `full_title` as `|safe` , and maybe also the `patron.name`. WDYT?
(Tests are failing)

ntarocco:
> I think we should also mark the `full_title` as `|safe` , and maybe also the `patron.name`. WDYT?
> (Tests are failing)

Thanks for the tests, weird error, I am checking.
For the full title, or any other variable, I think that they should be escaped indeed in the body, because we don't trust user input much... Probably patron name is ok, but it does not hurt I guess to keep it escaped. WDYT?

ntarocco:
Here the HTML version:

Screenshot 2021-03-16 at 09 29 06

Here the plaintext:

Screenshot 2021-03-16 at 09 28 51

KonstantinaStoikou (approved):

jrcastro2 (commented):
LGTM!

PR #1075, opened 2021-03-31
series: add series type to schema

### LINKED ISSUE #1073: Add series_type to series
Add `series_type` to series schema, mappings etc. similar to `document_type`
in the mapping, add copy_to = document_type so that the values will appear under document type facet of the literature search (frontsite)
`series_type` will be an enum, waiting for values from the library

KonstantinaStoikou:
Closes https://github.com/inveniosoftware/invenio-app-ils/issues/1073

### DIFF
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -913,11 +913,13 @@ RECORDS_REST_FACETS = dict(
             moi=dict(terms=dict(field="mode_of_issuance")),
             language=dict(terms=dict(field="languages")),
             relation=dict(terms=dict(field="relation_types")),
+            sertype=dict(terms=dict(field="series_type")),
         ),
         post_filters=dict(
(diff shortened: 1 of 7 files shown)


invenio_app_ils/config.py
```diff
@@ -913,11 +913,13 @@ def _(x):
             moi=dict(terms=dict(field="mode_of_issuance")),
             language=dict(terms=dict(field="languages")),
             relation=dict(terms=dict(field="relation_types")),
+            sertype=dict(terms=dict(field="series_type")),
```
kpsherva, line 916:
```suggestion
            series_type=dict(terms=dict(field="series_type")),
```

KonstantinaStoikou: I named it `sertype` to be similar to the `doctype` filter for documents.

PR #1084, opened 2021-05-06
circulation: prevent creating stale loan requests

kpsherva:


invenio_app_ils/circulation/api.py
```diff
@@ -204,8 +223,18 @@ def checkout_loan(
...
-    pid = ils_circulation_loan_pid_minter(record_uuid, data=new_loan)
-    loan = Loan.create(data=new_loan, id_=record_uuid)
+
+    # check if there is an existing request
+    loan = patron_has_request_on_document(
+        patron_pid, kwargs.get("document_pid")
+    )
+    if loan:
+        loan = loan_cls.get_record_by_pid(loan.pid)
+        pid = IlsCirculationLoanIdProvider.get(loan["pid"]).pid
```
ntarocco, line 233: isn't it enough to simply do:
```suggestion
        pid = loan.pid
```
?

kpsherva: we don't have pid attr on loan, I double checked
Screenshot 2021-05-07 at 11 05 49

PR #1091, opened 2021-05-21
items: add document type to resolver

kpsherva:
addresses https://github.com/CERNDocumentServer/cds-ils/issues/515

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@
 Changes
 =======
 
+Version 1.0.0a43 (released 2021-05-27)
+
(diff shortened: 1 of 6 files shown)


jrcastro2 (approved):

ntarocco (commented):
Why this? Do you have a GH Issue for this? I don't see the `closes` mention or anything else.

You need to add this to the ES mappings too and to check what happens when you start indexing this.
Are you going to re-index everything?
Did you check if everything will work?
Did you check ES doc on what happens when you add a new field?

ntarocco (approved):

PR #1109, opened 2021-09-27
[DEPENDS ON CIRCULATION] Bulk loan extension

kpsherva:


invenio_app_ils/permissions.py
```diff
@@ -145,7 +145,8 @@ def __init__(self, record):
 
 def views_permissions_factory(action):
     """Return ILS views permissions factory."""
-    is_authenticated_user = ["circulation-loan-request", "patron-loans"]
+    is_authenticated_user = ["circulation-loan-request", "patron-loans",
+                             "bulk-loan-extension"]
```
kpsherva, line 149: I would like to discuss this added permission because it can be misleading

invenio_app_ils/circulation/api.py
```diff
@@ -247,6 +250,30 @@ def checkout_loan(
...
+        try:
+            extended_loan = current_circulation.circulation.trigger(
+                loan_record,
+                **dict(
+                    params,
+                    trigger="bulk_extend",
+                )
+            )
+            yield extended_loan
+        except Exception:
```
kpsherva: if there is a problem with extension we simply ignore it, but maybe there are better ideas?

ntarocco: I would return a list of success and failed so that it can be displayed to the user, otherwise the user might be tempted to click multiple times on the bulk extend button! :)

invenio_app_ils/circulation/transitions/transitions.py
```diff
@@ -45,3 +45,12 @@ def before(self, loan, initial_loan, **kwargs):
...
 
         _update_loan_end_date(loan)
+
+
+class ILSBulkItemOnLoanToItemOnLoan(ILSItemOnLoanToItemOnLoan):
+    """Extend action to perform a item loan extension."""
+
+    def after(self, loan, initial_loan, **kwargs):
+        """Apply actions after state change."""
+        kwargs["send_notification"] = False
```
ntarocco: is this used somewhere?

invenio_app_ils/circulation/api.py
ntarocco: Can we change the exception here to `CirculationException` and also `IlsException` maybe?

invenio_app_ils/circulation/search.py
```diff
@@ -48,6 +50,33 @@ def get_all_expiring_loans(expiring_in_days):
...
+        .filter("terms", state=states)
+    )
+
+
+def get_all_range_expiring_ils_loans_by_patron_pid(expiring_in_days,
+                                                   patron_pid):
+    """Return ILS loans expiring in the next <days> days for given patron."""
+    return get_all_range_expiring_loans_by_patron_pid(
+        expiring_in_days, patron_pid)\
+        .filter("term", item_pid__type=ITEM_PID_TYPE)
```
ntarocco, line 77: is this to avoid to extend the ILLs?

invenio_app_ils/circulation/serializers/response.py
```diff
@@ -0,0 +1,49 @@
...
+             code=200, headers=None, **kwargs):
+        """Generate the response object."""
+        response_data = {"extended_loans": [],
+                         "not_extended_loans": []}
+        if isinstance(extended_loans, list):
+            for loan in extended_loans:
+                response_data["extended_loans"]\
+                    .append(serializer.serialize(loan["pid"], loan))
+        else:
+            abort(204)
```
ntarocco: as discussed: to remove? Just for-loop...

invenio_app_ils/circulation/templates/invenio_app_ils_circulation/notifications/bulk_extend.html
```diff
@@ -0,0 +1,43 @@
...
+"{{ obj.document.full_title }}" <{{ spa_routes.HOST }}{{ spa_routes.PATHS['literature']|format(pid=obj.document.pid) }}> until {{ obj.loan.end_date }}
+{% endfor %}
+{% endif %}
+
+{% if not_extended_loans %}
+your loans for
+{% for obj in not_extended_loans %}
+"{{ obj.document.full_title }}" <{{ spa_routes.HOST }}{{ spa_routes.PATHS['literature']|format(pid=obj.document.pid) }}>
+{% endfor %}
+could not be extended. Contact the library for more information or consult your profile page.
```
ntarocco: Can we add the link to the profile page?

invenio_app_ils/circulation/transitions/transitions.py
```diff
@@ -45,3 +45,12 @@ def before(self, loan, initial_loan, **kwargs):
         super().before(loan, initial_loan, **kwargs)
 
         _update_loan_end_date(loan)
+
+
+class ILSBulkItemOnLoanToItemOnLoan(ILSItemOnLoanToItemOnLoan):
```
ntarocco: This might not be needed now...

invenio_app_ils/circulation/loaders/schemas/json/bulk_extend.py
```diff
@@ -0,0 +1,24 @@
...
+    """Loan update dates schema."""
+
+    class Meta:
+        """Meta attributes for the schema."""
+
+        from marshmallow import EXCLUDE
+
+        unknown = EXCLUDE
+
+    patron_pid = fields.String()
```
ntarocco: in other places in the code we use `fields.Str` but it is probably the same. Why not having it `required=True`?

invenio_app_ils/circulation/views.py
```diff
@@ -144,6 +173,24 @@ def post(self, **kwargs):
...
+class BulkLoanExtensionResource(IlsCirculationResource):
+    """Bulk loan extension resource."""
+
+    view_name = "bulk_loan_extension"
+
+    @need_permissions("bulk-loan-extension")
+    def post(self, **kwargs):
+        """Loan checkout post method."""
+        data = self.loader()
+        if "patron_pid" not in data:
```
ntarocco, line 186: It might that the `patron_pid` sent is a wrong one. Maybe safer to set it to the current logged in user or failed if different?

invenio_app_ils/circulation/loaders/schemas/json/bulk_extend.py
kpsherva: Its the same class in the end (I've seen it in the marshmallow code). I added the required

invenio_app_ils/circulation/views.py
kpsherva: I added the check below now

invenio_app_ils/circulation/search.py
kpsherva: yes, because they are handled in a special way

ntarocco (approved):
Good job! 👏🏼 Please create an issue to add tests :)

PR #1112, opened 2021-09-28
index: reindex document after item deletion

### LINKED ISSUE #580: The number of items is not updated in the brief format when we delete an item
**Describe the bug**
If we delete a physical item, the number of item seems to be correctly modified in the document record, but not if we look at it in the search result. 

**To Reproduce**
- For this document (22c1y-has82), the item has been deleted earlier this week
- On the document page, we see clearly that there is no physical item:
see the screenshot: 
Screenshot 2021-08-11 at 12 36 37
- But if we look at it from a search here: https://catalogue.library.cern/backoffice/documents?q=&l=list&order=desc&p=1&s=15&sort=created
It is marked: 1 copy available for loan.
See the screenshot:
Screenshot 2021-08-11 at 12 39 03

This has an impact on the frontside search, the document appears as available in paper: 
Screenshot 2021-08-11 at 12 42 36

**Expected behavior**
The information should be up-to-date everywhere.

kpsherva:
* reindex after eitem deletion
* closes https://github.com/CERNDocumentServer/cds-ils/issues/580

### DIFF
--- a/invenio_app_ils/ext.py
+++ b/invenio_app_ils/ext.py
@@ -31,6 +31,7 @@ from .items.api import ITEM_PID_TYPE
 from .locations.api import LOCATION_PID_TYPE
 from .locations.receivers import register_location_signals
 from .patrons.api import PATRON_PID_TYPE
+from .records.receivers import register_record_signals
 from .series.api import SERIES_PID_TYPE
 from .vocabularies.api import VOCABULARY_PID_TYPE
 
@@ -298,6 +299,7 @@ class InvenioAppIlsREST(InvenioAppIls):
         """Register signals."""
         register_circulation_signals()
         register_files_signals()
+        register_record_signals()
         if app.config["EXTEND_LOANS_LOCATION_UPDATED"]:
             register_location_signals()
 
--- a/invenio_app_ils/documents/indexer.py
+++ b/invenio_app_ils/documents/indexer.py
@@ -158,3 +158,13 @@ class DocumentIndexer(RecordIndexer):
         super().index(document)
         eta = datetime.utcnow() + current_app.config["ILS_INDEXER_TASK_DELAY"]
         index_referenced_records.apply_async((document,), eta=eta)
+
+
+@shared_task(ignore_result=True)
+def index_document(document_pid):
+    """Async document indexer."""
+    document_class = current_app_ils.document_record_cls
+
+    document_indexer = current_app_ils.document_indexer
+    document = document_class.get_record_by_pid(document_pid)
+    document_indexer.index(document)
--- a/invenio_app_ils/records/listeners.py
+++ b/invenio_app_ils/records/listeners.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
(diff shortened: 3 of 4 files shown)


invenio_app_ils/items/receivers.py
```diff
@@ -0,0 +1,38 @@
...
+    record = kwargs["record"]
+    schema = record["$schema"]
+    item_class = current_app_ils.item_record_cls
+    eta = datetime.utcnow() + current_app.config["ILS_INDEXER_TASK_DELAY"]
+    is_item = schema.endswith(item_class._schema)
+    if is_item:
+        document_pid = record["document_pid"]
+        try:
+            index_document.apply_async((document_pid,), eta=eta)
+        except PIDDoesNotExistError:
```
ntarocco: is this exception raised even by an async task?

invenio_app_ils/items/receivers.py
```diff
@@ -0,0 +1,38 @@
...
+    """Register location signals."""
+    after_record_delete.connect(item_record_delete_listener, weak=False)
+
+
+def item_record_delete_listener(sender, *args, **kwargs):
+    """Listens for record deletes to reindex parent document."""
+    record = kwargs["record"]
+    schema = record["$schema"]
+    item_class = current_app_ils.item_record_cls
+    eta = datetime.utcnow() + current_app.config["ILS_INDEXER_TASK_DELAY"]
```
ntarocco: Very minor: move the `record = kwargs["record"]` and the `eta = ...` below, inside the `if` where it is used, to keep things more readable.

invenio_app_ils/ext.py
```diff
@@ -298,6 +299,8 @@ def _register_signals(self, app):
         """Register signals."""
         register_circulation_signals()
         register_files_signals()
+        register_item_signals()
```
ntarocco: Suggestion: we could maybe refactor this and have:
```
register_record_delete_signal()
```
and instead of `register_location_signals` have `register_record_update_signal()`.

Then, register the signal only once, in one method, and inside the method check what record. E.g.:

```
def register_record_delete_signal():
    after_record_delete.connect(record_delete_listener, weak=False)

def record_delete_listener(sender, *args, **kwargs):
    schema = record["$schema"]
    is_eitem = schema.endswith( current_app_ils.eitem_record_cls._schema)
    ... is_item ...
    if is_eitem....
    elif is_item ....
```

and

```
def register_record_update_signal():
    after_record_update.connect(record_update_listener, weak=False)

def record_update_listener(sender, *args, **kwargs):
    record = kwargs["record"]
    if isinstance(record, current_app_ils.location_record_cls):
        if app.config["EXTEND_LOANS_LOCATION_UPDATED"]:
            notify_location_updated(record["pid"])
```

ntarocco (commented):
I have the feeling that we will hav to

invenio_app_ils/items/receivers.py
kpsherva: true, it was a leftover, thanks for spotting it

invenio_app_ils/ext.py
kpsherva: I was considering it, and I decided to split in the end to separate modules, because I thought it would be following the convention of the code residing in the each module (e-items handling in eitems folder etc) but it is not a strong preference, I can handle it on the top application level.

kpsherva: in the end I implemented the hybrid, signals are registered in the respective modules, but handled by one common record delete listener

PR #1119, opened 2021-10-05
notifications: fix bulk extend notification

kpsherva:


invenio_app_ils/circulation/templates/invenio_app_ils_circulation/notifications/bulk_extend.html
```diff
@@ -26,18 +26,18 @@
...
+your loans have been extended as follows:<br/>
 {% for obj in extended_loans %}
 <a href="{{ spa_routes.HOST }}{{ spa_routes.PATHS['literature']|format(pid=obj.document.pid) }}">"{{ obj.document.full_title }}"</a> until <b>{{ obj.loan.end_date }}</b><br/>
-{% endfor %}<br/><br/>
+{% endfor %}
 {% endif %}
 
 {% if not_extended_loans %}
-your loans for: <br/><br/>
+{% if extended_loans %}<br/> {% endif %}
```
ntarocco, line 36: you can move this `<br>` just above, between endfor and endif and remove it from here:
```
{% endfor %}
<br />
{% endif %}
```

ntarocco (approved):

invenio_app_ils/circulation/templates/invenio_app_ils_circulation/notifications/bulk_extend.html
kpsherva: it is supposed to add the newline only when we have both extended loans and failed extensions, with no failed extensions the email will have 3x`<br/>` tags (because of the footer template)

PR #1122, opened 2021-10-11
literature search: improved relevance of results

### LINKED ISSUE #610: standards search: improve results visibility
Searching for standard ISO 15616-1 dies not yield the required results (as of the first ones), while using exact identifier `ISO-15616-1` works as expected. The search should show the correct standards also for the first search phrase

NRodriguezcuellar:
when searching by identifiers, the results are now still relevant even if the identifiers are not spelled exactly how the identifiers are saved

closes https://github.com/CERNDocumentServer/cds-ils/issues/610

### DIFF
--- a/invenio_app_ils/literature/search.py
+++ b/invenio_app_ils/literature/search.py
@@ -46,6 +46,7 @@ class LiteratureSearch(RecordsSearch):
 
     boosted_fields = [
         "identifiers.value^12.0",
+        "identifiers.value.text^12.0",
         "title^8.0",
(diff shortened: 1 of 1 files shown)


kpsherva (approved):

kpsherva:
what about the mapping of the series? You've mentioned an error appearing ?

NRodriguezcuellar:
> what about the mapping of the series? You've mentioned an error appearing ?

That was my bad, i forgot a comma hehe 😅  after that it worked correctly

PR #1128, opened 2021-10-29
Documents: allow ignoring part of the has references check

### LINKED ISSUE #608: import delete: display issues when deleting records with dependencies
from the user feedback:

> https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/118
No 2 and 32: no error shown in the preview, but error in the import. Document not updated (the EBC identifier should be removed). E-item well deleted but not showed in the import.

> Difference Preview / Import : https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/153 vs. https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/155 -> The documents are marked as "update" in the preview and "delete" in the import: this is fine because the import is correct, but still confusing not to have the same message in both modes.

> Difference Preview / Import : https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/153 vs. https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/155 -> Line 4 has no error in the preview and an error in the import

> https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/155 line 192 marked as 'update' but no link to the document. If document not found, it should be marked as non matched.

task: 
- [x] show the deleted eitem on errored deletes(when document cannot be deleted)
- [x]  show in the preview delete label on the document if document can be safely deleted
- [x] show error if document has dependencies (on delete preview)
- [x] fix update labels in not matched documents
- [x] display pid of deleted eitem both in preview and real delete.
- [x] display pid of created series in real import mode.

Pineirin:
closes https://github.com/CERNDocumentServer/cds-ils/issues/608

### DIFF
--- a/invenio_app_ils/documents/api.py
+++ b/invenio_app_ils/documents/api.py
@@ -175,16 +175,22 @@ class Document(IlsRecordWithRelations):
             document_pid=self["pid"],
         )
 
-    def has_references(self):
+    def has_references(self, excluded=None):
         """Check if record has references."""
-        return any([
-            self.search_loan_references().count(),
-            self.search_item_references().count(),
-            self.search_eitem_references().count(),
-            self.search_doc_req_references().count(),
-            self.search_order_references().count(),
-            self.search_brw_req_references().count()
-        ])
+        references = {
+            "loan": self.search_loan_references().count(),
+            "item": self.search_item_references().count(),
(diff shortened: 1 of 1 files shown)


Pineirin:
In the end, the change was not needed.

PR #1129, opened 2021-11-04
search: add keyword normaliser to mapping

kpsherva:
* normalize series title search
* normalize document search

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,10 @@
 
 Changes
 =======
+Version 1.0.0a55 (released 2021-11-05)
+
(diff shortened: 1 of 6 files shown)


jrcastro2 (approved):
LGTM!
Just wondering if it makes sense to also normalize other fields like `alternative_titles` ?

kpsherva:
> LGTM! Just wondering if it makes sense to also normalize other fields like `alternative_titles` ?

in my opinion not really since we will use it for very specific searches. The alternative titles are additional metadata, if it will be searched it will be anyway with text type, which already is normalised

ntarocco (approved):

PR #1132, opened 2021-11-10
Bulk extend: library can extend overdue loans

### LINKED ISSUE #621: Bug - extend all loans should be possible even for overdue loans
It is impossible to use this feature for loans that are already overdue, but it should be possible: it happens often that people renew after the overdue date. 

The rule should be that we are able to use this feature for loans overdue, or to expire in the next 3 days (and as it is already the case, when this is not an external loan and when the book is not requested by someone else).

NRodriguezcuellar:
Overdue loans can now be bulk extended in addition to soon expiring loans.

closes https://github.com/CERNDocumentServer/cds-ils/issues/621

### DIFF
--- a/invenio_app_ils/circulation/api.py
+++ b/invenio_app_ils/circulation/api.py
@@ -28,7 +28,7 @@ from invenio_pidstore.models import PIDStatus
 from invenio_pidstore.providers.recordid_v2 import RecordIdProviderV2
 
 from invenio_app_ils.circulation.search import (
-    get_all_range_expiring_ils_loans_by_patron_pid,
+    get_all_expiring_or_overdue_loans_by_patron_pid,
 )
 from invenio_app_ils.errors import (
     IlsException,
@@ -45,7 +45,6 @@ from invenio_app_ils.minters import pid_minter
 from invenio_app_ils.proxies import current_app_ils
 
 lt_es7 = ES_VERSION[0] < 7
-
 
 # override default `invenio-circulation` minters to use the base32 PIDs
 # CIRCULATION_LOAN_PID_TYPE is already defined in `invenio-circulation`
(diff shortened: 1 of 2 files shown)


invenio_app_ils/circulation/api.py
```diff
@@ -11,9 +11,10 @@
 from copy import copy, deepcopy
 from datetime import date, timedelta
 from functools import partial
+from itertools import chain
 
 from elasticsearch import VERSION as ES_VERSION
-from flask import current_app
+from flask import current_app, g
```
ntarocco:
```suggestion
from flask import current_app
```

invenio_app_ils/circulation/api.py
```diff
@@ -42,11 +44,11 @@
 from invenio_app_ils.fetchers import pid_fetcher
 from invenio_app_ils.items.api import Item
 from invenio_app_ils.minters import pid_minter
+from invenio_app_ils.permissions import backoffice_permission
```
ntarocco:
```suggestion

```

invenio_app_ils/circulation/api.py
```diff
@@ -260,9 +262,15 @@ def bulk_extend_loans(patron_pid, **kwargs):
     loans_search = get_all_range_expiring_ils_loans_by_patron_pid(
         expiring_in_days=days, patron_pid=patron_pid)
 
+    overdue_loans = get_all_overdue_loans() \
```
ntarocco: The method `get_all_range_expiring_ils_loans_by_patron_pid` was introduced in the context of the bulk extension feature. Why not changing that method instead of performing 2 searching (more efficient)?
`get_all_range_expiring_ils_loans_by_patron_pid` -> `get_expiring_and_overdue_loans`

ntarocco (commented):
I think there is no test modified or added... I think that test should have failed after this change? Can you please have a look?

kpsherva (approved):

NRodriguezcuellar:
> I think there is no test modified or added... I think that test should have failed after this change? Can you please have a look?

its is not failing because i only test the data i create myself in the test function :)

PR #1139, opened 2021-11-24
global: fixed application bootstrapping

NRodriguezcuellar:


setup.py
```diff
@@ -66,7 +66,8 @@
     "invenio-jsonschemas>=1.1.1,<1.2.0",
     "invenio-oaiserver>=1.3.0",
     "invenio-pidstore>=1.2.2,<1.3.0",
-    "invenio-records-rest>=1.7.0,<1.8.0",
+    "invenio-records-rest>=1.8.0,<1.9.0",
+    "invenio-rest>=1.2.4",
```
zzacharo: This is redundant I think :)

NRodriguezcuellar: are you sure that invenio-rest 1.2.4 is bottom pinned somewhere?

setup.py
```diff
@@ -66,7 +66,8 @@
     "invenio-jsonschemas>=1.1.1,<1.2.0",
     "invenio-oaiserver>=1.3.0",
     "invenio-pidstore>=1.2.2,<1.3.0",
-    "invenio-records-rest>=1.7.0,<1.8.0",
+    "invenio-records-rest>=1.9.0",
```
kpsherva:
```suggestion
    "invenio-records-rest>=1.9.0,<1.10.0",
```

setup.py
kpsherva: invenio-rest is bottom pinned in invenio-records-rest, you can remove from here

NRodriguezcuellar: alright :+1:

kpsherva (approved):

PR #1143, opened 2021-12-01
Circulation: allow sending args to invenio-records-rest csv serializer

Pineirin:


invenio_app_ils/circulation/serializers/csv.py
```diff
@@ -20,6 +20,10 @@
 class LoanCSVSerializer(CSVSerializer):
     """Serialize loan."""
 
+    def __init__(self, *args, **kwargs):
```
kpsherva, line 23: why is this constructor needed?

Pineirin:
This code is redundant.

PR #1145, opened 2021-12-07
Mappings: make imprint.publisher field case insensitive

### LINKED ISSUE #661: Field imprint.publisher is case-sensitive
## From the library
We have noticed that searches with upper vs. lower case do not give the same results for the field `imprint.publisher`. Front site and back office are affected.

Examples:
https://catalogue.library.cern/backoffice/documents?q=imprint.publisher%3A%28MDPI%29&l=list&order=asc&p=1&s=15&sort=bestmatch
vs.
https://catalogue.library.cern/backoffice/documents?q=imprint.publisher%3A%28mdpi%29&l=list&order=asc&p=1&s=15&sort=bestmatch

https://catalogue.library.cern/backoffice/documents?q=mdpi&l=list&order=asc&p=1&s=15&sort=bestmatch
vs.
https://catalogue.library.cern/backoffice/documents?q=MDPI&l=list&order=asc&p=1&s=15&sort=bestmatch

https://catalogue.library.cern/search?q=mdpi&l=grid&order=asc&p=1&s=15&sort=bestmatch
vs.
https://catalogue.library.cern/search?q=MDPI&l=grid&order=asc&p=1&s=15&sort=bestmatch

## Technical analysis

The field `imprint.publisher` is mapped as `keyword` type when searching, and the search query case is not correctly taken into account.

## TASKS:
- [x] leave a note to update sandbox/production indices after completing this task

Pineirin:
closes: CERNDocumentServer/cds-ils#661
Screenshot from 2021-12-07 11-51-41
Screenshot from 2021-12-07 11-51-47
Screenshot from 2021-12-07 14-28-49

### DIFF
--- a/invenio_app_ils/documents/mappings/v6/documents/document-v1.0.0.json
+++ b/invenio_app_ils/documents/mappings/v6/documents/document-v1.0.0.json
@@ -354,7 +354,12 @@
               "type": "keyword"
             },
             "publisher": {
-              "type": "keyword"
+              "type": "keyword",
+              "fields": {
+                "keyword": {
+                  "type": "text"
+                }
+              }
             },
             "reprint": {
               "type": "text"
--- a/invenio_app_ils/documents/mappings/v7/documents/document-v1.0.0.json
+++ b/invenio_app_ils/documents/mappings/v7/documents/document-v1.0.0.json
@@ -333,7 +333,12 @@
             "type": "keyword"
           },
           "publisher": {
-            "type": "keyword"
+            "type": "keyword",
+            "fields": {
+              "keyword": {
+                "type": "text"
+              }
+            }
           },
           "reprint": {
             "type": "text"


invenio_app_ils/documents/mappings/v6/documents/document-v1.0.0.json
```diff
@@ -354,7 +354,16 @@
               "type": "keyword"
             },
             "publisher": {
-              "type": "keyword"
+              "type": "text",
```
kpsherva: this change is quite complicated to introduce in production, read https://www.elastic.co/guide/en/elasticsearch/reference/7.15/indices-put-mapping.html#updating-field-mappings
so lets avoid modifying existing fields

invenio_app_ils/documents/mappings/v6/documents/document-v1.0.0.json
```diff
@@ -354,7 +354,16 @@
...
+              "fields": {
+                "keyword": {
+                  "type": "keyword"
+                },
+                "normalized_keyword": {
```
kpsherva: combination of text + keyword + normalised keyword is a bit overkill for this use case, simple text analyser is enough (to be tested)

kpsherva: there is still a simpler solution. The normalized keyword analyser was introduced for the specific use case of importer matching, and decreases searching possibilities (`"CErn"` search query would most likely not work) so I would not copy this solution

invenio_app_ils/documents/mappings/v6/documents/document-v1.0.0.json
```diff
@@ -9,6 +9,13 @@
             "lowercase"
           ]
         }
+      },
+      "analyzer": {
+        "analyzer_case_insensitive":{
```
kpsherva: I am not sure how this solution is different then keyword_normalised?

kpsherva (approved):

PR #1147, opened 2021-12-17
Documents: check if document exists when indexing

Pineirin:
Ensure application handles reindexing records related to a deleted document.
Issue appeared when reinxeding a document (and it's records) that was very recently deleted.

### DIFF
--- a/invenio_app_ils/circulation/indexer.py
+++ b/invenio_app_ils/circulation/indexer.py
@@ -14,6 +14,7 @@ from flask import current_app
 from invenio_circulation.pidstore.pids import CIRCULATION_LOAN_PID_TYPE
 from invenio_circulation.proxies import current_circulation
 from invenio_indexer.api import RecordIndexer
+from invenio_pidstore.errors import PIDDeletedError
 
 from invenio_app_ils.circulation.utils import resolve_item_from_loan
 from invenio_app_ils.documents.api import DOCUMENT_PID_TYPE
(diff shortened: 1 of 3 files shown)


invenio_app_ils/documents/indexer.py
```diff
@@ -63,9 +64,13 @@ def get_related_records(document_pid):
     """Get referenced records via relations."""
     referenced = []
     doc_record_cls = current_app_ils.document_record_cls
-    record = doc_record_cls.get_record_by_pid(document_pid)
+    try:
+        record = doc_record_cls.get_record_by_pid(document_pid)
+    except PIDDeletedError:
+        # Document might have been deleted while reindexing asyncronously.
+        return []
```
kpsherva: nit: maybe it would be better to instead of allocating memory for a new list
```suggestion
        return referenced
```

kpsherva (approved):

invenio_app_ils/circulation/indexer.py
```diff
@@ -86,19 +87,24 @@ def index_extra_fields_for_loan(loan_dict):
...
+        )
+    except PIDDeletedError:
+        # Document might have been deleted while reindexing asyncronously.
+        return
+
     document = document_record.replace_refs()
 
     items_available_for_loan_count = document.get("circulation", {}).get(
-        "available_items_for_loan_count"
+        "available_items_for_loan_count", 0
```
Pineirin: We should either include a default value or use [] if we always expect to get the value.
@kprzerwa I'm not sure if we always expect the value and, or, if the default value is needed, what should be it.

invenio_app_ils/circulation/indexer.py
```diff
@@ -86,19 +87,24 @@ def index_extra_fields_for_loan(loan_dict):
...
     )
     loan_dict[
         "available_items_for_loan_count"
     ] = items_available_for_loan_count
 
     can_circulate_items_count = document.get("circulation", {}).get(
-        "can_circulate_items_count"
+        "can_circulate_items_count", 0
```
Pineirin: Same comment as above.

kpsherva:
sorry, I find the commit message a bit confusing, because no error is raised in your code, so instead
```
Indexer: Handle errors when indexing a deleted document.
* ensure application raises error when a field that should be available isn't.
```

I would propose 
```
documents: check if document exists when indexing
```

ntarocco (approved):
LGTM! Let's see the answer to your comment about `.get` or `[]`

ntarocco (approved):
LGTM

kpsherva:
I would suggest only to remove or edit the commit message (the detailed description, not the title) since the description mentions raising an error, what does not happen in this PR

kpsherva (approved):

PR #1148, opened 2021-12-20
document: fix publisher mapping naming

kpsherva:


jrcastro2 (approved):

ntarocco (approved):
Just to be sure: there is no facet for this field, right?

kpsherva:
> Just to be sure: there is no facet for this field, right?

@ntarocco no we don't have facets for the publisher

Issue #316, opened 2019-05-06
Invalid propTypes for AggregatorValues


topless:
For `Documents`, `Items` and `Loans` searches the propTypes are incorrect

```
Invalid prop `values` of type `object` supplied to `AggregatorValues`, expected `array`
```

The AggregatorValues propTypes are defined in https://github.com/inveniosoftware/react-searchkit

kpsherva:
checked, the warning is not there anymore. Closing

PR #1159, opened 2022-07-14
logout: fix redirecting to 'Not found' page after logging out

anikachurilova:
* add a separate view to be shown right after logging out
* addresses [Sign out issue #316](https://github.com/inveniosoftware/react-invenio-app-ils/issues/316)


invenio_app_ils/views.py
```diff
@@ -39,3 +39,12 @@ def success_response(self, user):
             )
         )
         return jsonify(user_payload), 200
+
+
+logged_out_blueprint = Blueprint('logged_out', __name__)
```
ntarocco: This is great, but I am thinking that we could add an extra safe check and register this view **only** in dev env, WDYT? My fear is that this is registered and used on production envs, and we don't want that.

I would convert this to a function, similar to [here](https://github.com/inveniosoftware/invenio-app-ils/blob/master/invenio_app_ils/ill/views.py#L30):

Pseudocode below:
```
def create_root_blueprint(...):
   ....
   if current_app.config is dev:
       register_the_blueprint
       
   # otherwise do nothing
```

tests/conftest.py
```diff
@@ -121,3 +122,10 @@ def app_with_notifs(app):
     # override default app jinja_loader to add the new path
     app.jinja_loader = enhanced_jinja_loader
     yield app
+
+
+@pytest.fixture(scope="module")
```
ntarocco, line 127: I think that this should not be needed? Try to remove it, tests should pass

invenio_app_ils/config.py
```diff
@@ -207,6 +207,7 @@ def _(x):
 
 ACCOUNTS_REST_CONFIRM_EMAIL_ENDPOINT = "/accounts/confirm-email"
 
+SECURITY_POST_LOGOUT_VIEW = "/logged_out"
```
ntarocco: In the same spirit of the comment below, I would set this to `/logged_out` only if `dev` instance.
To be able to do that, you will have to move it to the `ext.py` on app initialization. This is because you need the `current_app` context, which you don't have in the config file when parsed (the app is not running).

I would also add a nice comment explaining why it is needed. WDYT?

invenio_app_ils/views.py
```diff
@@ -39,3 +39,12 @@ def success_response(self, user):
...
+
+
+@logged_out_blueprint.route('/logged_out')
```
ntarocco: `dash` nicer than `underscore` in the URL, WDYT?
```suggestion
@logged_out_blueprint.route('/logged-out')
```

tests/test_post_logout_redirect.py
```diff
@@ -0,0 +1,23 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Simple test of correct patron redirection after being logged out."""
+
+from flask import url_for
+
+from tests.helpers import user_login
+
+
+def test_correct_redirection_after_logout(client, users):
```
ntarocco, line 15: Good job! 👏🏼 👏🏼 👏🏼 
Thank you for adding the test!

ntarocco (commented):
Very good job, thank you! I have added a few comments just to be extra safe, we can discuss them IRL. :)

PR #1181, opened 2023-02-17
Backend: Bump the python version to 3.11

hudymi:
### Description

Updated the Python version to the latest in the backend, as version 3.6 is no longer supported. Also, there is a problem with running 3.6 version on the latest Apple processors (like M2)


kpsherva:
Thank you for the contribution and subsequently bringing the upgrade to our attention.
Closing as the base image was finally changed to almalinux, aligning with the rest of Invenio products

PR #1191, opened 2024-01-08
circulation: Made loan request start date to be configurable and validated at the backend

### LINKED ISSUE #835: Require before a certain date
People can now select very near dates for loan requests in 'I require it before a certain date', which means that if they place the request in the evening and select the same day, the loan request will be cancelled at midnight and we will not be able to fulfill it. 

**Describe the solution you'd like**
Grey today and tomorrow, so people can only place request in 2 days which let us time to process it. If we are November 7:

image

sakshamarora1:
:heart: Thank you for your contribution!

### Description

Closes: https://github.com/CERNDocumentServer/cds-ils/issues/835

### DIFF
--- a/invenio_app_ils/circulation/config.py
+++ b/invenio_app_ils/circulation/config.py
@@ -74,8 +74,8 @@ from .utils import (
 ILS_CIRCULATION_NOTIFICATION_OVERDUE_REMINDER_INTERVAL = 3
 #: The maximum duration of a loan request
 ILS_CIRCULATION_LOAN_REQUEST_DURATION_DAYS = 60
-#: Loan Request Offset Value
-ILS_CIRCULATION_LOAD_REQUEST_OFFSET = 0
+#: The minimum number of days after which a loan can start from the date of request
+ILS_CIRCULATION_LOAN_REQUEST_OFFSET = 0
 #: Period of time in days, before loans expire, for notifications etc.
 ILS_CIRCULATION_LOAN_WILL_EXPIRE_DAYS = 7
 #: Optional delivery methods when requesting a new loan. Set to empty object to
--- a/invenio_app_ils/circulation/config.py
+++ b/invenio_app_ils/circulation/config.py
@@ -74,6 +74,8 @@ from .utils import (
 ILS_CIRCULATION_NOTIFICATION_OVERDUE_REMINDER_INTERVAL = 3
 #: The maximum duration of a loan request
 ILS_CIRCULATION_LOAN_REQUEST_DURATION_DAYS = 60
+#: Loan Request Offset Value
+ILS_CIRCULATION_LOAD_REQUEST_OFFSET = 0
 #: Period of time in days, before loans expire, for notifications etc.
 ILS_CIRCULATION_LOAN_WILL_EXPIRE_DAYS = 7
 #: Optional delivery methods when requesting a new loan. Set to empty object to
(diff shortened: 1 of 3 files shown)


invenio_app_ils/circulation/loaders/schemas/json/loan_request.py
```diff
@@ -104,4 +105,12 @@ def postload_checks(self, data, **kwargs):
                     "request_expire_date": [message],
                 }
             )
+        elif end - start < loan_request_offset:
+            message = "The request end date can only be {} days after the request start date.".format(loan_request_offset.days)
```
jrcastro2: Question: Is this error message correct? If I understand correctly it should be something like:
```suggestion
            message = "The requested start date must be at least {} days from today.".format(loan_request_offset.days)

```
""""
Or am I missing something?

invenio_app_ils/circulation/loaders/schemas/json/loan_request.py
```diff
@@ -104,4 +105,12 @@ def postload_checks(self, data, **kwargs):
...
+            raise ValidationError(
+                {
+                    "request_start_date": [message],
+                    "request_expire_date": [message],
```
jrcastro2:
```suggestion
```
Comment: I think the error is only on the start_date, unless I am missing something

invenio_app_ils/circulation/config.py
```diff
@@ -74,6 +74,8 @@
 ILS_CIRCULATION_NOTIFICATION_OVERDUE_REMINDER_INTERVAL = 3
 #: The maximum duration of a loan request
 ILS_CIRCULATION_LOAN_REQUEST_DURATION_DAYS = 60
+#: Loan Request Offset Value
+ILS_CIRCULATION_LOAD_REQUEST_OFFSET = 0
```
ntarocco:
```suggestion
ILS_CIRCULATION_LOAN_REQUEST_OFFSET = 0
```
Can you also please describe a little bit better what this config will do? `Offset` might not very clear to everyone.

invenio_app_ils/circulation/loaders/schemas/json/loan_request.py
```diff
@@ -104,4 +105,12 @@ def postload_checks(self, data, **kwargs):
                     "request_expire_date": [message],
                 }
             )
+        elif end - start < loan_request_offset:
```
ntarocco, line 121: Was there any unit/functional test on this? Would it be possible to add tests?

In particular, I would test:
- if I put a negative value: it looks like you can do that, we should block it
- if I put `0`
- if I put a few days
- what happens if I set the config to `2 days` and I make the request now at 11am: will the start date be in 2 days, at 11am (exactly 48 hours), or the check will only compare days and I need to select the third day?

invenio_app_ils/circulation/loaders/schemas/json/loan_request.py
sakshamarora1: This is indeed more straightforward, will update this, thank you!

invenio_app_ils/circulation/config.py
sakshamarora1: oops, I misspelled, thanks!

I will update the comment to describe it clearly, is this okay?
```
#: The minimum number of days after which a loan can start from the date of request
```

invenio_app_ils/circulation/loaders/schemas/json/loan_request.py
sakshamarora1: I will look into the unit test addition for this.

Regarding the 4th point: It will disable today and tomorrow only. The check only compares on the days.

PR #1192, opened 2024-01-11
anonymization: fix orders with patron_pid

ntarocco:
* fixes an error when anonymizing an order that does not have the patron_pid in the order lines


ntarocco:
Closing in favour of: https://github.com/inveniosoftware/invenio-app-ils/pull/1193

PR #1197, opened 2024-02-19
facets: fix keyed_filter query import

kpsherva:


jrcastro2 (approved):
LGTM! Only thing are ES tests failing as dependency was removed

PR #1202, opened 2024-02-22
relations: Add functionality to sort json refs by relation_type

### LINKED ISSUE #827: Edition not always in order in front office
In front office. I notice that sometimes editions are not in the right order, [for example](https://catalogue.library.cern/literature/z0maq-x0x40), editions are in this order: ed. 4th; ed. 3rd; ed. 1st.
editions_order
Another example with more complex editions names: https://catalogue.library.cern/literature/3ngs2-mve49

decision: sort by publication year

sakshamarora1:
:heart: Thank you for your contribution!

### Description

Closes: https://github.com/CERNDocumentServer/cds-ils/issues/827

### DIFF
--- a/invenio_app_ils/documents/jsonresolvers/document_relations.py
+++ b/invenio_app_ils/documents/jsonresolvers/document_relations.py
@@ -26,7 +26,7 @@ def jsonresolver_loader(url_map):
     def relations_resolver(document_pid):
         """Resolve record relations and add metadata."""
         document = Document.get_record_by_pid(document_pid)
-        return document.relations
+        return document.relations(sort_by = ["edition"])
 
     url_map.add(
         Rule(
--- a/invenio_app_ils/documents/jsonresolvers/document_relations.py
+++ b/invenio_app_ils/documents/jsonresolvers/document_relations.py
@@ -26,7 +26,7 @@ def jsonresolver_loader(url_map):
     def relations_resolver(document_pid):
         """Resolve record relations and add metadata."""
         document = Document.get_record_by_pid(document_pid)
-        return document.relations(sort_by = ["edition"])
+        return document.relations
 
     url_map.add(
         Rule(
--- a/invenio_app_ils/records_relations/api.py
+++ b/invenio_app_ils/records_relations/api.py
@@ -421,12 +421,10 @@ class RecordRelationsSequence(RecordRelations):
 class IlsRecordWithRelations(IlsRecord):
     """Add relations functionalities to records."""
 
-    @property
-    def relations(self):
+    def relations(self, sort_by=None):
         """Get record relations."""
         from .retriever import get_relations
-
-        return get_relations(self)
+        return get_relations(self, sort_by)
 
     def clear(self):
         """Clear IlsRecordWithRelations record."""
@@ -438,7 +436,7 @@ class IlsRecordWithRelations(IlsRecord):
     def delete(self, **kwargs):
         """Delete record with relations."""
         related_refs = set()
-        for name, related_objects in self.relations.items():
+        for name, related_objects in self.relations().items():
             for obj in related_objects:
                 related_refs.add("{pid_value}:{pid_type}".format(**obj))
         if related_refs:
--- a/invenio_app_ils/records_relations/api.py
+++ b/invenio_app_ils/records_relations/api.py
@@ -421,10 +421,11 @@ class RecordRelationsSequence(RecordRelations):
 class IlsRecordWithRelations(IlsRecord):
     """Add relations functionalities to records."""
 
-    def relations(self, sort_by=None):
+    @property
+    def relations(self):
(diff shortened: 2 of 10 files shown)


invenio_app_ils/documents/jsonresolvers/document_relations.py
```diff
@@ -26,7 +26,7 @@ def jsonresolver_loader(url_map):
     def relations_resolver(document_pid):
         """Resolve record relations and add metadata."""
         document = Document.get_record_by_pid(document_pid)
-        return document.relations
+        return document.relations(sort_by = ["edition"])
```
jrcastro2: Question: I guess that `sort_by` is a list because we want to be able to define a different sort type for the different type of relations?

invenio_app_ils/records_relations/retriever.py
```diff
@@ -170,6 +170,11 @@ def get(self):
                 r = self._build_relation_obj(sibling_pid, name)
                 relations.setdefault(name, [])
                 relations[name].append(r)
+
+            # Pre-requisite: fields being sorted should be of the same type.
+            if sort_by and name in sort_by and name in relations.keys():
```
jrcastro2: Question: Does this mean that the `EDITION_RELATION` can only be sorted by `edition`, `LANGUAGE_RELATION` only by `language` and `OTHER_RELATION` by `other`?

jrcastro2 (commented):
Left a couple of questions as it looks a bit complex :sweat_smile:

invenio_app_ils/documents/jsonresolvers/document_relations.py
sakshamarora1: Yes, as relations have a separate object list in the response

invenio_app_ils/records_relations/retriever.py
sakshamarora1: Yes!
Do we want the ability to sort by another field apart from the relation type?

invenio_app_ils/records_relations/api.py
```diff
@@ -421,12 +421,10 @@ def remove(self, previous_rec, next_rec, relation_type):
 class IlsRecordWithRelations(IlsRecord):
     """Add relations functionalities to records."""
 
-    @property
-    def relations(self):
+    def relations(self, sort_by=None):
```
kpsherva: this is a breaking change in the class interface, did you test all the possible occurences?

kpsherva: in addition: I think the sort should be somewhat defined per relation type (maybe config of relation types?) not as an argument of the function - we risk the list of arguments growing infinitely

ntarocco: we could go for `kwargs` instead:
```suggestion
    def relations(self, **kwargs):
```

invenio_app_ils/records_relations/retriever.py
```diff
@@ -224,14 +229,14 @@ def get(self):
         return relations
 
 
-def get_relations(record):
+def get_relations(record, sort_by=None):
```
ntarocco:
```suggestion
def get_relations(record, **kwargs):
```
and also inject `kwargs` in all the `.get()` methods

invenio_app_ils/records_relations/retriever.py
```diff
@@ -170,6 +170,11 @@ def get(self):
                 r = self._build_relation_obj(sibling_pid, name)
                 relations.setdefault(name, [])
                 relations[name].append(r)
+
+            # Pre-requisite: fields being sorted should be of the same type.
+            if sort_by and name in sort_by and name in relations.keys():
+                relations[name].sort(key=lambda rec: rec["record_metadata"][name])
```
ntarocco: This might not good enough: we might be sorting too late, sorting the results of what was retrieved from the db, queried with a default sorting.

If this is what we want to do, then it might better to have the client, consuming the REST API response, to sort the results.
If instead we want to drive the sorting from the backend, then the sorting param should be applied to the db query.
However, this is very tricky, due to the nature of data stored as JSON blob.

ntarocco: Discussed IRL, decision to be taken.

invenio_app_ils/relations/api.py
```diff
@@ -27,6 +27,7 @@
     "invenio_app_ils.relations.nodes:PIDNodeRelated",
     "invenio_pidrelations.serializers.schemas.RelationSchema",
     "invenio_app_ils.relations.api.SiblingsRelation",
+    [],
```
kpsherva:
```suggestion
    None,
```
Setting mutable parameters is in general considered as bad practice, since we can change it on a runtime - especially in this case it is configuration so we don't want to allow for changes

ntarocco (approved):

PR #1209, opened 2024-05-16
records: loaders: schemas: Move IdentifierSchema from documents

sakshamarora1:


kpsherva (approved):
Nice! thank you for improving the schemas

PR #1213, opened 2024-06-03
circulation: Support self checkout by patrons

### LINKED ISSUE #839: Self-checkout station
As the library is open 24/7, people also would like to borrow outside the opening hours. A self-checkout station would allow people to register themselves their loans in the library before leaving with the books.

**Describe the solution you'd like**
Examples of devices: 
https://www.bibliotheca.com/solutions/self-service-checkout/
https://d-techinternational.com/products/self-service-2/
but will require an integration with the Library catalogue.

**Describe alternatives you've considered**
Using a standard PC and barcode reader.

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/839

### DIFF
--- a/invenio_app_ils/circulation/config.py
+++ b/invenio_app_ils/circulation/config.py
@@ -47,6 +47,7 @@ from invenio_app_ils.permissions import (
     loan_extend_circulation_permission,
     patron_owner_permission,
     superuser_permission,
+    loan_checkout_permission,
 )
 
 from .api import ILS_CIRCULATION_LOAN_FETCHER, ILS_CIRCULATION_LOAN_MINTER
@@ -161,7 +162,7 @@ CIRCULATION_LOAN_TRANSITIONS = {
             dest="ITEM_ON_LOAN",
             trigger="checkout",
             transition=ILSToItemOnLoan,
-            permission_factory=backoffice_permission,
+            permission_factory=loan_checkout_permission,
         ),
     ],
     "PENDING": [
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -865,6 +865,7 @@ RECORDS_REST_FACETS = dict(
 
 # ILS
 # ===
+
 ILS_VIEWS_PERMISSIONS_FACTORY = views_permissions_factory
 """Permissions factory for ILS views to handle all ILS actions."""
 
@@ -1071,3 +1072,6 @@ ILS_PATRON_ANONYMOUS_CLASS = AnonymousPatron
 ILS_PATRON_SYSTEM_AGENT_CLASS = SystemAgent
 
 DB_VERSIONING_USER_MODEL = None
+
+# Feature Toggles
+ILS_SELF_CHECKOUT_ENABLED = False
--- a/invenio_app_ils/permissions.py
+++ b/invenio_app_ils/permissions.py
@@ -16,7 +16,7 @@ from invenio_access import action_factory
 from invenio_access.permissions import Permission, authenticated_user, superuser_access
 from invenio_records_rest.utils import allow_all, deny_all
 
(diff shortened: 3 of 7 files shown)


invenio_app_ils/circulation/config.py
```diff
@@ -315,3 +315,6 @@
         },
     )
 )
+
+# Feature Toggles
+CIRCULATION_SELF_CHECKOUT_ENABLED = SELF_CHECKOUT_ENABLED
```
kpsherva: Could you clarify why is this assignment needed?

invenio_app_ils/config.py
```diff
@@ -1071,3 +1078,6 @@ def _(x):
 ILS_PATRON_SYSTEM_AGENT_CLASS = SystemAgent
 
 DB_VERSIONING_USER_MODEL = None
+
+# Feature Toggles
+SELF_CHECKOUT_ENABLED = False
```
kpsherva: to keep the config variables naming consistent:
```suggestion
ILS_SELF_CHECKOUT_ENABLED = False
```

invenio_app_ils/circulation/config.py
sakshamarora1: Since, circulation is an extension it uses the circulation config and not the main config, so this is needed to get the value from the main config.

invenio_app_ils/permissions.py
```diff
@@ -141,11 +141,9 @@ def __init__(self, record):
 
 def views_permissions_factory(action):
     """Return ILS views permissions factory."""
-    is_authenticated_user = [
-        "circulation-loan-request",
-        "patron-loans",
-        "bulk-loan-extension",
-    ]
+    is_authenticated_user = current_app.config.get(
+        "ILS_AUTHENTICATED_USER_PERMISSIONS", []
```
kpsherva: well done for caring for flexibility of the configuration!
In this case it is a bit overkill though, because we already had the config:
```
ILS_VIEWS_PERMISSIONS_FACTORY = views_permissions_factory
```

invenio_app_ils/circulation/config.py
kpsherva: I wouldn't fit this variable under circulation because we are not bringing this change to the circulation module. The prefixes like `CIRCULATION_` or `ILS_` are meant to hint from which module the variable is coming from

invenio_app_ils/circulation/api.py
```diff
@@ -192,6 +193,19 @@ def checkout_loan(
         circulate.
     """
     loan_cls = current_circulation.loan_record_cls
+    # In case of self checkout confirm:
+    #   1. self checkout feature flag is enabled
+    #   2. The patron ids match
+    if (
+        kwargs.get("self_checkout", False)
```
kpsherva: just a reminder: we agreed to check if the action is performed by the librarian, not to pass the kwarg from the UI, since it is more reliable check.

invenio_app_ils/circulation/api.py
```diff
@@ -192,6 +193,19 @@ def checkout_loan(
...
+        and current_app.config.get("CIRCULATION_SELF_CHECKOUT_ENABLED", False)
+        and patron_pid != current_user.id
+    ):
```
kpsherva: for readability (check if the logic is correct ;) the point is to have less content heavy conditional statements to be able to understand the bussiness logic behind it easier:
```suggestion
    self_checkout_enabled = current_app.config.get("CIRCULATION_SELF_CHECKOUT_ENABLED", False)
    is_librarian = .... check if librarian
    is_patron_current_user = patron_pid == current_user.id
    if self_checkout_enabled and (not is_librarian or not is_patron_current_user):
```

PR #1214, opened 2024-06-05
documents: mappings: normalize accents and strip special chars

### LINKED ISSUE #543: Search on strings with accent, apostrophe, ampersand

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/543

docs for reference:
https://opensearch.org/docs/latest/analyzers/#built-in-analyzers
https://opensearch.org/docs/latest/analyzers/token-filters/index/

`strip_special_chars` is a custom char_filter which uses pattern_replace (https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-charfilters.html)

### DIFF
--- a/invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
+++ b/invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
@@ -2,11 +2,14 @@
   "settings": {
     "analysis": {
       "normalizer": {
-        "lowercase_normalizer": {
+        "custom_normalizer": {
           "type": "custom",
-          "char_filter": [],
+          "char_filter": [
+            "strip_special_chars"
+          ],
           "filter": [
-            "lowercase"
+            "lowercase",
+            "asciifolding"
           ]
         }
       },
@@ -14,12 +17,33 @@
         "min_length_2": {
           "tokenizer": "standard",
           "filter": [ "min_length_2" ]
+        },
+        "custom_analyzer": {
+          "tokenizer": "standard",
+          "type": "custom",
+          "char_filter": [
+            "strip_special_chars"
+          ],
+          "filter": [
+            "lowercase",
+            "asciifolding"
+          ]
         }
       },
       "filter": {
         "min_length_2": {
           "type": "length",
           "min": 2
+        },
+        "asciifolding": {
+          "type": "asciifolding"
+        }
+      },
+      "char_filter": {
+        "strip_special_chars": {
+          "type": "pattern_replace",
+          "pattern": "[\\p{Punct}\\p{S}]",
+          "replacement": ""
         }
       }
     }
@@ -123,6 +147,10 @@
                 "type": "text",
                 "analyzer": "min_length_2",
                 "search_analyzer": "min_length_2"
+              },
+              "normalized_keyword": {
+                "type": "keyword",
+                "normalizer": "custom_normalizer"
               }
             }
           },
@@ -617,7 +645,9 @@
                 "type": "keyword"
               },
               "title": {
-                "type": "text"
+                "type": "text",
+                "analyzer": "custom_analyzer",
+                "search_analyzer": "custom_analyzer"
               }
             },
             "type": "object"
@@ -638,7 +668,9 @@
                 "type": "keyword"
               },
               "title": {
-                "type": "text"
+                "type": "text",
+                "analyzer": "custom_analyzer",
+                "search_analyzer": "custom_analyzer"
               }
             },
             "type": "object"
@@ -656,7 +688,9 @@
                 "type": "keyword"
               },
               "title": {
-                "type": "text"
+                "type": "text",
+                "analyzer": "custom_analyzer",
+                "search_analyzer": "custom_analyzer"
               },
               "volume": {
                 "type": "keyword",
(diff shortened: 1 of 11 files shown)


invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
```diff
@@ -2,24 +2,48 @@
...
           "min": 2
+        },
+        "asciifolding": {
+          "type": "asciifolding"
+        }
+      },
+      "char_filter": {
+        "strip_special_chars": {
+          "type": "pattern_replace",
+          "pattern": "[\\p{Punct}\\p{S}]",
```
sakshamarora1: Update this with an exhaustive list of special chars if that is preferred.

invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
```diff
@@ -617,7 +645,9 @@
                 "type": "keyword"
               },
               "title": {
-                "type": "text"
+                "type": "text",
```
sakshamarora1: Check with all on what fields need accent removing, and update accordingly based on review

invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
```diff
@@ -2,24 +2,48 @@
   "settings": {
```
sakshamarora1: Currently updated in document-v1.0.0 for better review of the changes
Move to v2.0.0 when finalised

sakshamarora1 (commented):
Currently I have only added the filters for documents index for an initial review
Will add for the other indices after confirming with an exhaustive list.

invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
```diff
@@ -2,24 +2,48 @@
...
     "analysis": {
       "normalizer": {
-        "lowercase_normalizer": {
+        "custom_normalizer": {
           "type": "custom",
-          "char_filter": [],
+          "char_filter": [
+            "strip_special_chars"
```
kpsherva: just a question to clarify: will this strip the characters when sending the search query or when indexing?

invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
kpsherva: maybe a regex already findable in some documentation?

invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
kpsherva: this will be quite intense change, if we start considering all of the fields. I would leave it at the title for the moment, and seek feedback if it helped.

invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
sakshamarora1: Since I mention this in the custom_analyzer which I mention for both index analysing and search analysing, so both!

invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
sakshamarora1: I used this for reference: https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html

invenio_app_ils/documents/mappings/os-v1/documents/document-v1.0.0.json
sakshamarora1: Just the main title or the other title fields too (as it is currently)?

zzacharo (approved):
This seems good! Maybe we can run a series of tests when we deploy this on sandbox!

invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
```diff
@@ -0,0 +1,978 @@
...
+            "asciifolding"
+          ]
+        }
+      },
+      "analyzer": {
+        "min_length_2": {
+          "tokenizer": "standard",
+          "filter": [ "min_length_2" ]
+        },
+        "custom_analyzer": {
```
kpsherva, line 21: correct me if I am wrong, but if we use this analyzer as search_analyzer per field, ie. for `title` it will work only if the search query will be `title:l'occasion` not `l'occasion`. "Global" query string `l'occasion` will still fail on the apostrophe, right?
are we missing global search analyzer somewhere ?

sakshamarora1: We are using this custom_analyser as an `index_analyzer` too,  so it will work for both, lucene syntax and global.

kpsherva:
alternative I found in the docs, in case they were not tested:

- could language analyzers help with accents and apostrophies ? https://opensearch.org/docs/latest/analyzers/language-analyzers/

ntarocco (commented):
It looks good to me, nice work. However, I had the feeling that you will have to apply the same to the mappings resolving a document, for example when eitem/item we resolve the document. WDYT?

PR #1216, opened 2024-06-19
search: use AND query parser instead of default operator OR

### LINKED ISSUE #529: Improve the search behaviour
Some users might be confused by the search behaviour, it is high visible for the standards (#516), but it creates confusion on other document types as well.  

2 aspects seem to be improved: 
- When searching 2 or more terms in the search, the implicite boolean should be AND and not OR (this is the case in Inspire)
- The order of the terms shouldn't have an impact on the search results, especially if they are part of the title and the author. 

Some examples are explained here: 
https://docs.google.com/document/d/1Xw4I4sOwHTNqKYR6JrhOfMkPwlNTuE4cmZgGKmbTzDE/edit#

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/529

### DIFF
--- a/invenio_app_ils/search_permissions.py
+++ b/invenio_app_ils/search_permissions.py
@@ -53,15 +53,17 @@ def search_filter_record_permissions():
     return dsl.Q("bool", filter=[combined_filter])
 
 
-def ils_search_factory(self, search, validator=None):
-    """Search factory with Query String validator.
+def ils_search_factory(self, search, query_parser=None, validator=None):
+    """Search factory with Query String parser and validator.
 
     :param self: REST view.
     :param search: Elastic search DSL search instance.
+    :param query_parser: Custom Query parser DSL.
+    :param validator: Custom Validator for query string.
     :returns: Tuple with search instance and URL arguments.
     """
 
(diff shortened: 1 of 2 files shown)


invenio_app_ils/search_permissions.py
```diff
@@ -53,15 +53,17 @@ def search_filter_record_permissions():
     return dsl.Q("bool", filter=[combined_filter])
 
 
-def ils_search_factory(self, search, validator=None):
-    """Search factory with Query String validator.
+def ils_search_factory(self, search, query_parser=None, validator=None):
```
kpsherva, line 56: is there any reason why you didn't add the new kwarg at the end of the function signature? I am worried that it will break the usages if it is passed implicitly as argument not kwarg

sakshamarora1: Yes, because we are passing this function to the config `search_factory_imp` where `invenio_records_rest` passes the query parser/modifier as the 2nd param.
Ref: https://github.com/inveniosoftware/invenio-records-rest/pull/329/files

PR #1221, opened 2024-10-07
mapping: update mappings

jrcastro2:
* add custom analyzer to multiple fields for documents and series, needed to be used with cross_fields type search

### DIFF
--- a/invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
+++ b/invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
@@ -16,7 +16,9 @@
       "analyzer": {
         "min_length_2": {
           "tokenizer": "standard",
-          "filter": [ "min_length_2" ]
+          "filter": [
+            "min_length_2"
+          ]
         },
         "custom_analyzer": {
           "tokenizer": "standard",
@@ -70,10 +72,14 @@
         "type": "date"
       },
       "abstract": {
-        "type": "text"
+        "type": "text",
+        "analyzer": "custom_analyzer",
+        "search_analyzer": "custom_analyzer"
       },
       "alternative_abstracts": {
-        "type": "text"
+        "type": "text",
+        "analyzer": "custom_analyzer",
+        "search_analyzer": "custom_analyzer"
       },
       "alternative_identifiers": {
         "properties": {
@@ -98,7 +104,9 @@
             "type": "keyword"
           },
           "value": {
-            "type": "text"
+            "type": "text",
+            "analyzer": "custom_analyzer",
+            "search_analyzer": "custom_analyzer"
           }
         },
         "type": "object"
@@ -120,6 +128,8 @@
               },
               "name": {
                 "type": "text",
+                "analyzer": "custom_analyzer",
+                "search_analyzer": "custom_analyzer",
                 "fields": {
                   "keyword": {
                     "type": "keyword"
@@ -131,6 +141,8 @@
           },
           "alternative_names": {
             "type": "text",
+            "analyzer": "custom_analyzer",
+            "search_analyzer": "custom_analyzer",
             "fields": {
               "keyword": {
                 "type": "keyword"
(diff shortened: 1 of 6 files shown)


invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
```diff
@@ -84,7 +90,14 @@
...
           "value": {
-            "type": "keyword"
+            "type": "text",
+            "analyzer": "custom_analyzer",
+            "search_analyzer": "custom_analyzer",
+            "fields": {
+              "keyword": {
+                "type": "keyword"
+              }
+            }
```
kpsherva: normally, in order to avoid problems with reindexing, this could be done in opposite way

```suggestion
            "type": "keyword",
            "fields": {
              "text": {
                "type": "text",
                "analyzer": "custom_analyzer",
                "search_analyzer": "custom_analyzer",
              }
            }
```

in this way you add a new field instead of changing the type of the existing one.
I don't think we need the analyzer for this field

invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
```diff
@@ -113,13 +128,22 @@
                     "type": "keyword"
                   },
                   "value": {
-                    "type": "keyword"
+                    "type": "text",
```
kpsherva: same comment as above, please remove the analyser

invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
```diff
@@ -145,12 +171,17 @@
...
-                "analyzer": "min_length_2",
-                "search_analyzer": "min_length_2"
+                "analyzer": "custom_analyzer",
+                "search_analyzer": "custom_analyzer"
               },
               "normalized_keyword": {
                 "type": "keyword",
                 "normalizer": "custom_normalizer"
+              },
+              "min_length_2": {
```
kpsherva, line 167: I am not fully understanding why would you add this as a field, could you explain what is the difference?

invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
```diff
@@ -521,7 +592,14 @@
                 "type": "keyword"
               },
               "title": {
-                "type": "keyword"
+                "type": "keyword",
+                "fields": {
```
kpsherva, line 584: note to myself: how likely would it be someone searches by license?

invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
```diff
@@ -563,9 +643,9 @@
           "journal_title": {
             "type": "keyword",
             "fields": {
-                "text": {
-                  "type": "text"
-                }
+              "text": {
```
kpsherva, line 634: I think in this case it would make sense to add the analyser

invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
```diff
@@ -891,7 +985,14 @@
             "type": "keyword"
           },
           "value": {
-            "type": "keyword"
+            "type": "keyword",
+            "fields": {
```
kpsherva: I think it will be unlikely to search by URL, we could remove this part, @ntarocco WDYT?

kpsherva (commented):
I left couple of comments, please check. 
Do we need to do the same with items mappings? Or it does not have custom analysers?

invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
jrcastro2: I applied the analyzer as well to this field to be used by the cross_fields type, while at the same time keeping the behaviour of before by keeping the min_length_2 (which will index tokens of at least 2 characters). 

Alternatively we could add the min_length_2 in the `custom_analyzer` but it would imply that all the rest of the fields using this would also start indexing tokens of at least 2, which might not be needed for many of them

jrcastro2:
> I left couple of comments, please check. Do we need to do the same with items mappings? Or it does not have custom analysers?

Items do have the custom_analyzer applied only to the title. Let me know if we should add it as well

invenio_app_ils/documents/mappings/os-v1/documents/document-v2.0.0.json
kpsherva: it won't work well for authors names, because some names are 2 char long so I wouldn't apply it to all the fields.
my hesitation is that I find it a bit strange to have a dedicated field for this, since in the other fields by default we will have more tokens (not only the ones longer than 2 chars)

PR #1222, opened 2024-10-16
permissions: add self-checkout

ntarocco:


invenio_app_ils/permissions.py
```diff
@@ -134,25 +134,30 @@ def patron_owner_permission(record):
...
         return backoffice_permission()
+
+    # ensure that only the loan's patron can do operations on this loan
     if len(args):
         loan = args[0]
     else:
-        loan = kwargs.get("record", {})
-    is_patron_current_user = current_user.id == int(loan.get("patron_pid"))
+        loan = kwargs["record"]
+    is_patron_current_user = current_user.id == int(loan["patron_pid"])
```
ntarocco, line 152: We should really avoid using `.get` when the dict field is required. We need to fail early in these cases.

invenio_app_ils/permissions.py
```diff
@@ -163,36 +168,41 @@ def __init__(self, record):
         super().__init__(UserNeed(int(record["patron_pid"])), backoffice_access_action)
 
 
+_is_authenticated_user = [
```
ntarocco, line 168: Moved outside to declare them once, and not on each func call.

setup.cfg
```diff
@@ -28,6 +28,7 @@ zip_safe = False
 install_requires =
     # Invenio core modules
     invenio-app>=1.3.4,<1.4.0
+    invenio-db[postgresql,mysql]>=1.0.14,<2.0.0
```
kpsherva: won't this install both dbs? also, seems like versioning is missing?

tests/api/ils/items/test_items_update.py
```diff
@@ -26,6 +26,11 @@ def get_active_loan_pid_and_item_pid():
                 if total > 0:
                     return t["pid"], active_loan[0]["pid"]
 
+    item = Item.get_record_by_pid("itmeid-1")
```
kpsherva: typo, probably this is breaking tests

setup.cfg
ntarocco: This will install only the drivers/packages, we do this everywhere now.
Versioning was optional, it is now [installed by default](https://github.com/inveniosoftware/invenio-db/blob/master/setup.cfg#L37).

tests/api/ils/items/test_items_update.py
ntarocco: Thanks a lot!

kpsherva (approved):

sakshamarora1 (approved):

PR #1231, opened 2025-05-13
Closings: Added endpoint to return date ranges during which a location is closed

### LINKED ISSUE #967: Do not selection of holidays in the calendar in Loans
In back office. For now, we still have the possibility to select holidays and weekend days for the return date when modifying the loan dates manually in "Loans and requests" (https://catalogue.library.cern/backoffice/loans?q=&l=list&order=desc&p=1&s=15&sort=created).
This is a risk for errors and it is confusing for both librarians and patrons, as there is a risk that a librarian select a holiday or a Saturday/Sunday for the return date. For example here the 1st May or even the weekend days are clickable, but it should be grayed:
Image

JakobMiesner:
Closes: https://github.com/CERNDocumentServer/cds-ils/issues/967

### DIFF
--- a/invenio_app_ils/closures/views.py
+++ b/invenio_app_ils/closures/views.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2025-2025 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio App ILS Closure views."""
+
+from flask import Blueprint, abort
+from flask_login import current_user
+from datetime import date
+from invenio_records_rest.views import pass_record
+from invenio_rest import ContentNegotiatedMethodView
+from invenio_app_ils.locations.api import LOCATION_PID_TYPE
+from invenio_app_ils.closures.serializers import closure_periods_response
+from invenio_app_ils.closures.api import get_closure_periods
+from invenio_app_ils.permissions import backoffice_permission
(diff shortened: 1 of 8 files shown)


invenio_app_ils/closures/views.py
```diff
@@ -0,0 +1,66 @@
...
+from invenio_app_ils.closures.serializers import closure_periods_response
+from invenio_app_ils.closures.api import get_closure_periods
+from invenio_app_ils.permissions import backoffice_permission
+from invenio_app_ils.records.permissions import RecordPermission
+
+
+def create_closures_blueprint(app):
+    """Create location closed periods blueprint."""
+    blueprint = Blueprint("invenio_app_ils_closures", __name__, url_prefix="")
+    endpoints = app.config.get("RECORDS_REST_ENDPOINTS", [])
```
ntarocco: Given that if the config is wrong, everything will be broken, I would avoid `.get()` and fail early.
```suggestion
    endpoints = app.config["RECORDS_REST_ENDPOINTS"]
```

I guess it is appearing in multiple places (let's ignore those for the moment)

invenio_app_ils/closures/views.py
```diff
@@ -0,0 +1,66 @@
...
+
+    options = endpoints.get(LOCATION_PID_TYPE, {})
```
ntarocco:
```suggestion
    options = endpoints[LOCATION_PID_TYPE]
```

ntarocco (commented):
LGTM! 👏🏼 Great work and very clean!

invenio_app_ils/closures/views.py
JakobMiesner: Done

PR #1232, opened 2025-05-16
circulation: self-checkout of MISSING items marks them as CAN_CIRCULATE

### LINKED ISSUE #978: Self-checkout improvement
Original request: https://cern.service-now.com/nav_to.do?uri=u_request_fulfillment.do?sysparm_query=number=RQF3055743

Could you please explore if it would be possible, when an item with status 'missing' is scanned via the self-checkout, that the item automatically becomes back to 'can circulate' in our backoffice (as the copy is there) and that the loan is issued for the user?

JakobMiesner:
Closes: https://github.com/CERNDocumentServer/cds-ils/issues/978

As I changed the parameters of the function `_set_item_to_can_circulate` I checked, that there are no uses of it in https://github.com/CERNDocumentServer/cds-ils (even though it is marked as private)

### DIFF
--- a/invenio_app_ils/circulation/api.py
+++ b/invenio_app_ils/circulation/api.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2018-2020 CERN.
+# Copyright (C) 2018-2025 CERN.
 #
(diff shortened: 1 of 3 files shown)


kpsherva (approved):
Good job! 🚀

PR #1233, opened 2025-05-30
circulation: added NOT-SPECIFIED delivery method

### LINKED ISSUE #968: Add option "not specified" when doing a loan request for a patron
In back office. We often request a loan for patron in back office, and we don't know which delivery patron would prefer. For now we only have two options, delivery to office and pick-up:

Image

I add per default "pick-up", but this can be confusing for patron as they would maybe prefer the other option.

JakobMiesner:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/968

### DIFF
--- a/invenio_app_ils/circulation/config.py
+++ b/invenio_app_ils/circulation/config.py
@@ -81,6 +81,7 @@ ILS_CIRCULATION_LOAN_WILL_EXPIRE_DAYS = 7
 #: Optional delivery methods when requesting a new loan. Set to empty object to
 # disable it
 ILS_CIRCULATION_DELIVERY_METHODS = {
+    "NOT-SPECIFIED": "Not specified",
     "PICKUP": "Pick it up at the library desk",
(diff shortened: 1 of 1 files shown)


invenio_app_ils/circulation/config.py
```diff
@@ -81,6 +81,7 @@
 #: Optional delivery methods when requesting a new loan. Set to empty object to
 # disable it
 ILS_CIRCULATION_DELIVERY_METHODS = {
+    "NOT-SPECIFIED": "Not specified",
```
kpsherva, line 84: should this method show up as first on the list? which one would be the most commonly used?

JakobMiesner: For me it represents the "default" value and could thus come first in my Opinion. Another option is to use the most commonly used operation, but this might differ between different Libraries

kpsherva (approved):

PR #1235, opened 2025-06-17
internal location: add field accessible_by_patrons

### LINKED ISSUE #955: Hide call number for books in depot (to develop)
In front office. It happens often that patrons come at the desk saying they didn't find a book in the library, that was actually only in a depot (so not freely available unlike books in central library).
The reason is that patrons check the shelf in CLC seeing the book status is "On shelf", see the call number, and don't verify the exact location (Central library, depot 1, depot 2, etc.).
It would be better to hide the call number for books in the depot and replace it by "Available on request".

Image

JakobMiesner:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/955

### DIFF
--- a/invenio_app_ils/documents/jsonresolvers/document_item.py
+++ b/invenio_app_ils/documents/jsonresolvers/document_item.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2019-2020 CERN.
+# Copyright (C) 2019-2026 CERN.
 #
 # invenio-app-ils is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -40,6 +40,9 @@ def jsonresolver_loader(url_map):
                 "identifiers": item.get("identifiers"),
                 "internal_location": {
                     "name": item.get("internal_location", {}).get("name", ""),
+                    "restricted": item.get("internal_location", {}).get(
+                        "restricted", False
+                    ),
                     "location": {
                         "name": (
                             item.get("internal_location", {})
--- a/invenio_app_ils/internal_locations/loaders/jsonschemas/internal_location.py
+++ b/invenio_app_ils/internal_locations/loaders/jsonschemas/internal_location.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2018 CERN.
+# Copyright (C) 2018-2026 CERN.
 #
 # invenio-app-ils is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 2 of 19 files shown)


invenio_app_ils/documents/jsonresolvers/document_item.py
```diff
@@ -40,6 +40,9 @@ def items_resolver(document_pid):
                 "identifiers": item.get("identifiers"),
                 "internal_location": {
                     "name": item.get("internal_location", {}).get("name", ""),
+                    "accessible_by_patrons": item.get("internal_location", {}).get(
```
kpsherva: after some thinking, I am concerned about how targeted this field's name is.
it fulfils the need, but it is not extensible for the future, and not consistent with how we handle "restrictions" in invenio.
I would rather name it as "restrictions" and make it possibly a list, which gives us more flexibility of implementing more restrictions in the past. 
as an example: `restrictions: ["librarians"]` would mean the location is visible to only librarian, so it does not limit the functionality only to the patron
another example: `restrictions: ["archivists"]` could mark places only accessible to some personnel who can go into archives, but not the librarians.
It opens a door to cover more use cases, while not being complex in implementation.

JakobMiesner: In that case, isn't there maybe an already existing module in invenio for record&group specific permissions?

kpsherva: yes, there are already existing implementations. One example would be restriction on the document field.

JakobMiesner: The document also has the field `restricted` which is implemented and used by our `RecordPermission` class (It allows for role based permissions + `restricted`).
As it was the easiest option i chose to also add `restricted` to the `internal_location` instead of role based permissions for now. 

Note that while we have the `RecordPermission` class, the role based authentication is unused in all our systems, as the flag `ILS_RECORDS_EXPLICIT_PERMISSIONS_ENABLED` is disabled everywhere

invenio_app_ils/internal_locations/loaders/jsonschemas/internal_location.py
```diff
@@ -24,3 +24,4 @@ class Meta:
     legacy_ids = fields.List(fields.Str())
     notes = fields.Str()
     physical_location = fields.Str()
+    restricted = fields.Boolean(required=True)
```
kpsherva: shouldn't there be a default set to False?
what will happen with existing location records after we introduce this? Could you check?

JakobMiesner: Yes! Thank you!

JakobMiesner: as the field is not required they will have no value and as the default value in the `document_item` indexer is `False` all item shelves will remain visible

PR #1240, opened 2025-07-02
eitem: add api function to get eitems by provider

### LINKED ISSUE #879: add information about other e-items of the same provider
On the import time (or delete) e-items are matched by the creator value (in this case provider, value is inserted on import time), see
https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/importer/eitems/api.py#L15
this workflow won't work if the e-item was created manually by the librarian - leading to duplications and not deleted e-items. We want to avoid using "source" field of the e-item to match them due to possible human errors while editing this field.
Instead we should add an information in the report stating that another e-item of the currently processed provider was detected and manual action is required.
The information should be visible in "Actions and details" as well as the row should be highlighted in yellow

Image

JakobMiesner:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/879

### DIFF
--- a/invenio_app_ils/eitems/api.py
+++ b/invenio_app_ils/eitems/api.py
@@ -119,3 +119,35 @@ def eitem_event_builder(event, sender_app, obj=None, **kwargs):
     record = kwargs["record"]
     event.update(dict(eitem_pid=record["pid"], document_pid=record.get("document_pid")))
     return event
+
+
+def get_eitems_for_document_by_creator(document_pid, creator, case_insensitive=False):
+    """Find eitems by document pid and creator.
+
+    :param document_pid: The PID of the document of the eitems to search for.
+    :param creator: The creator to filter eitems by.
+    :param case_insensitive: Whether the creator match should be case insensitive.
+    """
+    eitem_search = current_app_ils.eitem_search_cls()
+    creator_match = eitem_search.search_by_document_pid(
+        document_pid=document_pid
+    ).filter(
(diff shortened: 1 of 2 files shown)


invenio_app_ils/eitems/api.py
```diff
@@ -119,3 +119,30 @@ def eitem_event_builder(event, sender_app, obj=None, **kwargs):
...
+    eitem_search = current_app_ils.eitem_search_cls()
+
+    creator_match = eitem_search.search_by_document_pid(
+        document_pid=document_pid
+    ).filter("term", created_by__value=provider)
+
+    source_match = (
+        eitem_search.search_by_document_pid(document_pid=document_pid)
+        .filter("term", source={"value": provider, "case_insensitive": True})
+        .exclude("term", created_by__value=provider)
```
kpsherva: the function is a bit particular to be placed like this in invenioILS directly - especially because we don't have the notion of bulk importer here at the core package.

I would split it to two get_eitems_for_document_by_source and get_eitems_for_document_by_creator and then use a combination of these in cds-ils

returning a tuple is also an indicator that one function (or method) is doing two jobs - the api interface is usually much cleaner if we stick to single responsibility

JakobMiesner: done

kpsherva:
LGTM, please add the release commit

kpsherva (approved):

PR #1241, opened 2025-07-17
update dependencies

### LINKED ISSUE #1006: Upgrade backend dependencies

JakobMiesner:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/1006

tests are failing because it depends on PRs for dependent packages to be merged (e.g. https://github.com/inveniosoftware/invenio-pidrelations/pull/32).
See the PRs in the issue linked above


invenio_app_ils/assets/semantic-ui/less/theme.config
```diff
@@ -0,0 +1,103 @@
+/*
```
kpsherva, line 1: workaround for invenio-theme to use as is. It expects this file to be present

invenio_app_ils/patrons/anonymization.py
```diff
@@ -60,7 +60,6 @@ def dump(search):
     patron_loans = dump(get_loans_by_patron_pid(patron_pid))
 
     patron_profile = UserProfile.get_by_userid(patron_pid).__dict__
-    del patron_profile["_sa_instance_state"]
```
kpsherva, line 63: what was the conclusion for this change? safe to remove?

JakobMiesner: Since the profile is no longer a standalone table, but a JSON column in the users table, it does not have the `_sa_instance_state` field anymore. The field is something SQLAlchemy adds to ORM instances

PR #1242, opened 2025-08-20
fix: ensure the invenio_app_ils configuration is loaded before cds_ils

JakobMiesner:


setup.cfg
```diff
@@ -123,8 +123,8 @@ invenio_base.api_blueprints =
     ils_patrons = invenio_app_ils.patrons.views:get_user_loan_information_blueprint
     ils_notifications = invenio_app_ils.notifications.views:get_notifications_blueprint
 invenio_config.module =
-    00_invenio_app_ils = invenio_app_ils.config
-    00_invenio_app_ils_circulation = invenio_app_ils.circulation.config
+    zy_invenio_app_ils = invenio_app_ils.config
```
zzacharo, line 126: was there any change on the configuration loading mechanism and you needed to add this change? the `00` was supposed to do that...

JakobMiesner: @zzacharo our [tests are failing](https://github.com/CERNDocumentServer/cds-ils/actions/runs/17092523587/job/48502036113?pr=1014) and we think its the configuration being loaded in a wrong order.
(we are getting a `connection to server at "localhost" (::1), port 5432 failed: FATAL:  password authentication failed for user "invenio"`)

PR #1247, opened 2025-09-23
permissions: add backoffice readonly action and permission

### LINKED ISSUE #969: Access token: implement read-only auth
The library has the need of extracting some business metrics via REST APIs. However, some endpoints require `librarian` access, for example:
https://catalogue.library.cern/api/eitems/?q=_created%3A%5B2025-05-01%20TO%202025-05-05%5D

In Invenio, access tokens are bound to a specific user, meaning that the token will have the same rights as the user.
Creating an access token for a user with `librarian` access rights will grant not only read permissions, but also update/delete permissions. We want to avoid that.

### Required changes

1. Grant read-only access to a specific user. Explore options between creating a new `read-only role`, or using OAuth scopes. See the [invenio-oauth2server doc](https://invenio-oauth2server.readthedocs.io/en/latest/).
2. Disable the possibility of creating access tokens.

JakobMiesner:
Update the permission system to allow for readonly-librarians.

* add `backoffice_readonly_access_action` and have it replace `backoffice_access_action` in read operations
   * Users with the `backoffice_readonly_access_action` can read everything that users with the`backoffice_access_action` could read before
* create `PatronOwnerReadPermission` and `backoffice_read_permission` that make use of `backoffice_readonly_access_action`
* renamed `send-notification-to-patron` into `get-notifications-sent-to-patron`, as it is only used for reading notifications

closes: https://github.com/CERNDocumentServer/cds-ils/issues/969

### DIFF
--- a/invenio_app_ils/notifications/views.py
+++ b/invenio_app_ils/notifications/views.py
@@ -38,7 +38,7 @@ def get_notifications_blueprint(_):
             "size": fields.Int(validate=[validate.Range(min=1, max=100)]),
         }
     )
-    @need_permissions("send-notification-to-patron")
+    @need_permissions("get-notifications-sent-to-patron")
     def get_notifications(
         recipient_user_id=None,
         pid_type=None,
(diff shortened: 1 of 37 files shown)


ntarocco (commented):
LGTM! 🚀

invenio_app_ils/notifications/views.py
```diff
@@ -38,7 +38,7 @@ def get_notifications_blueprint(_):
             "size": fields.Int(validate=[validate.Range(min=1, max=100)]),
         }
     )
-    @need_permissions("send-notification-to-patron")
+    @need_permissions("get-notifications-sent-to-patron")
```
kpsherva, line 41: thanks for renaming this, much more clear this way

kpsherva (approved):

PR #1251, opened 2025-10-10
stats: add stats to track ILSRecord insertions, deletions and updates

### LINKED ISSUE #1031: Library Statistics - 4. Number of changes to the Library collections
# Library Statistics

# The KPI - Number of changes to the Library collections
4. Number of changes to the Library collections

JakobMiesner:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/1031

### DIFF
--- a/invenio_app_ils/config.py
+++ b/invenio_app_ils/config.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2018-2020 CERN.
+# Copyright (C) 2018-2025 CERN.
 #
 # invenio-app-ils is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
@@ -15,6 +15,7 @@ You overwrite and set instance-specific configuration by either:
 
 import datetime
 from datetime import timedelta
+from functools import partial
 
 from celery.schedules import crontab
 from invenio_accounts.config import (
@@ -24,7 +25,7 @@ from invenio_records_rest.facets import terms_filter
 from invenio_records_rest.utils import allow_all, deny_all
 from invenio_stats.aggregations import StatAggregator
 from invenio_stats.processors import EventsIndexer
-from invenio_stats.queries import ESTermsQuery
+from invenio_stats.queries import DateHistogramQuery, ESTermsQuery
 
 from invenio_app_ils.document_requests.indexer import DocumentRequestIndexer
 from invenio_app_ils.documents.indexer import DocumentIndexer
@@ -41,6 +42,7 @@ from invenio_app_ils.literature.search import LiteratureSearch
 from invenio_app_ils.locations.indexer import LocationIndexer
 from invenio_app_ils.patrons.indexer import PatronIndexer
 from invenio_app_ils.series.indexer import SeriesIndexer
+from invenio_app_ils.stats.event_builders import ils_record_changed_event_builder
 from invenio_app_ils.vocabularies.indexer import VocabularyIndexer
 
 from .document_requests.api import (
@@ -95,6 +97,7 @@ from .permissions import (
     authenticated_user_permission,
     backoffice_permission,
     backoffice_read_permission,
+    superuser_permission,
     views_permissions_factory,
 )
 from .records.permissions import record_read_permission_factory
@@ -234,12 +237,20 @@ CELERY_BEAT_SCHEDULE = {
     "stats-process-events": {
         "task": "invenio_stats.tasks.process_events",
         "schedule": timedelta(minutes=30),
-        "args": [("record-view", "file-download")],
+        "args": [
+            (
+                "record-view",
+                "file-download",
+                "ils-record-changes-updates",
+                "ils-record-changes-insertions",
+                "ils-record-changes-deletions",
+            )
+        ],
     },
     "stats-aggregate-events": {
         "task": "invenio_stats.tasks.aggregate_events",
         "schedule": timedelta(hours=3),
-        "args": [("record-view-agg", "file-download-agg")],
+        "args": [("record-view-agg", "file-download-agg", "ils-record-changes-agg")],
     },
     "clean_locations_past_closures_exceptions": {
         "task": (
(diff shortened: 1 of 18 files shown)


invenio_app_ils/stats/event_builders.py
```diff
@@ -0,0 +1,75 @@
...
+        raise ValueError("Record must be provided to the event builder.")
+    if not isinstance(record, IlsRecord):
+        return None
+
+    # We have to get the pid_type directly through the private accessor,
+    # as the pid might not be registered directly after the record creation.
+    pid_type = record._pid_type
+
+    event_data = {
+        "timestamp": datetime.datetime.now().isoformat(),
```
ntarocco: we normally store dates in UTC timezone, without timezone info. The client can then transform it.
I don't exactly recall, is `now()` using the server timezone?

invenio_app_ils/stats/event_builders.py
```diff
@@ -0,0 +1,75 @@
...
+    """Add the user_id of the librarian performing the action to the event."""
+
+    user_id = None
+    if (
+        current_user
+        and current_user.is_authenticated
+        and has_request_context()
+        and backoffice_permission().allows(g.identity)
+    ):
+        user_id = current_user.get_id()
```
ntarocco, line 57: I am sure you were expecting my comment here ☺️ 
Can we avoid all these globals? Why `user_id` can be `None`?

invenio_app_ils/stats/event_builders.py
```diff
@@ -0,0 +1,75 @@
...
+
+    return event
+
+
+def add_ils_record_pid_to_event(event, sender_app, record=None, **kwargs):
+    """Add the record pid to the event."""
+    if not record:
+        raise ValueError("Record must be provided to the event builder.")
+    if not isinstance(record, IlsRecord):
+        return event
```
ntarocco: I would add a comment to explain why are skipping it

invenio_app_ils/stats/processors.py
```diff
@@ -0,0 +1,29 @@
...
+
+"""ILS stats preprocessors."""
+
+
+def add_record_change_ids(doc):
+    """Add unique_id and aggregation_id to the doc."""
+
+    # We use this field to group by during aggregation
+    doc["aggregation_id"] = (
+        f"{doc.get('pid_type')}__{doc.get('method')}__{doc.get('user_id')}"
```
ntarocco:
```suggestion
        f"{doc['pid_type']}__{doc['method']}__{doc['user_id']}"
```
unless they are really optional, but in that case you might have `__None`. Same below.

invenio_app_ils/stats/processors.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+
+    # unique_id identifies each individual event for invenio-stats.
+    # Including pid_value ensures distinctness even when multiple records
+    # are created within the same second (e.g. batch imports), preventing
+    # false deduplication caused by timestamp truncation by invenio-stats.
```
kpsherva: I would add more explicit description of the example when we need this

invenio_app_ils/stats/processors.py
```diff
@@ -0,0 +1,29 @@
...
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details
+
+"""ILS stats preprocessors."""
+
+
+def add_record_change_ids(doc):
+    """Add unique_id and aggregation_id to the doc."""
+
+    # We use this field to group by during aggregation
```
kpsherva: here I would also add an example

invenio_app_ils/config.py
```diff
@@ -1003,6 +1078,44 @@ def _(x):
...
+                method="method",
+            ),
+            metric_fields=dict(
+                count=("sum", "count", {}),
+            ),
+        ),
+    ),
+    "ils-record-changes-per-user": dict(
+        cls=ESTermsQuery,
+        permission_factory=superuser_permission,
```
kpsherva, line 1101: please add a recipe to out digital repositories on how to query this stat, make it detailed:)

invenio_app_ils/stats/event_builders.py
JakobMiesner: you are right, thanks for the catch

and now is using the server timezone, yes. See [here](https://docs.python.org/3/library/datetime.html#datetime.datetime.now)

invenio_app_ils/stats/event_builders.py
JakobMiesner: I wanted to ensure that the record has a pid but I solved it differently now

invenio_app_ils/stats/event_builders.py
JakobMiesner: @ntarocco  :D
I was able to remove one but don't know how to completely avoid them here unfortunately. 
invenio-stats is using it the same way [here](https://github.com/JakobMiesner/invenio-stats/blob/efb858a3500b4c269b02b09f50de382e0c6700fd/invenio_stats/utils.py#L89)

user_id can be None if either:
- the change to the ils record is not performed during a request (e.g. automatic update)
- the user that performed the change does not have backoffice permission. In that case we do not track the user_id for GDPR reasons

invenio_app_ils/stats/processors.py
JakobMiesner: you are right. Changed it.

invenio_app_ils/stats/processors.py
JakobMiesner: done

invenio_app_ils/stats/processors.py
JakobMiesner: done

invenio_app_ils/config.py
JakobMiesner: added ticket [here](https://github.com/orgs/CERNDocumentServer/projects/30/views/1?pane=issue&itemId=135706591&issue=CERNDocumentServer%7Ccds-ils%7C1040)

invenio_app_ils/stats/processors.py
```diff
@@ -0,0 +1,28 @@
...
+# under the terms of the MIT License; see LICENSE file for more details
+
+"""ILS stats preprocessors."""
+
+
+def add_record_change_ids(doc):
+    """Add unique_id and aggregation_id to the doc."""
+
+    # We use this field to group by during aggregation.
+    # e.g. the count of created eitems by a user with id 7 is tracked under eitmid__insert__None.
```
kpsherva: only one minor thing.... 
`eitmid__insert__None` the None part I find it a bit strange. Is this avoidable?

JakobMiesner: For changes not performed by users or by users that have no backoffice access the user id will be None. (please look [here](https://github.com/inveniosoftware/invenio-app-ils/pull/1251#discussion_r2460780574) for a more detailed description) 
I could do a check in the `aggregation_id` creation and replace the `None` with something like `no_user_id` so it would be `eitmid__insert__no_user_id`, but I dont really find it much clearer. And in the end the `aggreagtion_id` is only used for aggregation and will not really be interacted with.

kpsherva: could it be just `etmid_insert` instead? also... it should never happen that a non-librarian user creates any record except a document request. It seems like an overkill to have the ` <>_None` pattern in the first place

ntarocco (commented):
There is still a comment from Karolina, otherwise LGTM ☺️

invenio_app_ils/stats/processors.py
JakobMiesner: @kpsherva sounds good. I changed it :) I think the non-librarian record change can happen when e.g. a celery job changes any record. I don't think that the global `current_user` would evaluate to anything in that case.

kpsherva: also importer comes to mind, could happen in the importer. if it is none, maybe we should have value "script/importer"?

JakobMiesner: was moved into new ticket [here](https://github.com/orgs/CERNDocumentServer/projects/30/views/1?pane=issue&itemId=136673521&issue=CERNDocumentServer%7Ccds-ils%7C1042)

PR #1252, opened 2025-11-05
setup: bump oauthclient major version

palkerecsenyi:
* Bumping the major version of invenio-oauthclient to 6.0.0

* This includes https://github.com/inveniosoftware/invenio-oauthclient/pull/360 which is not necessarily breaking but requires a manual DB migration for very large instances, as documented in the module's upgrade guide.

* Further changes are yet come in v6, such as https://github.com/inveniosoftware/invenio-oauthclient/pull/328.


setup.cfg
```diff
@@ -42,7 +42,7 @@ install_requires =
     invenio-access>=4.1.0,<5.0.0
     invenio-accounts>=6.1.1,<7.0.0
     invenio-oauth2server>=3.2.0,<4.0.0
-    invenio-oauthclient>=5.1.0,<6.0.0
+    invenio-oauthclient>=6.0.0,<7.0.0
```
kpsherva, line 45: @JakobMiesner this change should also be reflected in cds-ils, let's schedule it for the next deployment

JakobMiesner:
If >50000 rows in oauthclient_remoteaccount table do [this](https://github.com/inveniosoftware/invenio-oauthclient/blob/master/docs/upgrading.rst#v600)

source @palkerecsenyi

JakobMiesner (approved):

PR #1253, opened 2025-11-05
fix: make terms query that gets eitems by creator use correct keyword

JakobMiesner:


kpsherva (approved):

invenio_app_ils/eitems/api.py
```diff
@@ -133,7 +133,7 @@ def get_eitems_for_document_by_creator(document_pid, creator, case_insensitive=F
         document_pid=document_pid
     ).filter(
         "term",
-        created_by__value={"term": creator, "case_insensitive": case_insensitive},
+        created_by__value={"value": creator, "case_insensitive": case_insensitive},
```
kpsherva, line 136: shouldn't it be?
```suggestion
        created_by={"value": creator, "case_insensitive": case_insensitive},
```

JakobMiesner: @kpsherva created_by is an object that contains value. The second "value" inside curly braces is an OS keyword.
See [here](https://github.com/inveniosoftware/invenio-app-ils/blob/29c489e4b19b2820f430476bbf1a500b7b8e5550/invenio_app_ils/eitems/mappings/os-v2/eitems/eitem-v3.0.0.json#L50) for the mapping.



image

PR #1256, opened 2025-11-07
Loan statistics

### LINKED ISSUE #1028: Library Statistics - 1. Turnover rate of the Library collection
# Library Statistics

# The KPIs:
1. Turnover rate of the Library collection:
   1. number of new loans / number of loanable items
   2. number of renewals / number of loanable items

### LINKED ISSUE #1029: Library Statistics - 2. & 3. - Average duration of loan & Availability of requested documents
# Library Statistics

# The KPIs
The library would like to aggregate the following metrics for loans
- average waiting time
- average loan duration
- average count of new loans
- (and more)

Hereby, they want to group the loans on certain parameters like
- creation date of loan
- state of the document while the loan was requested
- (and more)

### LINKED ISSUE #1033: Library Statistics - 6. Overdue loans
# Library Statistics

# The KPIs
6. Overdue loans
    1. Percentage of overdue loans (active overdue loans / active loans)

JakobMiesner:
* Add stat to track number of loan extensions in a period of time (closes https://github.com/CERNDocumentServer/cds-ils/issues/1028)
* Add endpoint that returns a histogram for loans where requested metrics are grouped by and aggregated (closes: https://github.com/CERNDocumentServer/cds-ils/issues/1029, closes: https://github.com/CERNDocumentServer/cds-ils/issues/1033)
* Extend the loans with a new stats object, which contains `waiting_time`, `loan_duration` and `available_items_during_request` (closes: https://github.com/CERNDocumentServer/cds-ils/issues/1029)

tests are failing because of dependency on [this PR](https://github.com/inveniosoftware/invenio-circulation/pull/161)


invenio_app_ils/circulation/indexer.py
```diff
@@ -97,3 +98,64 @@ def index_extra_fields_for_loan(loan_dict):
...
+    stats = {}
+
+    # Time ranges in days
+    if start_date and end_date:
+        loan_duration = (end_date - start_date).days
+        stats["loan_duration"] = loan_duration
+
+    if creation_date and start_date:
+        waiting_time = (start_date - creation_date).days
+        stats["waiting_time"] = waiting_time if waiting_time >= 0 else None
```
ptamarit, line 128: Isn't it better to return `0` than `None`?
If we want to distinguish self-checkout from normal loans, this won't be enough (since a normal loan can be retrieved on the same day).

invenio_app_ils/circulation/indexer.py
```diff
@@ -97,3 +98,64 @@ def index_extra_fields_for_loan(loan_dict):
...
+        )
+        hits = search_result["hits"]["hits"]
+        if len(hits) == 1:
+            request_transition_event = hits[0]["_source"]
+            available_items_during_request_count = request_transition_event["value"]
+            stats["available_items_during_request"] = (
+                available_items_during_request_count > 0
+            )
+        elif len(hits) > 1:
+            raise ValueError("Multiple request transition events.")
```
ptamarit: The error message should be more clear. Maybe include the loan ID and explain how to fix the problem.

invenio_app_ils/circulation/stats/api.py
```diff
@@ -49,3 +55,159 @@ def fetch_most_loaned_documents(from_date, to_date, bucket_size):
     )
 
     return res
+
+
+_OS_NATIVE_AGGREGATE_FUNCTION_TYPES = {"avg", "sum", "min", "max"}
+_VALID_AGGREGATE_FUNCTION_TYPES = _OS_NATIVE_AGGREGATE_FUNCTION_TYPES.union({"median"})
+_VALID_DATE_INTERVALS = {"1d", "1w", "1M", "1q", "1y"}
+
+_OS_VALID_FIELD_NAME_PATTERN = re.compile(r"^[A-Za-z0-9_.]+$")
```
ptamarit: Maybe put these at the top (before any function).

invenio_app_ils/circulation/stats/api.py
```diff
@@ -49,3 +55,159 @@ def fetch_most_loaned_documents(from_date, to_date, bucket_size):
...
+        Example: [{"field": "start_date", "interval": "monthly"}, {"field": "state"}]
+    :param metrics: List of metric dictionaries with 'field' and 'aggregation' keys.
+        Example: [{"field": "loan_duration", "aggregation": "avg"}]
+    :returns: OpenSearch aggregation results with multi-terms histogram and optional metrics
+    """
+
+    loan_cls = current_circulation.loan_record_cls
+    loan_date_fields = loan_cls.DATE_FIELDS + loan_cls.DATETIME_FIELDS + ["_created"]
+
+    # Validate group_by param
```
ptamarit: Marshmallow could be used for all the validations.

invenio_app_ils/circulation/stats/api.py
```diff
@@ -49,3 +55,159 @@ def fetch_most_loaned_documents(from_date, to_date, bucket_size):
...
+            description="group_by must contain at least one grouping field"
+        )
+
+    for grouping in group_by:
+        if "field" not in grouping:
+            raise InvalidParameterError(
+                description="Each group_by item must be a dict with 'field' key"
+            )
+
+        _validate_os_field_name(grouping["field"])
```
ptamarit: Would it be possible to have an allow list instead of a regexp?

invenio_app_ils/stats/processors.py
```diff
@@ -28,3 +34,48 @@ def add_record_change_ids(doc):
...
+    """
+
+    def run(self):
+        """Process events queue and reindex affected loans.
+
+        First index events related to loans and afterwards trigger a reindex of the affected loans.
+        The loan indexer can then consume the updated events index.
+        """
+        actions = [action for action in self.actionsiter()]
+        res = search.helpers.bulk(self.client, actions, stats_only=True, chunk_size=50)
```
ptamarit, line 77: Could you maybe clarify what this part does?

invenio_app_ils/circulation/indexer.py
JakobMiesner: I just wanted to make sure there are no negative values, which would skew the statistics. Like when doing a sum over the waiting time and there is one loan where the `start_date` was manually set before the `creation_date` I don't think it should influence the waiting time statistic. But it might be good to have a discussion about how to handle such cases?

Regarding distinguishing `self-checkout` and `normal` loans: one can filter with the `q` parameter for the delivery option.

invenio_app_ils/circulation/stats/api.py
JakobMiesner: That would surely be possible, but might limit in how the endpoint can be used and it requires more work from our side to keep track of all possible fields.
The histogram will be implemented for multiple record types and adding+maintaining an allow list might be tedious.
Also the endpoint is behind the backoffice permission, so I do not think it is too critical?

invenio_app_ils/circulation/stats/api.py
JakobMiesner: moved it :)

invenio_app_ils/circulation/stats/api.py
JakobMiesner: done

invenio_app_ils/stats/processors.py
JakobMiesner: I added a comment. Hope that makes it more clear

invenio_app_ils/circulation/indexer.py
JakobMiesner: done

invenio_app_ils/stats/processors.py
```diff
@@ -42,3 +48,46 @@ def filter_extend_transitions(query):
...
+        The loan indexer can then consume the updated invenio-stats events index.
+        This reindex is triggered so the loan index has up-to-date information.
+
+        Example:
+            When a loan is requested, an event is placed in the queue that stores the
+            loan PID and how many items were available at the time of the request.
+            When the event is indexed with this class, it is moved from the queue into
+            the events index. Afterwards, the loan is reindexed and, during this
+            process, the loan indexer requests the state of the document during request
+            from the events index.
```
kpsherva:
```suggestion
            the events index. Afterwards, the loan is reindexed and, during this
            process, the loan indexer gets the state of the document from the events index.
            
```

invenio_app_ils/circulation/indexer.py
```diff
@@ -97,3 +98,68 @@ def index_extra_fields_for_loan(loan_dict):
...
+        loan_duration = (end_date - start_date).days
+        stats["loan_duration"] = loan_duration
+
+    if creation_date and start_date:
+        waiting_time = (start_date - creation_date).days
+        stats["waiting_time"] = waiting_time if waiting_time >= 0 else None
+
+    # Document availability during loan request
+    stat_events_index_name = "events-stats-loan-transitions"
+    if current_search_client.indices.exists(index=stat_events_index_name):
```
kpsherva, line 132: I think it would be better to fail rather than silently not index things  if this conditional statatement is not fulfiled. ping @ntarocco WDYT?

invenio_app_ils/circulation/stats/views.py
```diff
@@ -131,3 +154,42 @@ def get(self, *args, **kwargs):
...
+            search,
+            parsed_args["group_by"],
+            parsed_args["metrics"],
+        )
+
+        response = {
+            "buckets": aggregation_buckets,
+        }
+
+        return jsonify(response)
```
kpsherva:
```suggestion
        return jsonify(response), 200 # or not 200 ?
```

invenio_app_ils/circulation/stats/views.py
```diff
@@ -131,3 +154,42 @@ def get(self, *args, **kwargs):
...
+        loan_cls = current_circulation.loan_record_cls
+        loan_date_fields = (
+            loan_cls.DATE_FIELDS + loan_cls.DATETIME_FIELDS + ["_created"]
+        )
+
+        schema = HistogramParamsSchema(loan_date_fields)
+        try:
+            parsed_args = schema.load(request.args.to_dict())
+        except ValidationError as e:
+            raise InvalidParameterError(description=e.messages)
```
kpsherva: is this returning json error in REST?

JakobMiesner: It is. InvalidParameterException inherits from class [RESTException](https://github.com/inveniosoftware/invenio-rest/blob/a48272243a466862fe089a25e395d8d475cee64b/invenio_rest/errors.py#L47).

invenio_app_ils/circulation/stats/views.py
JakobMiesner: done

invenio_app_ils/stats/processors.py
JakobMiesner: done

invenio_app_ils/circulation/indexer.py
JakobMiesner: for context: the check was added because if there is no event at the moment a loan is indexed, there will be no events index and the loan indexer will throw an error in the lines below the `if`, leading to not/wrongly indexed loans.
I could manually create the index in our systems during deploy and add it to the release notes that the stats index needs to be created for the system to work. But then the setup script should also create the index for new systems.

ntarocco: Agree, if it is required before running this code (weird, shouldn't it be created automatically if not existing?), then I would fail instead.

JakobMiesner: I added a warning when this happens but still would not fail. While the index is required for adding a specific stat field (namely `available_items_during_request`) to the loans, the stat is completely optional and I would not let the whole indexing of the loan fail because of it.
Also, the required index `events-stats-loan-transitions` is handled by `invenio-stats`. It is created up to 30 minutes after a loan goes through a transition for the first time. (30 minutes because the indexing of the events for `invenio-stats` is handled by a celery beat task)
Making it fail would also crash all tests and make new systems fail for at least the first 30 minutes.
If we really want the loan indexer to fail here I see only the following solution to execute for all tests and new systems: 
- Manually create the index in `invenio-app-ils`. I think this is a bad idea because the index is created and managed by `invenio-stats`.
- Creating a loan, sending it through a transition and forcing `invenio-stats` to index the event. This feels super hacky. 


Also note:
If this happens to a loan and it skips this stat, and an event for this loan later gets indexed, the loan will also be reindexed and collect the missing information.

invenio_app_ils/circulation/indexer.py
```diff
@@ -97,3 +98,75 @@ def index_extra_fields_for_loan(loan_dict):
...
+            stats["available_items_during_request"] = (
+                available_items_during_request_count > 0
+            )
+        elif len(hits) > 1:
+            raise ValueError(
+                f"Multiple request transition events for loan {loan_pid}."
+                "Expected zero or one."
+            )
+    else:
+        current_app.logger.warning(
```
kpsherva:
```suggestion
        current_app.logger.error(
```
and please throw an exception so we also get sentry alert

JakobMiesner: done

Issue #1257, opened 2025-11-12
Document Cover Not Generated


sinjup:
Package version : invenio-app-ils 2.0.1+

Document covers are not being generated when creating new documents because the cover builder function only checks the `cover_metadata` field for ISBN, but ISBN is actually stored in the `identifiers` array. This results in all documents showing placeholder covers instead of actual book covers from Open Library.
The root cause is that the `build_ils_demo_cover_urls()` function in `invenio_app_ils/literature/covers_builder.py` doesn't check the `identifiers` array where ISBN is actually stored in document records.

## Steps to Reproduce

1. Create a new document via REST API with an ISBN in the `identifiers` array:
```bash
curl -X POST http://localhost:5000/api/documents \
  -H "Content-Type: application/json" \
  -d '{
    "title": "The Great Gatsby",
    "authors": [{"full_name": "F. Scott Fitzgerald"}],
    "document_type": "BOOK",
    "publication_year": "1925",
    "languages": ["eng"],
    "identifiers": [
      {
        "scheme": "ISBN",
        "value": "978-0-7432-7356-5"
      }
    ]
  }'
```

2. Retrieve the created document:
```bash
curl -X GET http://localhost:5000/api/documents/{pid}
```

3. Check the `cover_metadata.urls` field in the response

4. See that `is_placeholder` is `true` and cover URLs point to placeholder image

## Expected behavior
The document should have actual book cover URLs from Open Library:
```json
{
  "cover_metadata": {
    "urls": {
      "is_placeholder": false,
      "small": "https://covers.openlibrary.org/b/isbn/978-0-7432-7356-5-S.jpg",
      "medium": "https://covers.openlibrary.org/b/isbn/978-0-7432-7356-5-M.jpg",
      "large": "https://covers.openlibrary.org/b/isbn/978-0-7432-7356-5-L.jpg"
    }
  }
}
```

## Actual behavior
The document has placeholder cover URLs:
```json
{
  "cover_metadata": {
    "urls": {
      "is_placeholder": true,
      "small": "http://localhost:5000/static/images/placeholder.png",
      "medium": "http://localhost:5000/static/images/placeholder.png",
      "large": "http://localhost:5000/static/images/placeholder.png"
    }
  }
}
```

## Root Cause Analysis

**File:** `invenio_app_ils/literature/covers_builder.py` (line 16)

**Current code:**
```python
def build_ils_demo_cover_urls(metadata):
    """Build working urls for demo data."""
[...]

PR #1258, opened 2025-11-12
Fix: Document Cover Not Generated #1257

AryaPRaj:
* BETTER Handles ISBN from multiple metadata sources with priority.

### Description

> Improved ISBN extraction logic to handle cases where ISBN is stored in the `identifiers` array instead of directly in metadata fields.

### DIFF
--- a/invenio_app_ils/literature/covers_builder.py
+++ b/invenio_app_ils/literature/covers_builder.py
@@ -13,7 +13,19 @@ from flask import url_for
 def build_ils_demo_cover_urls(metadata):
     """Build working ulrs for demo data."""
     cover_metadata = metadata.get("cover_metadata", {})
-    isbn = cover_metadata.get("ISBN", "")
+
+    isbn = (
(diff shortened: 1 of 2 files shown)


invenio_app_ils/literature/covers_builder.py
```diff
@@ -13,7 +13,19 @@
 def build_ils_demo_cover_urls(metadata):
     """Build working ulrs for demo data."""
     cover_metadata = metadata.get("cover_metadata", {})
-    isbn = cover_metadata.get("ISBN", "")
+
```
ntarocco, line 16: Thanks for this change. We can add this, however be aware that this is a just a `demo` implementation.
The idea is that each adopter will decide how to implement this, give that it often requires a subscription with a covers provider.

See for example [our CERN implementation here](https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/literature/tasks.py#L17).

kpsherva (approved):

PR #1261, opened 2026-01-04
stats: Track all loan transitions

JakobMiesner:
In the loan transition events introduced [here](https://github.com/inveniosoftware/invenio-app-ils/pull/1256) we previously only tracked extension events. 
This PR removes this filter and starts tracking events for all loan transitions.


invenio_app_ils/config.py
```diff
@@ -1162,13 +1157,13 @@ def _(x):
             aggregated_fields=["user_id"],
         ),
     ),
-    "loan-extensions": dict(
+    "loan-transitions": dict(
```
kpsherva, line 1160: to make sure: does this change influcence any existing stats in prod?

PR #1262, opened 2026-01-04
stats: Purchase Orders

### LINKED ISSUE #1034: Library Statistics - 7. Purchase orders
# Library Statistics

# The KPIs

7. Purchase orders:
    1. Average Waiting time: waiting time of purchase orders / number of purchase orders that were received
    2. Average ordering time: ordering time of purchase orders / number of purchase orders that were received
        

JakobMiesner:
* Implement an endpoint to expose Order statistics, which returns a histogram for stats where requested metrics are grouped by and aggregated
* Extend the orders index with a new stats object
* The stats object contains order_processing_time and document_request_waiting_time 

closes: https://github.com/CERNDocumentServer/cds-ils/issues/1034


invenio_app_ils/acquisition/mappings/os-v2/acq_orders/order-v1.0.0.json
```diff
@@ -358,6 +358,11 @@
       },
       "provider_pid": {
         "type": "keyword"
+      },
```
kpsherva, line 361: for this one, to be sure, did you check if it works as we expect it, while trying to index and object under this field?

invenio_app_ils/document_requests/indexer.py
```diff
@@ -26,14 +28,25 @@ def index_referenced_records(docreq):
...
     # fetch and index the document
     document_pid = docreq.get("document_pid")
-    referenced = []
     if document_pid:
         document_cls = current_app_ils.document_record_cls
         document = document_cls.get_record_by_pid(document_pid)
         referenced.append(dict(pid_type=DOCUMENT_PID_TYPE, record=document))
 
+    # fetch and index the related order (physical_item_provider)
+    physical_item_provider = docreq.get("physical_item_provider")
```
kpsherva, line 41: could you explain why we need this addition?

invenio_app_ils/stats/histogram/schemas.py
```diff
@@ -0,0 +1,110 @@
...
+
+from invenio_app_ils.errors import InvalidParameterError
+
+_OS_VALID_FIELD_NAME_PATTERN = re.compile(r"^[A-Za-z0-9_.]+$")
+_OS_NATIVE_AGGREGATE_FUNCTION_TYPES = {"avg", "sum", "min", "max"}
+_VALID_AGGREGATE_FUNCTION_TYPES = _OS_NATIVE_AGGREGATE_FUNCTION_TYPES.union({"median"})
+_VALID_DATE_INTERVALS = {"1d", "1w", "1M", "1q", "1y"}
+
+
+def validate_field_name(field_name):
```
kpsherva: is this function reused somewhere outside the SecureFieldName?

invenio_app_ils/stats/histogram/schemas.py
```diff
@@ -0,0 +1,110 @@
...
+                "Field names may contain only alphanumeric characters, underscores, "
+                "and dots."
+            )
+        )
+
+
+class SecureFieldNameField(fields.String):
+    """Marshmallow field that validates field names to prevent injection attacks."""
+
+    def _deserialize(self, value, attr, data, **kwargs):
```
kpsherva: in marshmallow, do we have a separate method of Field class to validate the value? otherwise, if the interface of the field class does provide it, the validation should most likely be done in the schema itself.
I find it a bit strange that validation is part of deserialize

invenio_app_ils/stats/histogram/views.py
```diff
@@ -0,0 +1,34 @@
...
+from invenio_app_ils.stats.histogram.serializers import histogram_stats_response
+
+
+def create_histogram_view(blueprint, app, pid_type, resource_cls, url_prefix):
+    """Add url rule for histogram view."""
+
+    assert url_prefix.startswith("/"), "url_prefix must start with /"
+    assert not url_prefix.endswith("/"), "url_prefix must not end with /"
+
+    endpoints = app.config.get("RECORDS_REST_ENDPOINTS")
```
kpsherva, line 19: from the point of view of ILS, this should be current_app.config
we need to make sure the value of the config is coming from the overlay application (cds-ils in our case)

invenio_app_ils/stats/histogram/views.py
```diff
@@ -0,0 +1,34 @@
...
+    default_media_type = document_endpoint.get("default_media_type")
+    histogram_serializers = {"application/json": histogram_stats_response}
+
+    histogram_stats_view_func = resource_cls.as_view(
+        resource_cls.view_name,
+        serializers=histogram_serializers,
+        default_media_type=default_media_type,
+        ctx={},
+    )
+    blueprint.add_url_rule(
```
kpsherva, line 30: as far as I understand, this will add histogram endpoints for all record types. Is there any pid_type which should be excluded, because we do not add stats for them?

invenio_app_ils/stats/histogram/views.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def create_histogram_view(blueprint, app, pid_type, resource_cls, url_prefix):
+    """Add url rule for histogram view."""
+
+    assert url_prefix.startswith("/"), "url_prefix must start with /"
+    assert not url_prefix.endswith("/"), "url_prefix must not end with /"
+
+    endpoints = app.config.get("RECORDS_REST_ENDPOINTS")
+    document_endpoint = endpoints.get(pid_type)
```
kpsherva: should it be more generic in naming?
```suggestion
    record_endpoint = endpoints.get(pid_type)
```

invenio_app_ils/stats/histogram/api.py
```diff
@@ -0,0 +1,109 @@
...
+    """Return the aggregation name used for a metric.
+
+    :param metric: Must include 'field' and 'aggregation' keys.
+    :returns: The aggregation field name in the form '<aggregation>_<field>'.
+    """
+
+    return f"{metric['aggregation']}__{metric['field']}"
+
+
+def get_record_statistics(date_fields, search, requested_group_by, requested_metrics):
```
kpsherva, line 27: I would prefer to review this part together

invenio_app_ils/acquisition/indexer.py
```diff
@@ -0,0 +1,72 @@
...
+
+from datetime import datetime
+
+from invenio_search import current_search_client
+
+from invenio_app_ils.acquisition.api import ORDER_PID_TYPE
+from invenio_app_ils.proxies import current_app_ils
+
+
+def index_stats_fields_for_order(order_dict):
```
kpsherva, line 18: how come here we don't check the feature flag or the index existence?

invenio_app_ils/stats/histogram/schemas.py
JakobMiesner: You are right. Moved it inside SecureFieldName

invenio_app_ils/acquisition/indexer.py
JakobMiesner: we check it when we register the indexing hook in `invenio_app_ils/ext.py`

invenio_app_ils/stats/histogram/views.py
JakobMiesner: thanks for catching that!

invenio_app_ils/stats/histogram/views.py
JakobMiesner: It does not automatically add it for all record types. Records have to register the view manually (for now only loans and orders do with document requests to come)

invenio_app_ils/stats/histogram/views.py
JakobMiesner: I followed the way it was done with other blueprints (see [here](https://github.com/search?q=repo%3Ainveniosoftware%2Finvenio-app-ils%20blueprint(app)&type=code)) which are then passed to `invenio_base.api_blueprints`. Does this not ensure we get the correct app?

invenio_app_ils/stats/histogram/schemas.py
JakobMiesner: true! changed it

invenio_app_ils/document_requests/indexer.py
JakobMiesner: A order now looks for the potential `document_request´ that references it during indexing. So when an order gets created through the `document_request´ edit page we might run into the following situation

1. Document request gets created and indexed without an attached order
2. Order gets created for document request
3. Order gets indexed and looks for related document request. It (falsely) finds none, as the order pid is not yet registered to the document request
4. Order pid gets added to document request and document request gets indexed

So I wanted the order to be indexed again after step 4

invenio_app_ils/acquisition/mappings/os-v2/acq_orders/order-v1.0.0.json
JakobMiesner: It does. When the first orders are indexed with stats the following mapping is generated.



```        
"stats": {
          "dynamic": "true",
          "properties": {
            "document_request_waiting_time": {
              "type": "long"
            },
            "order_processing_time": {
              "type": "long"
            }
          }
       },
```

invenio_app_ils/stats/histogram/views.py
kpsherva: could you check if we override the config, this part still gets the overriden values? you can temporarily set an environment variable and a breakpoint to check it

kpsherva (approved):
re-reviewed, LGTM!

invenio_app_ils/stats/histogram/views.py
JakobMiesner: I did `export INVENIO_RECORDS_REST_ENDPOINTS=[]` and I see the overwritten value appear in other pieces of code in `invenio_app_ils` where we use `app.config.get("RECORDS_REST_ENDPOINTS")`

PR #1263, opened 2026-01-05
stats: add feature flag for indexing stats data onto records

JakobMiesner:
- add feature flag for extending record indices with stats
- raise error when loan is indexed while loan transition events index does not exits


invenio_app_ils/circulation/indexer.py
```diff
@@ -129,44 +130,40 @@ def index_stats_fields_for_loan(loan_dict):
...
-                        {"term": {"pid_value": loan_pid}},
-                    ],
-                }
-            },
-        }
-
-        search_result = current_search_client.search(
-            index=stat_events_index_name, body=search_body
+    if not current_search_client.indices.exists(index=stat_events_index_name):
+        raise LoanTransitionEventsIndexMissingError()
```
kpsherva, line 134: I am confused now, let's have a chat

PR #1269, opened 2026-02-02
Checkout transaction date

kpsherva:


invenio_app_ils/circulation/loaders/schemas/json/loan_checkout.py
```diff
@@ -28,6 +28,8 @@ class LoanCheckoutSchemaV1(LoanBaseSchemaV1):
     end_date = DateString()
     force = fields.Bool(load_default=False)
 
+    transaction_date = DateTimeString(missing=set_missing_transaction_date)
```
JakobMiesner, line 31: @kpsherva do you think the other Schemas (e.g. LoanRequestSchemaV1) also require setting the transaction date like done here?

kpsherva: no, I've checked and the other workflows are using LoanSchemaV1, which already does this. Unless you can think of any other place I should check?

JakobMiesner: The ones I saw seem to be using `LoanBaseSchemaV1`, same as the loan checkout transition.

e.g. [`LoanRequestSchemaV1`](https://github.com/inveniosoftware/invenio-app-ils/blob/d6bc648470d9b0da371005875adbf21b2ba5d605/invenio_app_ils/circulation/loaders/schemas/json/loan_request.py#L106) uses `LoanBaseSchemaV1`, which does not add `transaction_date`.

I actually [see no usage of `LoanSchemaV1`](https://github.com/search?q=repo%3Ainveniosoftware%2Finvenio-app-ils%20LoanSchemaV1&type=code) ^^'

[All cases with LoanBaseSchemaV1 usage here](https://github.com/search?q=repo%3Ainveniosoftware%2Finvenio-app-ils%20LoanBaseSchemaV1&type=code)

Maybe we should add the `transaction_date` to `LoanBaseSchemaV1`?

kpsherva: https://github.com/inveniosoftware/invenio-circulation/blob/master/invenio_circulation/records/loaders/__init__.py#L15
this is the usage of LoanSchemaV1 and this is used to perform state changes on circulation (not via endpoint but via action triggers)
I wouldn't change the base schema, because LoanRequest already have the transaction date assigned by the Transition class, and SelfCheckout as well (goes directly from CREATED to ON_LOAN state, also through transitions)

JakobMiesner (approved):

sakshamarora1 (approved):

PR #1283, opened 2026-08-13
API loan dates update

### LINKED ISSUE #1004: Editable loan dates on 'returned' loan should not be possible
**Describe the bug**
It looks we can edit loan dates on returned loans.

**To Reproduce**
On the loan -> Edit loan dates
It should only be possible for ongoing loans.

**Expected behavior**
It should not be possible if the loan is 'returned', e.g. we should have a message saying the loan is finished and therefore the dates cannot be edited. Suggest that a new loan should be issued.

callumlewis05:
Loan dates are now only editable through the api if the loans are active or pending.

* closes https://github.com/CERNDocumentServer/cds-ils/issues/1004

### DIFF
--- a/invenio_app_ils/circulation/api.py
+++ b/invenio_app_ils/circulation/api.py
@@ -13,6 +13,8 @@ from flask_login import current_user
 from invenio_circulation.config import (
     CIRCULATION_STATES_LOAN_ACTIVE,
     CIRCULATION_STATES_LOAN_COMPLETED,
+    CIRCULATION_STATES_LOAN_REQUEST,
+    CIRCULATION_STATES_LOAN_CANCELLED,
 )
 from invenio_circulation.errors import CirculationException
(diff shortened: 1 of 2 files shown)


invenio_app_ils/circulation/api.py
```diff
@@ -375,39 +377,43 @@ def update_dates_loan(
...
     is_active = state in CIRCULATION_STATES_LOAN_ACTIVE
+    is_pending = state in CIRCULATION_STATES_LOAN_REQUEST
     is_completed = state in CIRCULATION_STATES_LOAN_COMPLETED
+    is_cancelled = state in CIRCULATION_STATES_LOAN_CANCELLED
 
     data = copy(record)
 
-    if is_active or is_completed:
-        today = date.today().strftime("%Y-%m-%d")
+    if is_active:
```
kpsherva, line 386: the code here because quite complex to read, I think we were adding layers of conditional statements throughout the years. Could you propose something to make it more readable, avoiding all the nested if statements?

